PR #5, opened 2022-02-16
Feature/refactor-eks-fetchers

ofiriro3:
In this PR, we upgraded the AWS fetchers to support the new EKS benchmark.
 https://github.com/elastic/security-team/issues/2429

- Closes: https://github.com/elastic/security-team/issues/2429


resources/fetchers/ecr_fetcher.go
```diff
@@ -24,28 +33,69 @@ type ECRResource struct {
 	EcrRepositories
 }
 
-func NewECRFetcher(awsCfg aws.Config, cfg ECRFetcherConfig) (Fetcher, error) {
-	ecr := NewEcrProvider(awsCfg)
+func NewECRFetcher(awsCfg AwsFetcherConfig, cfg ECRFetcherConfig) (Fetcher, error) {
+	ecrProvider := NewEcrProvider(awsCfg.Config)
+
+	privateRepoRegex := fmt.Sprintf(PrivateRepoRegexTemplate, *awsCfg.AccountID, awsCfg.Config.Region)
```
amirbenun, line 39: I am a bit confused, it seems like that's the only place you are using `AccountID`.
If that's the case, I think you should call `NewAWSIdentityProvider` right here (instead of the code block you pasted on a comment)

amirbenun (changes requested):
Nice work!

resources/fetchers/file_system_fetcher.go
```diff
@@ -50,7 +50,7 @@ func (f *FileSystemFetcher) Fetch(ctx context.Context) ([]FetchedResource, error
 	for _, filePattern := range f.cfg.Patterns {
 		matchedFiles, err := Glob(filePattern)
 		if err != nil {
-			logp.Err("Failed to find matched glob for %s, error - %+v", filePattern, err)
+			return nil, fmt.Errorf("failed to find matched glob for %s, error - %w", filePattern, err)
```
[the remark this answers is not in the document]
ofiriro3: Wasn't on purpose - I reverted it

resources/fetchers/elb_provider.go
```diff
@@ -19,7 +20,7 @@ func NewELBProvider(cfg aws.Config) *ELBProvider {
 	}
 }
 
-/// DescribeLoadBalancer method will return up to 400 results
+// DescribeLoadBalancer / DescribeLoadBalancer method will return up to 400 results
```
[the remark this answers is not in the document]
ofiriro3: I removed it, but it was a change that was suggested by the IDE

resources/fetchers/ecr_fetcher.go
ofiriro3: That's an option.

My idea was that all of the fetchers will receive as parameter `AwsFetcherConfig` our "Cloudbeat" AWS model, which has also the `AccountID`.

More fetchers in the future will need this data.

But I don't mind changing it if I haven't convinced you.

resources/fetchers/aws_identity_provider.go
```diff
@@ -0,0 +1,27 @@
...
+
+func NewAWSIdentityProvider(cfg aws.Config) *AWSIdentityProvider {
+	svc := sts.New(cfg)
+	return &AWSIdentityProvider{
+		client: svc,
+	}
+}
+
+// GetMyIdentity / This method will return your identity (Arn, user-id...)
+func (provider AWSIdentityProvider) GetMyIdentity(ctx context.Context) (*sts.GetCallerIdentityResponse, error) {
```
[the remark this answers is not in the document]
ofiriro3: I am wondering why not expose all of the data that the provider brings us. 

I mean, I totally agree that the func needs to return a non "aws" struct. 
But why do we wish to encapsulate it? and return only part of it.

I mean, If we will want the ARN role from the identity? we will write another helper function?
And if we want them both? will we call both functions causing the beats to do two calls to Amazon?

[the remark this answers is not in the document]
amirbenun: So I think somewhere in the middle should work:
`getIdentity` (private)
`GetAccountDetails` (public) - returns your own struct

or just
`GetAccountDetails` (public) - returns your own struct

resources/fetchers/ecr_fetcher.go
amirbenun: We can keep `AwsFetcherConfig` as the wrapper just AccountID should not be a part of this struct.

resources/fetchers/elb_provider.go
```diff
@@ -29,8 +29,7 @@ func (provider ELBProvider) DescribeLoadBalancer(ctx context.Context, balancersN
 	req := provider.client.DescribeLoadBalancersRequest(input)
 	response, err := req.Send(ctx)
 	if err != nil {
-		logp.Err("Failed to describe load balancers %s from elb, error - %+v", balancersNames, err)
```
amirbenun, line 32: I see you have deleted many error logs, why is that?

amirbenun (approved):

resources/fetchers/elb_provider.go
ofiriro3: I just want the beats to be aligned.

No need to log an error and then propagate it back to the caller and then log again.

PR #7, opened 2022-02-21
Type opa's input and output 

uri-weisman:


evaluator/evaluator_mock.go
```diff
@@ -0,0 +1,65 @@
+// Code generated by mockery v2.10.0. DO NOT EDIT.
```
uri-weisman, line 1: I've used [https://github.com/vektra/mockery](mockery) to generate this mock, it's a very handy tool.
I'll open a task to generate mocks for all files as part of the CI process/ pre-commit.

beater/cloudbeat.go
```diff
@@ -61,7 +61,7 @@ func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
 		return nil, err
 	}
 
-	evaluator, err := opa.NewEvaluator(ctx)
+	evaluator, err := evaluator.NewOpaEvaluator(ctx)
```
ofiriro3, line 64: Awesome !

transformer/events_creator.go
```diff
@@ -55,14 +53,14 @@ func (c *Transformer) processEachResource(results []fetchers.FetchedResource, me
 
 func (c *Transformer) createBeatEvents(policyResource fetchers.FetchedResource, metadata ResourceMetadata) error {
```
ofiriro3, line 54: rename policyResource

transformer/events_creator.go
```diff
@@ -7,28 +7,26 @@ import (
...
+	"github.com/elastic/cloudbeat/evaluator"
 	"github.com/elastic/cloudbeat/resources"
 	"github.com/elastic/cloudbeat/resources/fetchers"
-	"github.com/mitchellh/mapstructure"
 	"time"
 )
 
-type CB func(ctx context.Context, input interface{}) (interface{}, error)
-
 type Transformer struct {
```
ofiriro3, line 16: Not for this PR.
But I am wondering if the name "Transformer" is right for this class.
It seems that this struct is responsible for the lioness part of the pipeline.
Evaluating against the evaluator and decoding the data.
So actually it is the data processor as far as I understand. 

Just a point to think about

uri-weisman: Thought about it as well, `processor` is taken but some variation might be good as well.

transformer/events_creator.go
uri-weisman: Done

transformer/events_creator_test.go
```diff
@@ -41,61 +48,113 @@ var fetcherResult = fetchers.FileSystemResource{
...
 
-	m.Run()
+func (s *EventsCreatorTestSuite) SetupTest() {
+	s.mockedEvaluator = evaluator.MockedEvaluator{}
 }
 
-func TestTransformer_ProcessAggregatedResources(t *testing.T) {
-	tests := []testAttr{
+func (s *EventsCreatorTestSuite) TestTransformer_ProcessAggregatedResources() {
+	var tests = []testAttr{
```
ofiriro3, line 77: I am less admirer of this practice although that's what Go recommended.

I like it better to have multiple tests with different names that describe what the test tries to test, but your code is totally legit.

I feel that here the test cases are not testing the same exact thing, so I would expect here to see three tests.

But again - it's just a personal flavor - your code is totally legit.

ofiriro3: I would have extracted it to another variable at least, just for the sake of readability - but again - my personal flavor.

ofiriro3 (approved):

PR #8, opened 2022-02-21
set Default Leader Value to be true

oren-zohar:
- set default lease value to true 
- improve CI to fetch finding by type
_____

- Closes: https://github.com/elastic/security-team/issues/3131


DaveSys911 (commented):
@oren-zohar The `DefaultLeaderValue = true` as the fallback option can result in unexpected behavior.  Could you elaborate on what were u trying to accomplish?

oren-zohar:
> The DefaultLeaderValue = true as the fallback option can result in unexpected behavior. Could you elaborate on what were u trying to accomplish?

@DaveSys911 Current behavior is that when it can't get the lease info it will not fetch `kube-api` data at all.  We would like it to fetch the data in that case. What unexpected behavior are you expecting?

deploy/k8s/cloudbeat-ds.yaml
```diff
@@ -210,48 +209,48 @@ metadata:
...
-  - ""
-  resources:
-  - nodes/stats
-  verbs:
-  - get
-- nonResourceURLs:
-  - "/metrics"
-  verbs:
-  - get
+  - apiGroups: [""]
```
amirbenun, line 212: indentation in this entire block doesn't seem right

resources/conditions/kube_lease.go
```diff
@@ -14,6 +14,7 @@ import (
 const (
 	PodNameEnvar           = "POD_NAME"
 	DefaultLeaderLeaseName = "elastic-agent-cluster-leader"
+	DefaultLeaderValue     = true
```
amirbenun, line 17: It seems like the default value here may remain false.
If we create the cluster properly with `cloudbeat-ds.yaml` the entire flow should work as expected without fallbacks.
Did I miss anything?
If that's the case, let's rename the ticket to a more indicative name.

amirbenun (commented):
Great progress!

DaveSys911:
@oren-zohar

The same scenarios we are trying to address with leader election and  limit that fetcher to only one pod:
1. It can ddos the k8s API server, making it unavailable 
2. Generate costs by increasing the compute  on the k8s cluster or generating extra redundant data shipped to ELK

I think it's better for the CI to fail if the leader election mechanism is broken, then a default true fallback. 
Another point worth considering is that the agent has its own leader election mechanism, worth checking how other beats are consuming that while running inside the agent.(exp metricbeat).

Anyways, the CI looks fine to me so approving on my end.

DaveSys911 (approved):
Approved.
Please consider the leader election comments.

deploy/k8s/cloudbeat-ds.yaml
oren-zohar: automated formatting. should be ok

resources/conditions/kube_lease.go
oren-zohar: reverting back to true, still keeping the changes so it'll make it easy for us to change default value locally

PR #18, opened 2022-03-07
Post merge checks

oren-zohar:
- Post Merge checks:
  - Go report badge
  - Go coverage badge


.github/workflows/cloudbeat-ci.yml
```diff
@@ -20,9 +20,8 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v2
         with:
-          token:           ${{ secrets.GITHUB_TOKEN }}
```
DaveSys911, line 23: This checkout def works, not sure regarding the change.
Would be better to keep it as is.

PR #22, opened 2022-03-09
Enable process fetcher to fetch process configuration file

ofiriro3:
As part of implementing the entire Kubernetes benchmark, we would like to support the reading of a process-related configuration file.

Closes: https://github.com/elastic/security-team/issues/3263


resources/fetchers/process_fetcher.go
```diff
@@ -2,18 +2,27 @@ package fetchers
...
 )
 
 type ProcessResource struct {
-	PID  string        `json:"pid"`
-	Cmd  string        `json:"command"`
-	Stat proc.ProcStat `json:"stat"`
+	PID    string                 `json:"pid"`
+	Cmd    string                 `json:"command"`
+	Stat   proc.ProcStat          `json:"stat"`
+	Config map[string]interface{} `json:"config"`
```
ofiriro3, line 25: Wondering about the name of the field. 

Will we ever extend it?

resources/fetchers/process_fetcher.go
```diff
@@ -22,7 +31,12 @@ type ProcessesFetcher struct {
 
 type ProcessFetcherConfig struct {
 	BaseFetcherConfig
-	Directory string `config:"directory"` // parent directory of target procfs
+	Directory         string                               `config:"directory"` // parent directory of target procfs
+	RequiredProcesses map[string]ProcessInputConfiguration `config:"required_processes"`
+}
+
+type ProcessInputConfiguration struct {
+	RequiredFileInputArguments []string `config:"required-file-arguments"`
```
ofiriro3, line 39: Better name ?

resources/fetchers/process_fetcher.go
oren-zohar:
```suggestion
	Config common.MapStr `json:"config"`
```

oren-zohar: 🤔

oren-zohar (changes requested):
Looks good, with minor changes. Tests failed locally for some reason, also the integration test shows no data, can you please check both?

config/config.go
```diff
@@ -13,11 +13,18 @@ const DefaultNamespace = "default"
 const ResultsDatastreamIndexPrefix = "logs-cis_kubernetes_benchmark.findings"
 const MetadataDatastreamIndexPrefix = ".logs-cis_kubernetes_benchmark.metadata"
 
+type ProcessInputConfiguration struct {
```
ofiriro3, line 16: I thought about it a lot - Should the process configuration should be a part of the config or part of the process struct. I found reason in both.

In the end, I decided to put it here since I believe that the config is more "external" configuration than "internal".

I mean, the configuration of the process can be defined out of the general config and not the other way around.

ofiriro3 (commented):
Ready for review

cloudbeat.yml
```diff
@@ -17,6 +17,14 @@ cloudbeat:
   - name: kube-api
   - name: process
     directory: "/hostfs"
+    processes:
```
DaveSys911, line 20: You shouldn't edit this file directly - It's auto-generated.
Add your changes to the appropriate [template files ](https://github.com/elastic/cloudbeat/tree/main/_meta/config)
Then run `mage config` to generate updated cloudbeat.yml & cloudbeat. reference.yml files

DaveSys911 (changes requested):
Update cloudbeat.yml via mage, other than that LGTM

cloudbeat.yml
ofiriro3: fixed

DaveSys911: looks good to me

DaveSys911 (approved):
Approved

ofiriro3:
The go.mod and go.sum will be aligned with main (Oren just merged his code where he uses the latest main branch)

resources/fetchers/process_fetcher.go
```diff
@@ -2,15 +2,28 @@ package fetchers
...
 	"github.com/elastic/cloudbeat/resources/fetching"
+	"io/fs"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	"path/filepath"
+	"regexp"
+)
+
+const (
+	CMDArgumentMatcher = "\\b%s=\\/?(\\S+)"
```
amirbenun, line 19: Add a comment about what kind of strings it matches and what are the groups

resources/fetchers/process_fetcher_test.go
```diff
@@ -0,0 +1,184 @@
...
+		BaseFetcherConfig: fetching.BaseFetcherConfig{},
+		RequiredProcesses: map[string]ProcessInputConfiguration{
+			"kubelet": {CommandArguments: []string{"fetcherConfig"}}},
+		Fs: sysfs,
+	}
+	processesFetcher := &ProcessesFetcher{cfg: fetcherConfig}
+
+	fetchedResource, err := processesFetcher.Fetch(context.TODO())
+	assert.Nil(t, err)
+	assert.Equal(t, len(fetchedResource), 1)
```
amirbenun, line 54: When asserting the first argument is the `expected` and the second is the `actual`
Please make sure all over the pr that you are aligned with that

resources/fetchers/process_fetcher_test.go
```diff
@@ -0,0 +1,184 @@
+package fetchers
```
amirbenun, line 1: AMAZING

amirbenun (changes requested):
Very nice!

resources/fetchers/process_fetcher.go
```diff
@@ -19,36 +32,99 @@ type ProcessesFetcher struct {
...
+// As an input this function receives a ProcessInputConfiguration that contains CommandArguments, a string array that represents some process flags
+// The function extracts the configuration file associated with each flag and returns it.
+func (f *ProcessesFetcher) getProcessConfigurationFile(processConfig ProcessInputConfiguration, cmd string, processName string) map[string]interface{} {
+	configMap := make(map[string]interface{}, 0)
+	for _, argument := range processConfig.CommandArguments {
+		// The regex extracts the cmd line flag(argument) value
+		regex := fmt.Sprintf(CMDArgumentMatcher, argument)
+		matcher := regexp.MustCompile(regex)
+		if !matcher.MatchString(cmd) {
+			logp.Error(fmt.Errorf("failed to find argument %s in process %s", argument, processName))
```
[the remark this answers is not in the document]
ofiriro3: Why do you think that we shouldn't add an error here ?
According to the config you want to get a certain config argument for a certain process, I think that debug wise it's an error that should be printed (I don't find it any different from printing error when the config search for a file that doesn't exist).

As for the success log - I don't mind adding it, but looking at the other fetchers in case I was able to fetch file/kube/process I don't log anything, so what is the difference here. I get your point, but I think that it can cause a massive amount of logs, if we will logs all these small events.

resources/fetchers/process_fetcher_test.go
ofiriro3: opppppp, I saw it during my self review and I forgot to fix it - will do

oren-zohar (changes requested):

resources/fetchers/process_factory.go
```diff
@@ -2,22 +2,29 @@ package fetchers
...
+
+type ProcessesConfigMap map[string]ProcessInputConfiguration
+
 type ProcessFactory struct {
 }
 
 func init() {
 	manager.Factories.ListFetcherFactory(ProcessType, &ProcessFactory{})
 	gob.Register(ProcessResource{})
+	gob.Register(map[string]interface{}{})
```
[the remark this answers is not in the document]
ofiriro3: The usage of gob was removed - so I removed this code anyway

resources/fetchers/process_fetcher.go
```diff
@@ -19,36 +32,99 @@ type ProcessesFetcher struct {
 
 type ProcessFetcherConfig struct {
 	fetching.BaseFetcherConfig
-	Directory string `config:"directory"` // parent directory of target procfs
+	Directory         string `config:"directory"` // parent directory of target procfs
+	Fs                fs.FS
```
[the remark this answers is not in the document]
amirbenun: It is OK that some fetcher is dependent on the file system and also on the configuration, that's why we have the factory, so we can create each fetcher in a unique way.
I just mean that `FS` should not a part of the configuration. Configuration structs, in general, are more likely to be composed of low-level data types, and having `FS` in that context feels wrong.

oren-zohar (approved):

resources/fetchers/process_fetcher.go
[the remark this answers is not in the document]
amirbenun: 1. Regarding the case that we couldn't find the regex, as we both know, a configuration file is just one way of configuring a process, another way will be using command line arguments. In case our user uses cli args I don't see a point in endlessly printing error logs.
2. Regarding the success flow, I just want us to be able somehow to distinguish between the multiple running modes that the process fetcher has: with or without the configuration file. So the two info logs should cover that requirement.

resources/fetchers/process_factory.go
```diff
@@ -2,22 +2,29 @@ package fetchers
...
 
 const (
 	ProcessType = "process"
 )
 
+type ProcessInputConfiguration struct {
+	CommandArguments []string `config:"cmd-arguments"`
+}
+
+type ProcessesConfigMap map[string]ProcessInputConfiguration
```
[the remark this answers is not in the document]
amirbenun: I think that our current design inherently separates the two layers:
We have the `common.Config` that represents the configuration itself and the yaml file behind it.
And each fetcher can define what are the configuration arguments that belong to it. In our case, `ProcessInputConfiguration` is defined by the fetcher, and the factory is in charge of creating that fetcher based on the received yaml.

resources/fetchers/process_fetcher.go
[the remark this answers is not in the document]
ofiriro3: Got your point

go.mod
```diff
@@ -126,12 +127,11 @@ require (
 	github.com/eapache/go-resiliency v1.2.0 // indirect
 	github.com/elastic/elastic-agent-client/v7 v7.0.0-20210727140539-f0905d9377f6 // indirect
 	github.com/elastic/go-concert v0.2.0 // indirect
-	github.com/elastic/go-licenser v0.4.0 // indirect
```
oren-zohar, line 129: we'll need this one

oren-zohar (changes requested):

resources/fetchers/process_fetcher.go
```diff
@@ -19,53 +19,139 @@ package fetchers
...
+	configMap := make(map[string]interface{}, 0)
+	for _, argument := range processConfig.ConfigFileArguments {
+		// The regex extracts the cmd line flag(argument) value
+		regex := fmt.Sprintf(CMDArgumentMatcher, argument)
+		matcher := regexp.MustCompile(regex)
+		if !matcher.MatchString(cmd) {
+			logp.L().Infof("couldn't find a configuration file associated with flag %s for process %s", argument, processName)
+			continue
+		}
+		argValue := matcher.FindStringSubmatch(cmd)[1]
```
amirbenun, line 118: As a cautious step, I would make sure the array length >= 2.

amirbenun (approved):
Nice, another small issue

oren-zohar (approved):

PR #23, opened 2022-03-15
csp-2274 config fetchers

amirbenun:


_meta/config/cloudbeat.common.yml.tmpl
```diff
@@ -1,7 +1,12 @@
 cloudbeat:
   # Defines how often an event is sent to the output
   period: 30s
-  files: [
+  fetchers:
```
eyalkraft, line 4: Nice!
The file-system fetcher config struct seem to match this already but it needs to be validated somehow (manual test/ the better solution will be a simple test on `parseConfigFetchers` with the default configuration and a validation of the created parsed fetchers and their config).

eyalkraft (approved):
Beautiful!
Ideally next time two separate PR's (rename+refactor first, then rebase the original one)

DaveSys911 (approved):
LGTM

eyalkraft (approved):
Nice!

PR #33, opened 2022-03-30
Csp 2274 aws config fetcher no ctx creator

ofiriro3:
- Closes https://github.com/elastic/security-team/issues/3384


go.mod
```diff
@@ -53,7 +53,7 @@ require (
 	github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21 // indirect
 	github.com/eapache/queue v1.1.0 // indirect
 	github.com/elastic/beats/v7 v7.0.0-alpha2.0.20220328120630-c721abcf0650
```
ofiriro3, line 55: Once the latest csp changes will be merged - needs to replace it with the new `main` commit SHA/release

go.mod
```diff
@@ -157,6 +157,7 @@ require (
 	github.com/jonboulle/clockwork v0.2.2 // indirect
 	github.com/josephspurrier/goversioninfo v0.0.0-20190209210621-63e6d1acd3dd // indirect
 	github.com/karrick/godirwalk v1.15.6 // indirect
+	github.com/kisielk/godepgraph v0.0.0-20190626013829-57a7e4a651a9 // indirect
```
ofiriro3, line 160: Why did I remove it?

deploy/eks/cloudbeat-ds.yaml
```diff
@@ -0,0 +1,331 @@
+# Deploy a cloudbeat instance per node for cis policy
```
ofiriro3, line 1: Do we want to refine this file as part of this PR ?

DaveSys911: @ofiriro3 I think it's best to have one yaml.
I'd add the env vars added for eks to the existing cloudbeat yaml file.

JUSTFILE
```diff
@@ -6,24 +6,52 @@ install-kind:
...
-
-build-deploy-cloudbeat: build-cloudbeat load-cloudbeat-image deploy-cloudbeat
+  kubectl apply -f deploy/k8s/cloudbeat-ds.yaml -n kube-system
 
 build-cloudbeat-debug:
   GOOS=linux CGO_ENABLED=0 go build -gcflags "all=-N -l" && docker build -f Dockerfile.debug -t cloudbeat .
 
 deploy-cloudbeat-debug:
-  kubectl delete -f deploy/k8s/cloudbeat-ds-debug.yaml -n kube-system & kubectl apply -f deploy/k8s/cloudbeat-ds-debug.yaml -n kube-system
+   kubectl apply -f deploy/k8s/cloudbeat-ds-debug.yaml -n kube-system
```
DaveSys911, line 28: a few points here:
- as mentioned in the previous comment, I think there shouldn't be a separation between vanilla & eks yaml files so the same thing goes for the just commands.
- I think login-aws & push-ecr can be unified to one command
- U shouldn't use the latest tag to push to ecr, since you'll override what master pushed(I suggest using branch/pr ref for the tag instead).
- I'd unify them and add a [conditional](https://github.com/casey/just#conditional-expressions) to the just recipe . for exp `if env_var("EKS") == "true"` then invoke the EKS relevant step(publish-ecr) instead of kind-load-image.

DaveSys911 (commented):
Added 2 comments to address

deploy/eks/cloudbeat-ds.yaml
ofiriro3: I find it much easier to have a single YAML for each cloud provider. 

If you will pay close attention you would see that there are some differences. 
The fetchers that are being spawned are different as for some of the yaml file currently commented out because AWS won't allow you to do certain operations (like cluster-role that is being managed by AWS)

ofiriro3 (commented):
Ready for review @amirbenun 

If you have better names for the interfaces please share it with me

resources/providers/kubernetes_provider_mock.go
```diff
@@ -0,0 +1,70 @@
+// Code generated by mockery v2.10.4. DO NOT EDIT.
```
amirbenun, line 1: I didn't realize it uses `codegen` underlying.
We should understand what are the best practices for triggering the code generation in our SDLC.
Meaning, do we want to automate it or keep it a manual process?

resources/providers/aws/elb_provider.go
```diff
@@ -15,16 +15,21 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package fetchers
+package aws
```
amirbenun, line 18: A general note: `aws` package name is confusing, maybe `awslib` instead?

resources/fetchers/ecr_factory.go
```diff
@@ -0,0 +1,74 @@
...
+const (
+	ECRType = "aws-ecr"
+)
+
+func init() {
+	awsConfigProvider := aws.ConfigProvider{}
+	awsConfig := awsConfigProvider.GetConfig()
+	ecr := aws.NewEcrProvider(awsConfig.Config)
+	identityProvider := aws.NewAWSIdentityProvider(awsConfig.Config)
+	kubeGetter := providers.KubernetesProvider{}
```
amirbenun, line 24: General note, I think we should try pushing all of these preparation lines of code into the `Create` method.
I realized it may mess up the tests though it makes more sense.
Since the test is on the same package as the factory, you can practically do whatever you want from within the test.

amirbenun (changes requested):
Nice job!
I left some comments, the general notes should apply to all different factories/fetchers.
I agree the naming is hard, I understand what you tried to do with the naming.
Maybe we can just remove the ecr/elb... prefix from the describers? WDYT?

resources/providers/aws/elb_provider.go
ofiriro3: I struggled a lot with finding the best name for the package.

I just followed the beats convention at the end - the beats creates `aws` package for all the `aws` content.

But to be honest, I think that `awslib` is better, I hope that it won't confuse anyone.

go.mod
```diff
@@ -130,7 +130,6 @@ require (
 	github.com/dustin/go-humanize v1.0.0 // indirect
 	github.com/eapache/go-resiliency v1.2.0 // indirect
 	github.com/elastic/elastic-agent-client/v7 v7.0.0-20210727140539-f0905d9377f6 // indirect
```
ofiriro3, line 132: Do we need agent libs ?

resources/fetchers/ecr_factory.go
ofiriro3: So as we discussed I have made some research.

In general, go guidelines propose:
- No package-level variables
- No func init

Supporting articles 
https://leighmcculloch.com/posts/tool-go-check-no-globals-no-inits/
https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html

When diving deep a little bit deeper:
The only"approved" cases when init is legit are:
1. Registration of a package (Similar to the factory registration) 
2. Setup and initialization of specific package 
3. Variables initialization if cannot be done in initialization expr
4. running one-time computations

https://www.digitalocean.com/community/tutorials/understanding-init-in-go
https://medium.com/golangspec/init-functions-in-go-eac191b3860a
https://www.developer.com/languages/inti-function-golang/


Factory - 
```go 
func init() {
	Register("postgres", NewPostgreSQLDataStore)
	Register("memory", NewMemoryDataStore)
}
```
https://matthewbrown.io/2016/01/23/factory-pattern-in-golang

ofiriro3: So After my research, I have been wondering about few things:

1. What are the benefits of having `init()` function in each and every factory?
I guess that the obvious answer would be - because then I can customize each init (for each factory).
But then the question, why would we want to customize each factory?

We could for example - create only one init() function in the abstract factory and register all of the factories together. 
There is no mutual agreement on the right usage of the init(), but in general, the "godfathers" of the language suggest using it just in only extreme cases.

Implementing it this way can help us achieve more readability and reduce the usage of "magic" in our code.
BTW - we don't need to use init() at all and we can just put all of the register functions in the factory CTOR.
(Better explicit than implicit - Zen of python).

2. Should we put more logic in the init function?

So this is a very tricky question if looking at the first bullet - probably not because we shouldn't use init at all - or at least reduce that magic it does.

Implementation-wise, we will need to move the current init functionality to the "CreateForm" function.
But the thing that bothers me here - "CreateFrom" should be a private function. It feels a little bit smelly to test a private function (or at least a function that should be made private).

On the other hand, keeping the current solution where init() creates the factory is extending the "black magic" in our code, and will probably result in less readability.

ofiriro3: @amirbenun WDT?

resources/providers/kubernetes_provider_mock.go
ofiriro3: I added new code guidelines in the Readme.md file

resources/fetchers/ecr_factory.go
ofiriro3: As we spoke with @yashtewari - changes were made

resources/fetchers/iam_factory.go
```diff
@@ -0,0 +1,63 @@
...
+type IAMFactory struct {
+	extraElements func() (IAMExtraElements, error)
+}
+
+type IAMExtraElements struct {
+	iamProvider awslib.IAMRolePermissionGetter
+}
+
+func (f *IAMFactory) Create(c *common.Config) (fetching.Fetcher, error) {
+	cfg := IAMFetcherConfig{}
```
amirbenun, line 31: Add info log on factory creations (all the factories that are part of this the pr)

DaveSys911 (approved):
Approved please add a Todo in justfile & eks/cloudbeat-ds.yml
"Refactor with https://kustomize.io/ "

resources/fetchers/eks_factory.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/elastic/cloudbeat/resources/manager"
+)
+
+const (
+	EKSType = "aws-eks"
+)
+
+func init() {
+	manager.Factories.ListFetcherFactory(EKSType, &EKSFactory{
+		extraElements: getEksExtraElements,
```
amirbenun, line 17: I don't really like this pattern but let's try to think of a better way to load it on the next iterations

amirbenun (commented):
Nice, a few more before we merge

README.md
```diff
@@ -66,3 +67,52 @@ Use your favorite IDE to connect to the debugger on `localhost:40000` (for examp
...
+Notes
+- Place the test in the same package as the code it meant to test
+- File name should be aligned with the convention `original_file_mock`. For example: ecr_provider -> ecr_provider_mock
+
+Command example:
+```
+mockery --name=<interface_name> --with-expecter  --case underscore  --inpackage --recursive
+```
+#### General guidelines
+- The test file name should be
```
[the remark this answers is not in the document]
ofiriro3: Sorry, I missed it - was removed

go.mod
ofiriro3: Will do in other commit

go.mod
ofiriro3: fixed

amirbenun (approved):

PR #39, opened 2022-04-06
New-tests-ci

DaveSys911:


gurevichdmitry (approved):
LGTM

oren-zohar (changes requested):
integration tests are failing, can you please check?

DaveSys911:
> integration tests are failing, can you please check?

It's to be expected, there was a change to the helm files which this workflow won't support.
It's removed in this pr and cloudbeat-v2 will take it's place

gurevichdmitry (approved):

Dockerfile
```diff
@@ -1,10 +1,12 @@
 FROM debian
-RUN apt-get update
-RUN apt-get install -y ca-certificates
-RUN apt-get install -y curl
-RUN apt-get install -y iputils-ping
-RUN apt-get install -y jq
+RUN set -x && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
+      ca-certificates curl iputils-ping jq && \
```
oren-zohar, line 5: I think we can remove JQ since we don't use it in the integration test anymore, right? what about the other packages, are they still needed?

oren-zohar (changes requested):
Looks great

oren-zohar:
@gurevichdmitry It seems like the findings examples are not part of the new CI report, this is a regression. Is it possible to add it to the PR comment / Allure report?

DaveSys911:
Added to allure report, to comment would be messy
image

Dockerfile
DaveSys911: Let's leave them, this is a dev-only docker - these will be useful for debugging.

oren-zohar (approved):
Looks good, just need to add a findings link to the summary in the following PR @gurevichdmitry

PR #40, opened 2022-04-11
Pytest-symlink

DaveSys911:
Added kustomize & deploy/cloudbeat-pytest.yml for functionality tests
moved addopts to pyproject.toml


oren-zohar:
Why do we need more YML files that contain mostly the same information as the other YML files? Can we avoid it?

PR #45, opened 2022-04-14
Granting cloudbeat permissions for Kubernetes resources

jeniawhite:
Cloudbeat debug deployment YAML had incorrect permissions to perform Kubernetes queries and because of that `kube-api` wasn't running.

I've added the required permissions and roles in order for it to perform the expected tasks.

Fixes https://github.com/elastic/security-team/issues/3421


DaveSys911 (approved):
LGTM

PR #60, opened 2022-04-20
refactor cloudbeat metadata index flow

oren-zohar:
**What this PR change:**
- replace reporting cycle status to `.logs-cis_kubernetes_benchmark.metadata` index by debug level logging
- create cloudbeat logger
____
- Closes: https://github.com/elastic/security-team/issues/3670

### DIFF
--- a/beater/cloudbeat.go
+++ b/beater/cloudbeat.go
@@ -20,12 +20,9 @@ package beater
 import (
 	"context"
 	"fmt"
-	"time"
-
 	"github.com/elastic/cloudbeat/evaluator"
 
 	"github.com/elastic/beats/v7/libbeat/beat"
-	libevents "github.com/elastic/beats/v7/libbeat/beat/events"
 	"github.com/elastic/beats/v7/libbeat/common"
 	"github.com/elastic/beats/v7/libbeat/logp"
 	"github.com/elastic/beats/v7/libbeat/processors"
@@ -47,15 +44,13 @@ type cloudbeat struct {
 	data        *manager.Data
 	evaluator   evaluator.Evaluator
 	transformer transformer.Transformer
+	log         *logp.Logger
 }
-
-const (
-	cycleStatusStart = "start"
-	cycleStatusEnd   = "end"
-)
 
 // New creates an instance of cloudbeat.
 func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
+	log := logp.NewLogger("cloudbeat")
+
 	ctx, cancel := context.WithCancel(context.Background())
 
 	c := config.DefaultConfig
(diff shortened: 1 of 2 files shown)


beater/cloudbeat.go
```diff
@@ -47,15 +44,13 @@ type cloudbeat struct {
...
 }
 
-const (
-	cycleStatusStart = "start"
-	cycleStatusEnd   = "end"
-)
-
 // New creates an instance of cloudbeat.
 func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
+	log := logp.NewLogger("cloudbeat")
```
ofiriro3, line 52: I am wondering why did you decide to use "new logger here".

I tried to explore and research logging in general and logging good practices in Go.
As far as I read, I didn't see any objection to using one global log, unless there is a specific reason.

I tried to also to find some examples in the "beats" Repo and I couldn't find any other similar example (where the beat creates a new logger this way).

Most of our cloudbeat uses the following:
`logp.L().Info("some message")`

(which refers to a global log).

Why do you think there is a need to creates a new logger here ?

config/config.go
```diff
@@ -30,7 +30,6 @@ import (
 const DefaultNamespace = "default"
 
 const ResultsDatastreamIndexPrefix = "logs-cis_kubernetes_benchmark.findings"
-const MetadataDatastreamIndexPrefix = ".logs-cis_kubernetes_benchmark.metadata"
```
ofiriro3, line 33: Bye bye :)

beater/cloudbeat.go
```diff
@@ -141,19 +137,17 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
 			return nil
 		case fetchedResources := <-output:
 			cycleId, _ := uuid.NewV4()
-			// update hidden-index that the beat's cycle has started
-			bt.updateCycleStatus(cycleId, cycleStatusStart)
+			bt.log.Debugf("Cycle % has started", cycleId)
```
ofiriro3, line 140: 🥇

ofiriro3 (changes requested):
Fast and accurate PR ! Didn't have a Jedi imoji :/ 

One small question/comment

beater/cloudbeat.go
oren-zohar: this is about following go practices, but following elastics. see [here](https://github.com/elastic/beats/blob/eeda98b82f776d4f9104f2fad3958bdf291e8f59/x-pack/osquerybeat/beater/osquerybeat.go#L74
) for example

Creating a logger in this manner will allow us to filter its log by the key in Kibana. [Here](https://github.com/elastic/beats/blob/eeda98b82f776d4f9104f2fad3958bdf291e8f59/libbeat/logp/logger.go#L227) they mention that using .L() will create an unnamed logger, as we see in the logs in Kibana.

ofiriro3 (approved):

PR #61, opened 2022-04-25
Consume dataYaml in config, add config testing

yashtewari:
The `dataYaml` var specifies which CIS rules are "active", hence defining the scope of work to be done by the beat.

### DIFF
--- a/beater/cloudbeat.go
+++ b/beater/cloudbeat.go
@@ -40,12 +40,13 @@ type cloudbeat struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
-	config      config.Config
-	client      beat.Client
-	data        *manager.Data
-	evaluator   evaluator.Evaluator
-	transformer transformer.Transformer
-	log         *logp.Logger
+	config        config.Config
+	configUpdates <-chan *common.Config
+	client        beat.Client
+	data          *manager.Data
+	evaluator     evaluator.Evaluator
+	transformer   transformer.Transformer
+	log           *logp.Logger
 }
 
 // New creates an instance of cloudbeat.
@@ -54,8 +55,8 @@ func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
 
 	ctx, cancel := context.WithCancel(context.Background())
 
-	c := config.DefaultConfig
-	if err := cfg.Unpack(&c); err != nil {
+	c, err := config.New(cfg)
+	if err != nil {
 		cancel()
 		return nil, fmt.Errorf("error reading config file: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -37,12 +37,28 @@ type Config struct {
...
+			CISK8S []string `config:"cis_k8s"`
+		} `config:"activated_rules"`
+	} `config:"data_yaml"`
 }
 
 var DefaultConfig = Config{
 	Period: 10 * time.Second,
 }
 
+func New(cfg *common.Config) (Config, error) {
```
amirbenun, line 52: Not sure where it's used (besides tests)

config/config.go
```diff
@@ -37,12 +37,28 @@ type Config struct {
 	Period     time.Duration           `config:"period"`
 	Processors processors.PluginConfig `config:"processors"`
 	Fetchers   []*common.Config        `config:"fetchers"`
+
+	DataYaml struct {
```
amirbenun, line 41: Define a new struct?

amirbenun (commented):
Only two small comments

config/config.go
yashtewari: Had forgotten to add those changes to the commit, fixed

config/config.go
yashtewari: Done

beater/cloudbeat.go
```diff
@@ -132,6 +134,12 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
 		select {
 		case <-bt.ctx.Done():
 			return nil
+
+		case update := <-bt.configUpdates:
```
eyalkraft, line 138: Maybe this is enough for now, but I think theres a design problem with having this channel watched here.
I'm thinking of the next step - updating the bundle we use OPA to eval with.
The evaluator was already created, as part of `New`, As were the fetchers and the manager.
How will we change their config from this scope? Shouldn't we have some sort of cloudbeat factory here that will restart everything with the new config?
Let me know what you think

config/watcher.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"context"
+
+	"github.com/elastic/beats/v7/libbeat/common"
+	"github.com/elastic/beats/v7/libbeat/common/reload"
+	"github.com/elastic/beats/v7/libbeat/logp"
+)
+
+type reloader struct {
```
amirbenun, line 15: Where is it being instanciated? It feels like black magic

beater/cloudbeat.go
amirbenun: I agree with @eyalkraft  on this one. If we created a new cloudbeat for every reconfigure it would make a lot of sense.
Though there is a product impact on such a change:
After reconfiguring cloudbeat, when do we run the next fetchers iteration?
1. Based on the old configuration timeout
2. Based on the new configuration timeout
3. Right away, and then - based on the new configuration timeout

Creating a new cloudbeat instance probably means we choose (3) which I think is fine, just need to make sure that is accepted by everyone.

yashtewari: Thanks for your comments @eyalkraft, @amirbenun 

You're correct; other beats have different ways of implementing what ultimately becomes a restart of the `beater` interface. As discussed in our sync, this is just a fast, hacky way to get to our goals for 8.3. I considered going for the big architecture change, and we do want to get there eventually, but it didn't seem safe at this time partly because we don't have a truly "end to end" integration testing framework yet.

I've confirmed that this PR as it is serves our purpose for now. It's not a PR to reconfigure the beat, although it takes a step in that direction. It's just a PR to use config updates to get the `dataYaml`, while entirely ignoring all the other contents of the config update.

config/watcher.go
yashtewari: https://github.com/elastic/cloudbeat/blob/5d858d58c6b9d6750918f7ad22b9e3df226ad752/config/updates.go#L56

amirbenun (approved):
Approved as a short term solution

PR #62, opened 2022-04-25
Kube Fetcher logs + set Interval

oren-zohar:
**what this PR do**
- adds more fetchers logs
- set kube-api interval to 30s (hard-coded)


resources/fetchers/kube_provider.go
```diff
@@ -58,6 +59,7 @@ func GetKubeData(watchers []kubernetes.Watcher) []fetching.Resource {
 		}
 	}
 
+	logp.L().Info("Fetched: %v", ret)
```
eyalkraft, line 62: That's a lot of data for info isn't it?

eyalkraft (approved):

PR #64, opened 2022-04-26
fixed permissions required for cloudbeat k8s

DaveSys911:
Fixed permissions set for cloudbeat
image

pr opened to [elastic-package](https://github.com/elastic/elastic-package/pull/800) as well


deploy/k8s/standalone-agent.yml
```diff
@@ -7,45 +7,422 @@ metadata:
...
-        level: debug
+        meta:
+          package:
+            name: cis_kubernetes_benchmark
+            version: 0.0.1
+        data_stream:
+          namespace: default
+        streams:
+          - id: >-
+              cloudbeat-cis_kubernetes_benchmark.findings-e7db4bd9-4085-4775-aff6-7e6a46a23b88
```
oren-zohar, line 420: `cis_kubernetes_benchmark*` is deprecated, do you want to wait for the [migration](https://github.com/elastic/security-team/issues/3610) into `cloud_security_posture*`?

deploy/k8s/kustomize/base/cloudbeat-ds.yml
```diff
@@ -18,7 +18,7 @@ spec:
     spec:
```
oren-zohar, line 18: why do we need to change this file?

deploy/k8s/fleet-managed-agent.yml
```diff
@@ -173,17 +173,18 @@ rules:
       - pods
```
oren-zohar, line 173: where are we using this file?

deploy/k8s/standalone-agent.yml
```diff
@@ -7,45 +7,422 @@ metadata:
     k8s-app: elastic-agent
```
oren-zohar, line 7: I couldn't find usage for this file? is it for local testing? if so, how can I test it?

oren-zohar (changes requested):
not sure why we need to change anything in cloudbeat?

deploy/k8s/kustomize/base/cloudbeat-ds.yml
DaveSys911: Because we use it as part of our flow, and it should correspond.

deploy/k8s/fleet-managed-agent.yml
DaveSys911: We use it to deploy fleet managed agent, when we don't want to rely on elastic-package.

deploy/k8s/standalone-agent.yml
DaveSys911: It's exactly for that. You  can test it by running kubectl applying to your local cluster.

deploy/k8s/standalone-agent.yml
DaveSys911: A Todo with a gh issue link would be best

kfirpeled: Hey, the migration is being done at this moment. will be completed by the EOD.

amirbenun:
Looks good @DaveSys911.
Are these files copied from the elastic-agent repo?
I think it's good that we will be aligned with that.

oren-zohar (approved):
still need to understand how cloudbeat-ds worked without those permissions

DaveSys911:
Updated deployment yamls, unit tests failing https://github.com/elastic/cloudbeat/runs/6339307439?check_suite_focus=true merging since change isn't code related.

PR #68, opened 2022-04-27
Addition of common data layer to transform

jeniawhite:
The purpose of this layer is to collect common constant data about the system that will be used to populate/enhance the fetcher's results.

The layer populates its data upon the creation of the cloud beat instance prior to running all of the fetchers logic.

In the first drop, we are populating the cluster_id and node_id that will be used in order to build the resource_ids for the fetched resources.

Currently, this layer only supports the collection of data running on top of K8S (aligned to other data collection across cloud beats).

There is the first step and this collection layer might get moved closer to the fetcher's level.

Fixes https://github.com/elastic/security-team/issues/3532


transformer/common_data.go
```diff
@@ -0,0 +1,101 @@
...
+	n, err := client.CoreV1().Nodes().Get(context.Background(), hName, metav1.GetOptions{})
+	if err != nil {
+		logp.Error(fmt.Errorf("getClusterId error in Nodes get: %w", err))
+		return "", err
+	}
+	return string(n.ObjectMeta.UID), nil
+}
+
+func getHostName() (string, error) {
+	hName, err := ioutil.ReadFile("/etc/hostname")
```
uri-weisman, line 95: Plz store file path as a const.

transformer/events_creator.go
```diff
@@ -36,18 +36,27 @@ type Transformer struct {
...
+	cm 			  CommonData
 }
 
-func NewTransformer(ctx context.Context, eval evaluator.Evaluator, index string) Transformer {
+func NewTransformer(ctx context.Context, eval evaluator.Evaluator, index string) (Transformer, error) {
 	eventMetadata := common.MapStr{libevents.FieldMetaIndex: index}
 	events := make([]beat.Event, 0)
+	cm := NewCommonData()
+
+	err := cm.fetchCommonData()
```
uri-weisman, line 47: Should we pass a context as well? instead of generating a new one in `common_data.go`.

transformer/common_data.go
```diff
@@ -0,0 +1,101 @@
...
+	return string(n.ObjectMeta.UID), nil
+}
+
+func getNodeId() (string, error) {
+	hName, err := getHostName()
+	if err != nil {
+		logp.Error(fmt.Errorf("getNodeId error in getHostName: %w", err))
+		return "", err
+	}
+	client, err := kubernetes.GetKubernetesClient("", kubernetes.KubeClientOptions{})
```
uri-weisman, line 81: IMHO, might be better to reuse the k8s client.

uri-weisman (changes requested):
Good job Jenia 💪
I've added some minor comments.
BTW You probably need to modify the Transformer unit tests as well.

transformer/common_data.go
```diff
@@ -0,0 +1,104 @@
...
+		clusterId: "",
+		nodeId: "",
+	}
+}
+
+// TODO: Support environments besides K8S
+func (c *CommonData) fetchCommonData() error {
+	client, err := kubernetes.GetKubernetesClient("", kubernetes.KubeClientOptions{})
+	if err != nil {
+		logp.Error(fmt.Errorf("fetchCommonData error in GetKubernetesClient: %w", err))
```
eyalkraft, line 55: fyi our logging is a mess 😢 
https://github.com/elastic/security-team/issues/3020#issuecomment-1112003868

transformer/common_data.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		logp.Error(fmt.Errorf("fetchCommonData error in getNodeId: %w", err))
+		return err
+	}
+	c.nodeId = nodeId
+	return nil
+}
+
+func (c *CommonData) getClusterId() (string, error) {
+	n, err := c.kubeClient.CoreV1().Namespaces().Get(c.context, "kube-system", metav1.GetOptions{})
```
eyalkraft, line 75: "kube-system" should be a constant

transformer/events_creator.go
```diff
@@ -36,18 +36,27 @@ type Transformer struct {
 	eval          evaluator.Evaluator
 	eventMetadata common.MapStr
 	events        []beat.Event
+	cm 			  CommonData
```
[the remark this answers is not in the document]
eyalkraft: also weird indentation

eyalkraft (changes requested):
please ignore the comments that I added on the old version of the code

transformer/common_data.go
```diff
@@ -0,0 +1,104 @@
...
+	context context.Context
+	kubeClient k8s.Interface
+	clusterId string
+	nodeId string
+}
+
+const hostNamePath = "/etc/hostname"
+
+// TODO: Consider moving this layer to be custom for every resource type
+func NewCommonData(ctx context.Context) CommonData {
```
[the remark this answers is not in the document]
jeniawhite: @eyalkraft 
I agree with the single k8s client (unless there is an actual reason to use multiple due to concurrency or connection pool limitations).
Do you think that we should tackle the k8s client design change for fetchers/factories in this PR as part of this topic?
In the meanwhile, I've refactored and created a client that re-uses the cloudbeat's provided kubeconfig.

transformer/common_data_provider.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+const ( 
+	hostNamePath = "/etc/"
+	hostNameFile = "hostname"
+	namespace = "kube-system"
+)
+
+func NewCommonDataProvider(cfg config.Config) (CommonDataProvider, error) {
+	KubeClient, err := providers.KubernetesProvider{}.GetClient(cfg.KubeConfig, kubernetes.KubeClientOptions{})
```
eyalkraft, line 41: out of scope but probably we'll have to do it in the future (also in the `kube_fetcher/factory`) -
get also the `KubeClientOptions` from the config (or the whole client from outside).

transformer/events_creator.go
```diff
@@ -32,22 +32,30 @@ import (
...
 type Transformer struct {
-	context       context.Context
-	eval          evaluator.Evaluator
-	eventMetadata common.MapStr
-	events        []beat.Event
+	context       	   context.Context
+	eval          	   evaluator.Evaluator
+	eventMetadata 	   common.MapStr
+	events        	   []beat.Event
+	commonData 		   CommonDataInterface
```
eyalkraft, line 39: indentation?

transformer/common_data_provider.go
```diff
@@ -0,0 +1,101 @@
...
+		return CommonDataProvider{}, err
+	}
+
+	return CommonDataProvider{
+		kubeClient: KubeClient,
+		fsys: os.DirFS(hostNamePath),
+	}, nil
+}
+
+// TODO: Support environments besides K8S
```
eyalkraft, line 53:
```suggestion
// Note: As of today Kubernetes is the only environment supported by CommonDataProvider
```

transformer/events_creator.go
```diff
@@ -32,22 +32,30 @@ import (
...
 }
 
-func NewTransformer(ctx context.Context, eval evaluator.Evaluator, index string) Transformer {
+func NewTransformer(ctx context.Context, eval evaluator.Evaluator, cdProvider CommonDataProvider, index string) (Transformer, error) {
```
eyalkraft, line 42: Sorry for changing my mind, but now seeing that:
1. the cdp is only used once
2. this changes `NewTransformer` which is sort of a constructor to a possibly failing function

maybe we pass the common data directly? and use the provider in `beater/cloudbeat.go > New`

transformer/events_creator_test.go
```diff
@@ -157,7 +165,48 @@ func (s *EventsCreatorTestSuite) TestTransformer_ProcessAggregatedResources() {
...
+				fsys: fstest.MapFS{
+					"hostname": {
+						Data: []byte("testing_node"),
+					},
+				},
+			}
+
+			transformer, err := NewTransformer(ctx, &s.mockedEvaluator, cdp, testIndex)
+			s.NoError(err)
+
```
eyalkraft, line 209: what about checking the relevant field in the for loop of this test?
or in a separate test?

eyalkraft (changes requested):
Nice! last mile

eyalkraft (approved):

uri-weisman (approved):

PR #69, opened 2022-05-01
Refactor and enrich cloudbeat events

uri-weisman:
Refactor and add a name, sub-type, and type to the resource struct.


resources/fetchers/file_system_fetcher.go
```diff
@@ -104,7 +107,7 @@ func FromFileInfo(info os.FileInfo, path string) (FileSystemResource, error) {
 	usr, _ := user.LookupId(u)
 	group, _ := user.LookupGroupId(g)
 	mod := strconv.FormatUint(uint64(info.Mode().Perm()), 8)
-	inode := strconv.FormatUint(uint64(stat.Ino), 10)
```
uri-weisman, line 107: unnecessary conversion - already an uint64

resources/fetchers/kube_provider.go
```diff
@@ -62,19 +65,43 @@ func GetKubeData(watchers []kubernetes.Watcher) []fetching.Resource {
...
 
-	uid := metadata.UID
-	return string(uid), nil
+	return metadata
 }
 
-func (r K8sResource) GetData() interface{} {
-	return r.Data
+func (r K8sResource) GetSubType() string {
+	k8sObj := reflect.Indirect(reflect.ValueOf(r.Data))
```
eyalkraft, line 97: try and redesign to avoid calling `reflect.Indirect(reflect.ValueOf(r.Data))` twice

transformer/events_creator.go
```diff
@@ -50,30 +50,26 @@ func NewTransformer(ctx context.Context, eval evaluator.Evaluator, index string)
...
+		resourceMetadata := result.GetMetadata()
+		if err := c.createBeatEvents(result, resourceMetadata, cycleMetadata); err != nil {
+			logp.Error(fmt.Errorf("failed to create beat events for Cycle ID: %v, Resource Metadata: %v, Error: %v",
+				cycleMetadata.CycleId, resourceMetadata, err))
 		}
 	}
 }
 
-func (c *Transformer) createBeatEvents(fetchedResource fetching.Resource, metadata ResourceMetadata) error {
+func (c *Transformer) createBeatEvents(fetchedResource fetching.Resource, metadata fetching.ResourceMetadata, cycleMetadata CycleMetadata) error {
```
eyalkraft, line 72: now since `fetchedResource` itself has `.GetMetadata()` why do you pass both?

transformer/events_creator.go
```diff
@@ -88,10 +84,9 @@ func (c *Transformer) createBeatEvents(fetchedResource fetching.Resource, metada
 	}
 
 	timestamp := time.Now()
-	resource := ResourceFields{
-		ID:   metadata.ResourceId,
-		Type: metadata.Type,
-		Raw:  fetcherResult.Resource,
+	resource := fetching.ResourceFields{
+		ResourceMetadata: metadata,
+		Raw:              fetcherResult.Resource,
```
eyalkraft, line 89: type and subtype?
or are they waiting for another PR?

eyalkraft (changes requested):

resources/fetchers/iam_fetcher.go
```diff
@@ -60,3 +60,13 @@ func (r IAMResource) GetID() (string, error) {
 func (r IAMResource) GetData() interface{} {
 	return r.Data
 }
+
```
jeniawhite, line 63: You did not remove `GetID()` in `iam_fetcher.go`

resources/fetchers/kube_provider.go
```diff
@@ -62,19 +65,43 @@ func GetKubeData(watchers []kubernetes.Watcher) []fetching.Resource {
...
 	}
 
-	uid := metadata.UID
-	return string(uid), nil
+	return metadata
 }
 
-func (r K8sResource) GetData() interface{} {
-	return r.Data
+func (r K8sResource) GetSubType() string {
```
jeniawhite, line 96: Is there a reason not to fail and continue with an empty string value instead of the kind?
Just wondering if we are interested in sending data without kind that might create problems down the line of the pipeline that expects to groupBy or do any logic on these values?

transformer/events_creator.go
uri-weisman: No, they are part of the `ResourceMetadata` type.
plz see `transformer.go`

resources/fetchers/iam_fetcher.go
uri-weisman: Thanks 💪

resources/fetchers/kube_provider.go
eyalkraft: good question👆🏼
I think i'd prefer events with subtype missing than no events at all,
but the possibility for problems in Kibana resulting from this field missing is interesting...

uri-weisman: I was thinking about it...
I guess in the UI we can filter those events or just present them last than not sending them.
@eyalkraft, wdyt?

resources/fetchers/ecr_fetcher.go
```diff
@@ -94,10 +94,16 @@ func (f *ECRFetcher) getAwsPodRepositories(ctx context.Context) ([]string, error
...
-// GetID TODO: Add resource id logic to all AWS resources
-func (res ECRResource) GetID() (string, error) {
-	return "", nil
-}
 func (res ECRResource) GetData() interface{} {
 	return res
 }
+
+func (res ECRResource) GetMetadata() fetching.ResourceMetadata {
+	//TODO implement me
```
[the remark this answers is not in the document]
uri-weisman: Will do

resources/fetchers/kube_provider.go
eyalkraft: I think sending with empty subtype matches the current design pattern and is a good start (best effort).
for example currently [if a fetcher registration fails](https://github.com/elastic/cloudbeat/blob/25c12c588ec6e4841255aa3afa7aeeac7655ac56/resources/manager/factory.go#L69-L71) cloudbeat will log an error and continue.
anyways it's a good idea to ask @kfirpeled / @ari-aviran or someone else from the FE, or even comment [here](https://github.com/elastic/security-team/issues/3681).

eyalkraft: Additionally, except for bugs I don't see any reason for this method ever failing so I wouldn't waste too much time thinking/debating it.

kfirpeled: Thats a good point, 
In case of an issue - we can define it to be empty. And we would know to treat empty values as error.

What happens when there's no subtype? For example, in process. 
Just to make things easier, I think it is better to have value we would like to see. And not solve this in the UI only or assume if it is process there's no error and it is empty. Just so we can allow our users to built their own dashboard over the data without having some sort of logic.

resources/fetchers/process_fetcher.go
```diff
@@ -36,7 +36,9 @@ const (
 	// Expects format as the following: --<key><delimiter><value>.
 	// For example: --config=a.json
 	// The regex supports two delimiters "=" and ""
-	CMDArgumentMatcher = "\\b%s[\\s=]\\/?(\\S+)"
+	CMDArgumentMatcher  = "\\b%s[\\s=]\\/?(\\S+)"
+	ProcessResourceType = "process"
+	ProcessSubType      = "process"
```
kfirpeled, line 41: 💯

resources/fetchers/ecr_fetcher.go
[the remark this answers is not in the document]
uri-weisman: A task for implementing the missing `GetMetadata` method - https://github.com/elastic/cloudbeat/issues/71

resources/fetchers/kube_provider.go
```diff
@@ -62,19 +65,42 @@ func GetKubeData(watchers []kubernetes.Watcher) []fetching.Resource {
...
-	uid := metadata.UID
-	return string(uid), nil
+	return metadata
 }
 
-func (r K8sResource) GetData() interface{} {
-	return r.Data
+func getK8sSubType(k8sObj reflect.Value) string {
+	typeMeta, ok := k8sObj.FieldByName(k8sTypeMetadataField).Interface().(metav1.TypeMeta)
+	if !ok {
```
amitkanfer, line 98: now i see what you meant...

eyalkraft (approved):

Issue #73, opened 2022-05-03
Addition of platform dependent common utils


jeniawhite:
Motivation

While working with `info.Sys().(*syscall.Stat_t)`, I've encountered a problem when attempting to pull create time data.
Due to this issue https://github.com/golang/go/issues/29393, the data is stored in different keys (dependent on the platform).
When we run locally it should be accessed with `stat.Ctimespec` but inside the container `stat.Ctim`.

We can use the conditional compilation capability of go build https://github.com/golang/go/issues/35107.

Introduce a util layer that is being conditionally compiled with variations for different platforms and then re-use it inside go source code, without any consideration logic in the code to the platform.

Definition of done

We should be able to run on different platforms (development local env, and production env) and pull the data that is needed.

Please follow the following checklist at the beginning of your work, please comment with a suggested high-level solution. It should include:

- [ ]  Comment describing high level implementation details
- [ ]  Include API and data models
- [ ]  Include assumptions being taken
- [ ]  Provide backward/forward compatibility when changing data model schemas and key constants
- [ ]  Mention relevant individuals with a reason (getting feedback, FYI etc)
- [ ]  Submit a PR for our [technical index](https://github.com/elastic/security-team/blob/main/docs/cloud-security-posture-team/Technical_Index.md) that includes breaking changes/ new features before closing this ticket.

PR #74, opened 2022-05-03
File fetcher id change

jeniawhite:
Due to the gap https://github.com/elastic/cloudbeat/issues/73, decided to have the `path` without `creation date`.
As discussed with @eyalkraft and @tehilashn this means that this PR isn't interested in the physical file, but only in its path for uniqueness. 

I shall note that the path is mounted to the pod and on the host machine the location may be different.

Fixes https://github.com/elastic/security-team/issues/3599
Gaps https://github.com/elastic/cloudbeat/issues/73
Gaps https://github.com/elastic/cloudbeat/issues/78


eyalkraft (approved):

tehilashn:
Thanks @jeniawhite , as we discussed - let's create a separate task to address the physical file so we can prioritize it later.

PR #79, opened 2022-05-03
Add bug report template

gurevichdmitry:


DaveSys911:
> @DaveSys911 looks like to many checks are running for this one. Worth adding a task to improve that going forward, with relevant filters

100% agreed.We are kinda in "CI Limbo". I am waiting for onboarding to Buildkite CI so we can deprecate the current GH/Jenkins combo.

PR #86, opened 2022-05-09
Send dataYaml updates to bundle

yashtewari:


beater/cloudbeat.go
```diff
@@ -153,9 +155,35 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
...
+			}
+
+			if len(bt.config.Streams) == 0 {
+				logp.L().Infof("Did not receive any input stream, skipping.")
+				break
+			}
+
+			// TODO(yashtewari): Figure out the scenarios in which the integration sends
+			// multiple input streams. Since only one instance of our integration is allowed per
+			// agent policy, is it even possible that multiple input streams are received?
```
amirbenun, line 173: As far as I understand it's up to us to define what is being sent. We could as well change the Stream to be a singular instead of an array. Aren't we?

amirbenun (approved):
Looks good

beater/cloudbeat.go
```diff
@@ -153,9 +155,35 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
 
 		case update := <-bt.configUpdates:
 			if err := bt.config.Update(update); err != nil {
-				logp.L().Errorf("could not update cloudbeat config: %v", err)
+				logp.L().Errorf("Could not update cloudbeat config: %v", err)
+			}
+
+			policies, err := csppolicies.CISKubernetes()
+			if err != nil {
+				logp.L().Errorf("Could not load CIS Kubernetes policies: %v", err)
```
oren-zohar, line 163: this should throw an error, no? is there any reason we would want to continue with the rest of the function execution?

beater/cloudbeat.go
oren-zohar, line 173: this looks like it could be an issue instead

go.mod
```diff
@@ -54,7 +54,7 @@ require (
 	github.com/eapache/go-xerial-snappy v0.0.0-20180814174437-776d5712da21 // indirect
 	github.com/eapache/queue v1.1.0 // indirect
 	github.com/elastic/beats/v7 v7.0.0-alpha2.0.20220413140705-d101ba1d2ae5
-	github.com/elastic/csp-security-policies v0.0.16-go-lib
+	github.com/elastic/csp-security-policies v1.0.0
```
oren-zohar, line 57: nice

oren-zohar (changes requested):

beater/cloudbeat.go
yashtewari: I didn't find a way for it to be singular. On the integration end, we just define what to send, and the default structure is that it comes as one input in a list of streams. This would make sense to accommodate a wider variety of integration use cases.

beater/cloudbeat.go
yashtewari: Nice catch, true for all the errors here too. Fixed.

oren-zohar (approved):
logging and then using beak is not something I used to see, why not throw the error? also, the todo is still there.

PR #143, opened 2022-05-13
Fetcher sync implementation

amirbenun:
resolves https://github.com/elastic/security-team/issues/3414


resources/manager/data.go
```diff
@@ -43,15 +44,21 @@ type update struct {
...
+type fetcherResult struct {
+	resources []fetching.Resource
+	err       error
+}
+
 type ResourceMap map[string][]fetching.Resource
 
 // NewData returns a new Data instance with the given interval.
-func NewData(interval time.Duration, fetchers FetchersRegistry) (*Data, error) {
+func NewData(interval time.Duration, timeout time.Duration, fetchers FetchersRegistry) (*Data, error) {
```
amitkanfer, line 55: worth documentation what is `interval` and what is `timeout`

beater/cloudbeat.go
```diff
@@ -71,7 +72,7 @@ func New(b *beat.Beat, cfg *common.Config) (beat.Beater, error) {
 		return nil, err
 	}
 
-	data, err := manager.NewData(c.Period, fetchersRegistry)
+	data, err := manager.NewData(c.Period, time.Minute, fetchersRegistry)
```
amirbenun, line 75: On the next iteration, this should come from the yaml config

resources/manager/data.go
```diff
@@ -19,39 +19,42 @@ package manager
...
 
 // Data maintains a cache that is updated by Fetcher implementations registered
 // against it. It sends the cache to an output channel at the defined interval.
 type Data struct {
-	interval  time.Duration
-	output    chan ResourceMap
-	cycleData ResourceMap
-	fetchers  FetchersRegistry
-	wg        *sync.WaitGroup
+	timeout  time.Duration
```
amirbenun, line 33: Consider renaming the struct to `Manager`

oren-zohar: `Manger` is widely used in the agent, maybe something more specific?

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
-			}
+		case <-time.After(d.interval):
+			d.fetchIteration(ctx)
+		}
+	}
+}
+
+func (d *Data) fetchIteration(ctx context.Context) {
+	logp.L().Infof("manager trigger fetching using %d fetchers", len(d.fetchers.Keys()))
+	d.wg = &sync.WaitGroup{}
```
eyalkraft, line 91: kind of dangerous if this function is ever called in parallel.

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
-			return
+	ctx, cancel := context.WithTimeout(ctx, d.timeout)
+	defer cancel()
 
-		case <-ticker.C:
-			// Generate input ID?
-			// Send aggregated data at cycle tick
-			d.output <- d.cycleData
-			d.cycleData = make(ResourceMap)
+	result := make(chan fetcherResult, 1)
```
eyalkraft, line 125: Worth documenting that the buffer is required to avoid go-routine leaks in the case of timeout

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
 }
 
-func (d *Data) fetchWorker(ctx context.Context, updates chan update, k string) {
+func (d *Data) fetchAndSleep(ctx context.Context) {
+	// Happens once in a lifetime of cloudbeat and then enters the loop
+	d.fetchIteration(ctx)
 	for {
 		select {
+		case <-d.stop:
+			logp.L().Errorf("fetchers manager stopped")
```
eyalkraft, line 78: This is not an Error log

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
+			return
 		case <-ctx.Done():
+			logp.L().Errorf("fetchers manager canceled")
```
eyalkraft, line 81: This is not an Error log

beater/cloudbeat.go
```diff
@@ -211,7 +212,8 @@ func InitRegistry(c config.Config) (manager.FetchersRegistry, error) {
 
 // Stop stops cloudbeat.
 func (bt *cloudbeat) Stop() {
-	bt.data.Stop(bt.ctx, bt.cancel)
+	bt.cancel()
+	bt.data.Stop(bt.ctx)
```
eyalkraft, line 216: None of our `Stop` functions really cares about the `ctx` it receives so it has no effect but what you're doing here is calling a function with an already canceled context which would normally mean this called function won't do anything (since it's canceled).

eyalkraft: Maybe we should just remove the `ctx` from all the `Stop` functions.
Most fetchers don't do anything when `Stop` is called.. We just need to figure out the kube fetcher watchers stopping.

eyalkraft (changes requested):
Nice.
Let's understand what's the minimum that has to be solved now

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
-		}(key)
-	}
-
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		d.fetchManager(ctx, updates)
-	}()
-
+	go d.fetchAndSleep(ctx)
```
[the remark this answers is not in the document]
amirbenun: I don't think it should apply to the manager though, the `Beater` implementation in our source code is the one in `cloudbeat.go`

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
+// fetchProtected protect the fetching goroutine from getting panic
+func (d *Data) fetchProtected(ctx context.Context, k string) (val []fetching.Resource, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("fetcher %s recovered from panic: %v", k, r)
+		}
+	}()
+
+	val, err = d.fetchers.Run(ctx, k)
+	return
```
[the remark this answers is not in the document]
amirbenun: When recovering from panic it is inevitable

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
-	cancel()
+// fetchProtected protect the fetching goroutine from getting panic
+func (d *Data) fetchProtected(ctx context.Context, k string) (val []fetching.Resource, err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("fetcher %s recovered from panic: %v", k, r)
+		}
+	}()
+
+	val, err = d.fetchers.Run(ctx, k)
```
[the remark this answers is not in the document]
amirbenun: 1. Calling `Run` and `ShouldRun` is read-only on the map so concurrency is not really required.
2. I absolutely agree with you. Right now, our fetchers don't do anything but reading content from disk/memory. Of course, it's going to get changed in the future, and then it's up to the fetcher implementation to be context aware.

resources/manager/data.go
amirbenun: You are right, I added a comment.
You think there is something else should be done here?

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
-			val, err := d.fetchers.Run(ctx, k)
+	for _, key := range d.fetchers.Keys() {
+		d.wg.Add(1)
+		go func(k string) {
+			defer d.wg.Done()
+			val, err := d.fetchSingle(ctx, k)
 			if err != nil {
-				logp.L().Errorf("error running fetcher for key %q: %v", k, err)
+				logp.L().Errorf("error running fetcher for key %s: %v", k, err)
+			} else {
```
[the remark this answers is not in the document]
amirbenun: Nice catch

resources/manager/data.go
```diff
@@ -62,77 +65,95 @@ func (d *Data) Output() <-chan ResourceMap {
...
-			updates <- update{k, val}
-		}
-		// Go to sleep in each iteration.
-		time.Sleep(d.interval)
+		}(key)
 	}
+
+	d.wg.Wait()
+	logp.L().Infof("manager finished waiting and sending data after %d msecs", time.Since(start).Milliseconds())
+	d.output <- cycleData
```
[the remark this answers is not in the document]
amirbenun: There are some tickets already about the processing flow (maybe we should process resources in an async matter as well).
I don't think it will be smart to get into it on this pr, as long there are no degradations as part of my pr.
WDYT?

beater/cloudbeat.go
amirbenun: So I removed it from `data.Stop` though the evaluator (line 217) indeed requires the context.

resources/manager/data.go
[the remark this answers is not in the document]
eyalkraft: True, my bad

resources/manager/data.go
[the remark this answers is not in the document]
eyalkraft: 1. 👌 I was misled by a stackoverflow [comment](https://stackoverflow.com/a/38140573) but looking in the [docs](https://go.dev/doc/faq#atomic_maps):
> Map access is unsafe only when updates are occurring. As long as all goroutines are only reading—looking up elements in the map, including iterating through it using a for range loop—and not changing the map by assigning to elements or doing deletions, it is safe for them to access the map concurrently without synchronization.

2. 👌

resources/manager/data.go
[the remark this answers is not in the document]
eyalkraft: I'm with you

resources/manager/data.go
eyalkraft: Since this will change many times more a comment is sufficient for now.

eyalkraft (approved):

Issue #83, opened 2022-05-03
[BUG] Cloudbeat crashes when k8s node name differs from hostname in /etc/hosts


DaveSys911:
**Describe the bug**
As result of merging of pr #68 , where hostID is crafted from reading `/etc/hostname`

**Preconditions**
Environment state and/or platform specific configurations.

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. 

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: 
 - Browser: 
 - Kibana Version: 
 - Endpoint Version:
 - Other Version: 

**Additional context**

eyalkraft:
Thanks @DaveSys911!
What should be the behavior in this case?
How can we tell on which node we're running?
Can we find out on which pod we're running and get it from the pod information?

jeniawhite:
@DaveSys911 You are breaking the rules :D
<img width="677" alt="Screen Shot 2022-05-03 at 10 15 08" src="https://user-images.githubusercontent.com/12246907/166632531-b4a4a308-8d61-4e5d-888a-66195098d5ca.png">

@eyalkraft 
I will derive that from the pod, I'm pretty sure it is possible

DaveSys911:
@jeniawhite Good one,
 This doesn't imply hostname override rather that k8s API node name &a node's hostname might not be identical.
You can get the node name using the k8s downward API ,consuming the NODE_NAME env var from within the pod
<img width="714" alt="image" src="https://user-images.githubusercontent.com/16253938/167297874-50a4fb2d-3819-4616-a4a6-7f675fd8f5ae.png">

jeniawhite:
@DaveSys911 
Saw that as another option as well.

I've browsed a bit and saw this code:
https://github.com/elastic/beats/blob/master/libbeat/common/kubernetes/util.go#L122
I'm more interested in the discover in Kubernetes cluster part here:
https://github.com/elastic/beats/blob/master/libbeat/common/kubernetes/util.go#L157

What do you think about adding this logic to the code and avoiding the `NODE_NAME` env var?
Or we can obviously add the `NODE_NAME` var as well and run the full discovery with the rollback to `NODE_NAME`.

@eyalkraft any thoughts as well?

DaveSys911:
[...]

PR #146, opened 2022-05-15
Common data provider node name resolution

jeniawhite:
Switched to using libbeat internal utils in order to resolve the node name.

TODO: Should we continue if the initialization of the common data layer has failed?
If yes, then what should we do for the ids and other future data that has not been collected.
How would we in the current state indicate that there is an error (assuming that nobody would look in the logs since everything will work and we cannot understand from the ids that there are problems)?

Fixes https://github.com/elastic/cloudbeat/issues/83


eyalkraft (approved):

DaveSys911 (approved):
Tested on Vanilla k8s, seems to be working
image

image

PR #149, opened 2022-05-16
Fix bug in Config.Update

yashtewari:
Incoming config updates with slices smaller than the current slice didn't resize the current slice, resulting in undesirable values in the tail of the slice.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -63,10 +63,32 @@ func New(cfg *common.Config) (Config, error) {
 	return c, nil
 }
 
+// Update replaces values of those keys in the current config which are
+// present in the incoming config.
+//
+// NOTE(yashtewari): This will be removed with the planned update to restart the
+// beat with the new config.
 func (c *Config) Update(cfg *common.Config) error {
 	if err := cfg.Unpack(&c); err != nil {
 		return err
 	}
+
+	// Check if the incoming config has streams.
+	m := make(map[string]interface{})
+	if err := cfg.Unpack(&m); err != nil {
+		return err
+	}
+
+	if _, ok := m["streams"]; !ok {
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -66,7 +67,92 @@ func (s *ConfigTestSuite) TestNew() {
...
+	s.NoError(err)
+
+	for _, test := range tests {
+		cfg, err := common.NewConfigFrom(test.update)
+		s.NoError(err)
+
+		err = c.Update(cfg)
+		s.NoError(err)
+
+		s.Equal(test.expected, c.Streams[0].DataYaml.ActivatedRules.CISK8S)
```
amirbenun, line 155: Let's make sure that other config fields don't get overridden by mistake

config/config.go
```diff
@@ -64,10 +64,23 @@ func New(cfg *common.Config) (Config, error) {
...
+	// Shorter slices in the incoming config won't resize the corresponding slice
+	// in the current config, leaving undesirable values in the tail of the slice.
+	// This needs to be handled.
+	uc, err := New(cfg)
+	if err != nil {
+		return err
+	}
+
+	for i := range c.Streams {
+		c.Streams[i].DataYaml.ActivatedRules.CISK8S = uc.Streams[i].DataYaml.ActivatedRules.CISK8S
```
amirbenun, line 81: I would add a test to cover that.
Maybe try to add a new stream as a final state in your tests?

config/config_test.go
yashtewari: Have added tests for that.

config/config.go
yashtewari: Changed the implementation.

oren-zohar (approved):

PR #153, opened 2022-05-16
Add csp task template

eyalkraft:


DaveSys911 (approved):
Lgtm

oren-zohar:
I would suggest marking some parts as hidden, like that:
```
<!-- This is hidden unless you edit. -->
```

People will probably ignore some parts of the template but won't delete them, so we will see a lot of text from the template that it not relevant :( we can just mark most of it as hidden in the first place

eyalkraft:
@oren-zohar which parts?

oren-zohar:
you do it for each part, or just the parts you think users will probably are going to ignore the most

PR #165, opened 2022-05-22
removing explicit checkout 

DaveSys911:
Remove stale code


oren-zohar (changes requested):
we can do it in the unit-test wf, no? https://github.com/elastic/cloudbeat/blob/ce6600f0429d9ed6142144b8bd3050d76696c800/.github/workflows/unit-test.yml#L25
https://github.com/elastic/cloudbeat/blob/ce6600f0429d9ed6142144b8bd3050d76696c800/.github/workflows/unit-test.yml#L63

oren-zohar (approved):

PR #171, opened 2022-05-24
Clean up and align logs.

yashtewari:
## Objectives

1. Remove all usages of the global libbeat logger [`logp.L()`](https://github.com/elastic/beats/blob/ea28c0419dc4ede9318c4b34a732ce11b03482b7/libbeat/logp/logger.go#L228), and instead use locally instantiated loggers with added context in the form of selectors and other added options.

3. Clean up log messages and add new logs where fit.
4. Fix bad logs. For example, [`logp.Error`](https://github.com/elastic/beats/blob/ea28c0419dc4ede9318c4b34a732ce11b03482b7/libbeat/logp/fields.go#L40) has been used in many places which isn't actually a logging function. Also, any usage of the logp library in the global scope, or in the `init()` phase, [doesn't yield any logs](https://github.com/elastic/beats/blob/ea28c0419dc4ede9318c4b34a732ce11b03482b7/libbeat/logp/logger.go#L44).
5. Enable [`logp.TestingSetup`](https://github.com/elastic/beats/blob/ea28c0419dc4ede9318c4b34a732ce11b03482b7/libbeat/logp/core.go#L170) for all test suites in cloudbeat. This means that logs can now be seen with maximum verbosity as part of test results when verbose logging (`go test -v`) is used.
---

### Global logger vs logger as a dependency of structs and functions

There was some consideration to use a global logger, which is mostly what we've already been doing with usage of `logp.L()`. To implement this with the libbeat [`logp.Logger`](https://github.com/elastic/beats/blob/ea28c0419dc4ede9318c4b34a732ce11b03482b7/libbeat/logp/logger.go#L30), but still get the benefit of added context over the default `logp.L()`, a global cloudbeat logger would have to be declared and wrapped in a getter with [`sync.Once`](https://pkg.go.dev/sync#Once) because [logp loggers can't be declared in the global scope](https://github.com/elastic/beats/blob/ea28c0419dc4ede9318c4b34a732ce11b03482b7/libbeat/logp/logger.go#L44).

This is doable although it does introduce the risk of a future PR initializing the logger in the global scope, thereby breaking all logs. This risk doesn't seem trivial to automatically detect/avoid.

For the benefits of allowing for hierarchical context in loggers, with child loggers adding more and more information to the logs, and for the ability to be able to pass alternative/mock loggers to functions in unit tests, this PR favors changing function signatures to explicitly reflect the logger dependency. This becomes similar to passing `ctx` around as needed.

See also: [global logging is considered an anti-pattern in the Go community](https://dave.cheney.net/2017/01/23/the-package-level-logger-anti-pattern)


resources/fetchers/file_system_factory.go
```diff
@@ -35,22 +35,23 @@ func init() {
 type FileSystemFactory struct {
 }
 
-func (f *FileSystemFactory) Create(c *common.Config) (fetching.Fetcher, error) {
+func (f *FileSystemFactory) Create(log *logp.Logger, c *common.Config) (fetching.Fetcher, error) {
```
amirbenun, line 38: Add a create debug log

resources/fetchers/kube_factory.go
```diff
@@ -32,23 +32,24 @@ func init() {
 	manager.Factories.ListFetcherFactory(fetching.KubeAPIType, &KubeFactory{})
 }
 
-func (f *KubeFactory) Create(c *common.Config) (fetching.Fetcher, error) {
+func (f *KubeFactory) Create(log *logp.Logger, c *common.Config) (fetching.Fetcher, error) {
```
amirbenun, line 35: Add log

resources/fetchers/process_factory.go
```diff
@@ -36,20 +37,21 @@ func init() {
 	manager.Factories.ListFetcherFactory(ProcessType, &ProcessFactory{})
 }
 
-func (f *ProcessFactory) Create(c *common.Config) (fetching.Fetcher, error) {
+func (f *ProcessFactory) Create(log *logp.Logger, c *common.Config) (fetching.Fetcher, error) {
```
amirbenun, line 40: Adda a log

resources/manager/factory.go
```diff
@@ -42,56 +43,56 @@ func newFactories() factories {
 func (fa *factories) ListFetcherFactory(name string, f fetching.Factory) {
 	_, ok := fa.m[name]
 	if ok {
-		logp.L().Warnf("fetcher %q factory method overwritten", name)
+		panic(fmt.Errorf("fetcher factory with name %q listed more than once", name))
```
amirbenun, line 46: Nice catch

amirbenun (commented):
Very nice, I like the convention of tracing the functions.
I wonder if there are some existing tools for golang that can do it for us.
Maybe it worth adding a task for it?

amirbenun (approved):

PR #190, opened 2022-05-30
Bump prominent logs from Debug to Info level, add new logs.

yashtewari:


transformer/events_creator.go
```diff
@@ -77,17 +78,20 @@ func (c *Transformer) createBeatEvents(fetchedResource fetching.Resource, cycleM
 	fetcherResult := fetching.Result{Type: resMetadata.Type, Resource: fetchedResource.GetData()}
 
 	result, err := c.eval.Decision(c.context, fetcherResult)
-
 	if err != nil {
 		c.log.Errorf("Error running the policy: %v", err)
 		return err
 	}
 
+	c.log.Debugf("Eval decision for input: %v -- %v", fetcherResult, result)
```
amirbenun, line 86: I think this one is going to spam us even on debug. Should we remove it?

beater/cloudbeat.go
```diff
@@ -154,47 +154,54 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
...
+			bt.log.Infof("Publishing %d events to index", len(events))
 			bt.client.PublishAll(events)
-			bt.log.Debugf("Cycle % has ended", cycleId)
+
+			bt.log.Infof("Cycle %v has ended", cycleId)
 		}
 	}
 }
 
+func (bt *cloudbeat) configUpdate(update *common.Config) error {
```
amirbenun, line 177: let's drop that function for now and revert to the original structure

transformer/events_creator.go
yashtewari: There is already an equally severe log at debug level from our `libbeat` dependency, which is the full content of each event published to ES. And in the ongoing error scenario, this is exactly the log which helped us to figure out what was going wrong. So I think we can keep it.

amirbenun (approved):
LGTM

PR #191, opened 2022-05-30
Update CI to support producing release candidate builds(aka "Staging")

DaveSys911:
Update ci to run staging builds.
Added drop down list to enable invoking manual Staging build("Release candidates") via an option in cloudbeat's jenkins job
https://internal-ci.elastic.co/job/cloudbeat/job/cloudbeat-mbp/
image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -150,7 +150,6 @@ pipeline {
           environment {
             PATH = "${env.PATH}:${env.WORKSPACE}/bin"
             HOME = "${env.WORKSPACE}"
-            // SNAPSHOT = "true"
           }
           when {
             beforeAgent true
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -32,6 +33,7 @@ pipeline {
     booleanParam(name: 'test_ci', defaultValue: false, description: 'Enable test')
     booleanParam(name: 'release_ci', defaultValue: true, description: 'Enable build the release packages')
     booleanParam(name: 'its_ci', defaultValue: true, description: 'Enable async ITs')
+    choice(name: 'Build_type', choices: ['snapshot', 'staging'], description: 'Choose Snapshot or Staging Build type(Default: Snapshot)')
```
amirbenun, line 36: seems like the casing convention should be `build_type` lowercase

amirbenun: Also seems like the description specifies what the default value is, but there is no `defaultValue` argument.

DaveSys911: @amirbenun fixed casing convetion.
choice param doesn't support `defaultValue`, instead it takes the first param as default.

amirbenun (approved):

PR #196, opened 2022-05-30
Make cloudbeat wait for initial reconfiguration from Fleet server.

yashtewari:
The reconfiguration logic makes cloudbeat align with the CIS rules enabled/disabled by the user.

Prior to this change, the "default" behavior would kick of and send findings for _all_ CIS rules to the index, before the Fleet server could tell it which findings to send. This is undesirable as it pollutes the index with unwanted findings.

### DIFF
--- a/beater/cloudbeat.go
+++ b/beater/cloudbeat.go
@@ -132,6 +132,7 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
 	}
 	defer b.Manager.Stop()
 
+	// Wait for Fleet-side reconfiguration only if cloudbeat is running in Agent-managed mode.
 	if b.Manager.Enabled() {
 		bt.log.Infof("Waiting for initial reconfiguration from Fleet server...")
 		update, err := bt.reconfigureWait(reconfigureWaitTimeout)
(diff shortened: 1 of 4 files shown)


beater/cloudbeat.go
```diff
@@ -37,6 +37,10 @@ import (
 	"github.com/gofrs/uuid"
 )
 
+const (
+	reconfigureWaitTimeout = 5 * time.Minute
```
amirbenun, line 41: Instead of using a const, I would get it from the config.
Then, you can add a unit test to assert on the expected behavior when it timed out.

beater/cloudbeat_test.go
```diff
@@ -0,0 +1,180 @@
...
+	type incomingConfigs []struct {
+		after  time.Duration
+		config *common.Config
+	}
+
+	testcases := []struct {
+		timeout  time.Duration
+		configs  incomingConfigs
+		expected *common.Config
+	}{
```
amirbenun, line 107: Let's add a test case where a valid configuration (for example `configWithDataYaml`) is followed by an invalid one (for example `configNoDataYaml`) and make sure that the valid config still applies.

amirbenun (approved):

amirbenun:
All the comments can be fixed on a different PR.

PR #167, opened 2022-05-22
Upgrade go verison to 1.18

uri-weisman:
Upgrade go version to latest - it will be merged after the 8.3.x branch is forked.

PR #201, opened 2022-06-01
Process cloudbeat events asynchronously 

uri-weisman:


beater/cloudbeat.go
```diff
@@ -174,20 +179,19 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
...
-			// TODO: send events through a channel and publish them by a configured threshold & time
-			events := bt.transformer.ProcessAggregatedResources(fetchedResources, cycleMetadata)
-
-			bt.log.Infof("Publishing %d events to index", len(events))
-			bt.client.PublishAll(events)
-
-			bt.log.Infof("Eval cycle %v published %d events after %s", cycleId, len(events), time.Since(cycleStart))
+		// Flush events to ES after a pre-defined interval, meant to clean residuals after a cycle is finished.
+		case <-time.Tick(flushInterval * time.Second):
+			logp.L().Infof("Publish cloudbeat events to elasticsearch after %d seconds", flushInterval)
```
eyalkraft, line 184: Let's handle the case of no events to publish (no log or calling to Publish). We don't want 4 hrs of this log while no actual events were sent.

beater/cloudbeat.go
```diff
@@ -174,20 +179,19 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
...
+			bt.client.PublishAll(eventsToSend)
+			eventsToSend = nil
```
eyalkraft, line 186: not sure this is what you want. appending on nil later will fail. what about using make(..)?

eyalkraft (commented):
Nice beginning but still each fetcher is fetching EVERYTHING before passing anything forward.
Let's say I configure my file fetcher to check thousands of files. I don't want to collect all of them and only them pass them forward.
This is of course irrelevant for the file fetcher but imagine we have a single fetcher for s3 buckets. we can't collect information for thousands of them before moving on.
In other words - we can't be waiting for `fetchSingle`. the design change needed is deeper.

Edit:
Rule examples from `CIS Amazon Web Services Foundations
Benchmark
v1.4.0 - 05-28-2021`:

> 2.1.1 Ensure all S3 buckets employ encryption-at-rest (Manual)

> 2.1.3 Ensure MFA Delete is enable on S3 buckets (Automated)

beater/cloudbeat.go
uri-weisman: Actually, we can still append and iterate a Nil Slice even though its value is nil.
https://go.dev/play/p/jm8p0ErGaqA

eyalkraft: cool!

beater/cloudbeat.go
```diff
@@ -164,7 +167,9 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
 	}
 
 	output := bt.data.Output()
+	go bt.transformer.ProcessAggregatedResources(bt.ctx, output)
```
uri-weisman, line 170: An Improvement will be to run a loop that spawns as many Go routines as we have CPU cores available.

uri-weisman:
> Nice beginning but still each fetcher is fetching EVERYTHING before passing anything forward. Let's say I configure my file fetcher to check thousands of files. I don't want to collect all of them and only them pass them forward. This is of course irrelevant for the file fetcher but imagine we have a single fetcher for s3 buckets. we can't collect information for thousands of them before moving on. In other words - we can't be waiting for `fetchSingle`. the design change needed is deeper.

Agreed, I've updated the PR to be E2E asynchronously, starting from fetchers till the events are being sent to ES.

eyalkraft (commented):
Partial review for now

beater/cloudbeat.go
```diff
@@ -287,3 +301,25 @@ func (bt *cloudbeat) Stop() {
 func (bt *cloudbeat) configureProcessors(processorsList processors.PluginConfig) (procs *processors.Processors, err error) {
 	return processors.New(processorsList)
 }
+
+func pipelineStep[In any, Out any](ctx context.Context, inputChannel chan In, fn func(context.Context, In) Out) chan Out {
```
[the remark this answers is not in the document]
uri-weisman: Yes, depends on:
- https://github.com/elastic/cloudbeat/pull/167

beater/cloudbeat.go
```diff
@@ -287,3 +301,25 @@ func (bt *cloudbeat) Stop() {
...
+
+		for s := range inputChannel {
+			select {
+			case <-ctx.Done():
+				break
+			default:
+			}
+
+			go func(ctx context.Context, s In) {
+				outputCh <- fn(ctx, s)
```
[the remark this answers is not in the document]
uri-weisman: It might be confusing but the function actually gets the concrete argument of type `In` and not `channel In`.
Do you still think this design needs to be modified?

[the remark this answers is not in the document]
eyalkraft: Hmm my mistake.
In this case it's closer to option 2.
wdyt? Is this worth moving to another module?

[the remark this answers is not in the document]
eyalkraft: And also now I see what you met with concurrency.
Current implementation starts n (2) go routines per resource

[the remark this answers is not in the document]
uri-weisman: Yeah, I'll create a new module.
It's not really related to the `beater` module.

beater/cloudbeat.go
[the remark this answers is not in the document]
uri-weisman: Very important but problematic to perform in this case as the Output channel will become the Input channel of the next step.

ofiriro3 (commented):
Nice job! few comments and questions

beater/cloudbeat.go
```diff
@@ -175,19 +179,29 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
...
-			bt.log.Infof("Publishing %d events to index", len(events))
-			bt.client.PublishAll(events)
+			if len(eventsToSend) < eventsThreshold {
+				continue
+			}
 
-			bt.log.Infof("Eval cycle %v published %d events after %s", cycleId, len(events), time.Since(cycleStart))
+			logp.L().Infof("Publish to elasticsearch - capacity reached to %d events", eventsThreshold)
+			bt.client.PublishAll(eventsToSend)
+			eventsToSend = nil
```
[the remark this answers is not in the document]
eyalkraft: related previous [comment](https://github.com/elastic/cloudbeat/pull/201#discussion_r886737218).

Both scenarios you described are safe.
A slice is a pointer (plus additional fields) to an underlying array. A slice as an argument passes by value, which means the pointer value is copied. The underlying array is only collected by the garbage collector after no more references exist. Appending to a nil slice seems to create a new underlying array.

https://go.dev/play/p/U8BGAi8Z8zI

resources/fetchers/kube_fetcher.go
```diff
@@ -145,7 +145,7 @@ func (f *KubeFetcher) initWatchers() error {
 	return nil
 }
 
-func (f *KubeFetcher) Fetch(ctx context.Context) ([]fetching.Resource, error) {
+func (f *KubeFetcher) Fetch(ctx context.Context, resCh chan<- fetching.ResourceInfo, cMetadata fetching.CycleMetadata) error {
```
eyalkraft, line 148: I think a cleaner approach could be:
1. On fetcher creation, give it the result channel
2. My next comment on registry

resources/manager/registry.go
```diff
@@ -90,13 +89,13 @@ func (r *fetchersRegistry) ShouldRun(key string) bool {
...
-func (r *fetchersRegistry) Run(ctx context.Context, key string) ([]fetching.Resource, error) {
+func (r *fetchersRegistry) Run(ctx context.Context, key string, resCh chan<- fetching.ResourceInfo, cycleMetadata fetching.CycleMetadata) error {
 	registered, ok := r.reg[key]
 	if !ok {
-		return nil, fmt.Errorf("fetcher %v not found", key)
+		return fmt.Errorf("fetcher %v not found", key)
 	}
 
-	return registered.f.Fetch(ctx)
+	return registered.f.Fetch(ctx, resCh, cycleMetadata)
```
eyalkraft, line 98: What if the fetchers had an in channel, that would provide a new cycle metadata whenever a cycle should start?
Registering then will be actually registering the in channels of fetchers, and here the cycle metadata would have been sent on each one of the registered in channels.
This incorporates the pipeline flow further.

WDYT?

Maybe still worth splitting our modules to

1. Simple Synchronous modules with no in/out channels but simple interfaces
2. Pipeline wrappers that handle the pipeline work

resources/fetchers/ecr_fetcher.go
```diff
@@ -56,21 +56,24 @@ type ECRResource struct {
 func (f *ECRFetcher) Stop() {
 }
 
-func (f *ECRFetcher) Fetch(ctx context.Context) ([]fetching.Resource, error) {
+func (f *ECRFetcher) Fetch(ctx context.Context, resCh chan<- fetching.ResourceInfo, cMetadata fetching.CycleMetadata) error {
```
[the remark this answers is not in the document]
uri-weisman: Thanks @ofiriro3, Might be better if the `registry` will pass the "resource channel" upon fetchers creation

uri-weisman:
>  how about making it dynamic? meaning, receiving an integer parameter that defines the amount of fetching that is allowed in parallel? "concurrency level" of some sort? For files it can be 0 (infinite) or some other very large number, and for buckets and such, could be 5 or 10 ...? We also don't want to be fully serialized, we might end up with very low utilization.

Nice idea @amitkanfer , can be dynamic or even a single configurable value that will be factored in each fetcher according to some pre-defined constant.
First, we want to make cloudbeat's data flow asynchronously and then the following tasks will support concurrency.

resources/fetching/fetcher.go
```diff
@@ -31,7 +32,7 @@ type Factory interface {
 
 // Fetcher represents a data fetcher.
 type Fetcher interface {
-	Fetch(context.Context) ([]Resource, error)
+	Fetch(context.Context, chan<- ResourceInfo, CycleMetadata) error
```
amirbenun, line 35: Honestly, I am not in favor of doing that change before we have more information about the requirement.
@oren-zohar is about to start working on an epic called "Cloudbeat performance" that will give us more knowledge on this.

Disadvantages of doing this premature optimization:
1. Changing a clear and simple API to a more complicated one.
2. Working with channels could make some trouble that we can't expect.
3. Right now, the channel size is 0, meaning every time we push some object into it, we can't push another one until the next step of the pipeline is pulling it. It will cause some kind of ping-pong of context switches between the sender and the receiver.

I think we can avoid the `fetcher` change on this PR, and deal with it in the future only if we understand it's inevitable.

amirbenun (commented):
Very interesting PR Uri!
In my opinion this PR should focus on the asynchronization of the latter steps in the pipeline (evaluating, shipping) and losing the part where a manager have to wait for all fetchers before we run these steps.

resources/fetching/fetcher.go
uri-weisman: Thanks, @amirbenun.
This is the first step of the pipeline, therefore it determines the pace for the rest of the steps.
Moreover, due to this addition, we don't need to wait for the manager synchronization- once a resource is collected it's pushed to the next step for processing.

> Right now, the channel size is 0, meaning every time we push some object into it, we can't push another one until the next step of the pipeline is pulling it. It will cause some kind of ping-pong of context switches between the sender and the receiver.

I agree, it might be better to use a buffered channel in order to not be blocked.

resources/manager/data.go
```diff
@@ -88,78 +75,72 @@ func (d *Data) fetchAndSleep(ctx context.Context) {
...
-				defer mu.Unlock()
-				cycleData[k] = val
 			}
 		}(key)
 	}
 
 	d.wg.Wait()
 	d.log.Infof("Manager finished waiting and sending data after %d milliseconds", time.Since(start).Milliseconds())
-	d.output <- cycleData
+	d.log.Infof("Cycle %s has ended", cycleId.String())
```
uri-weisman, line 103: The Cycle has not really finished, only the task of fetching all resources by `fetchers`.
I wonder If we're fine with sending events after logging this.

resources/fetching/fetcher.go
amirbenun: Thanks for the reply @uri-weisman 
> Moreover, due to this addition, we don't need to wait for the manager synchronization- once a resource is collected it's pushed to the next step for processing.

I didn't say the manager should wait for **all fetchers** to finish. All I am saying is the manager should wait for **a fetcher** to complete a cycle before it publishes its resources to the next pipeline's steps. Other fetchers won't be affected by a corrupted one.
It will keep a clear API and may save some trouble in the future.

uri-weisman: Since your comment I made a minor change in which the fetcher receives the `resource ch` upon creation and not as an argument, nevertheless, channels are still in use.

In the future, I guess we don't want to collect all resources and only then start processing (CSPM use case), Eyal commented about it as well - https://github.com/elastic/cloudbeat/pull/201#pullrequestreview-991871121

I understand your point regarding channels, and most definitely I'll analyze how system resources ( memory, cpu, etc ) have changed due to this PR.

amirbenun: Sounds good, anyway I think the process should be the opposite.
We should first have a benchmark of how cloudbeat affects the system, and only then we should think of how it should be improved. Pushing such a change must be backed up with numbers.

eyalkraft (changes requested):
Partial review, to be continued

1. Remove the binary `beater/bundle/tmpBundles/bundle.tar.gz`

eyalkraft:
check out @amirbenun's https://github.com/elastic/security-team/issues/3334

resources/fetchers/elb_fetcher.go
```diff
@@ -53,23 +54,25 @@ type ELBResource struct {
...
+	
 	for _, loadBalancer := range result {
-		results = append(results, ELBResource{LoadBalancersDescription(loadBalancer)})
+		f.resourceCh <- fetching.ResourceInfo{
+			Resource:      ELBResource{LoadBalancersDescription(loadBalancer)},
+			CycleMetadata: cMetadata,
+		}
 	}
-	return results, err
+	return err
```
eyalkraft, line 75: nil?

resources/manager/data_test.go
```diff
@@ -54,27 +65,53 @@ func (f *DelayFetcher) Stop() {
...
+//	return &DataFetcher{message, false, ch, wg}
+//}
+
+func (f *PanicFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
+	defer f.wg.Done()
 	panic(f.message)
 }
 
+//
+//func (f *DataFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
```
eyalkraft, line 94: ?

resources/manager/data_test.go
```diff
@@ -54,27 +65,53 @@ func (f *DelayFetcher) Stop() {
...
+//	wg         *sync.WaitGroup
+//}
+
+func newPanicFetcher(message string, ch chan fetching.ResourceInfo, wg *sync.WaitGroup) fetching.Fetcher {
+	return &PanicFetcher{message, false, ch, wg}
 }
 
-func (f *PanicFetcher) Fetch(ctx context.Context) ([]fetching.Resource, error) {
+//
+//func newDataFetcher(message string, ch chan fetching.ResourceInfo, wg *sync.WaitGroup) fetching.Fetcher {
```
eyalkraft, line 84: ?

resources/manager/data_test.go
```diff
@@ -54,27 +65,53 @@ func (f *DelayFetcher) Stop() {
 type PanicFetcher struct {
 	message    string
 	stopCalled bool
+	resourceCh chan fetching.ResourceInfo
+	wg         *sync.WaitGroup
 }
 
-func newPanicFetcher(message string) fetching.Fetcher {
-	return &PanicFetcher{message, false}
+//type DataFetcher struct {
```
eyalkraft, line 72: ?

beater/cloudbeat.go
```diff
@@ -33,12 +35,13 @@ import (
...
 	"github.com/elastic/elastic-agent-libs/logp"
-
-	"github.com/gofrs/uuid"
 )
 
 const (
 	reconfigureWaitTimeout = 10 * time.Minute
+	flushInterval          = 10 * time.Second
+	eventsThreshold        = 75
+	resourceChBuffer       = 50
```
eyalkraft, line 44:
```suggestion
	resourceChBuffer       = 10000
```
We don't want the writers to be blocked.

eyalkraft (approved):
LGTM! small few changes as comments.
Also thoughts for the future:

Thinking about how to decouple `Fetch` from `CycleMetadata` because the only thing happening with the cycle metadata is it's added to the elements written to the channel.

eyalkraft (approved):

uri-weisman:
> check out @amirbenun's [elastic/security-team#3334](https://github.com/elastic/security-team/issues/3334)

@amirbenun and I had a look at the `autodiscover`, and it seems that our cycle logic isn't compatible with the service.
This PR introduces a pipeline that is asynchronously within the cycle boundaries, meaning that we'll wait for all fetchers to finish before the next cycle is started.
Currently, it looks that the `autodiscover` service would not support it.

PR #202, opened 2022-06-01
More logs

yashtewari:


amitkanfer:
👑

ofiriro3 (approved):
Does it mean that cloudbeat will be blocked until it will get the right configuration ?

is this the expected behavior?

amirbenun (approved):

PR #207, opened 2022-06-06
Cloudbeat to support processing of k8s cert files 

uri-weisman:
FS fetcher to watch after k8s certificate & key files.
Supports CIS benchmark Rules 1.1.20 & 1.1.21.

### DIFF
--- a/tests/commonlib/io_utils.py
+++ b/tests/commonlib/io_utils.py
@@ -41,17 +41,8 @@ def get_k8s_yaml_objects(file_path: Path) -> list[str: dict]:
     """
     if not file_path:
         raise Exception(f'{file_path} is required')
-    result_list = []
     with file_path.open() as yaml_file:
-        yaml_objects = yaml.safe_load_all(yaml_file)
-        for yml_doc in yaml_objects:
-            if yml_doc:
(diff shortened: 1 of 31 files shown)


tests/commonlib/io_utils.py
```diff
@@ -104,7 +104,7 @@ def exec_command(container_name: str, command: str, param_value: str, resource:
                 f"File {resource} does not exist or mount missing.")
 
         if command == 'chmod':
-            os.chmod(path=resource, mode=int(param_value))
+            os.chmod(path=resource, mode=int(param_value, base=8))
```
uri-weisman, line 107: int() is removing the leading zero and as `os.chmod` is expecting an octal value it results with wrong permissions.
Example:
```
int('0664') = 644
os.chmod(filepath, 644)
```
Causing the file permissions to be:
`-w--wx--- (230)`

uri-weisman:
Related:
- https://github.com/elastic/csp-security-policies/pull/101

oren-zohar (approved):
💯 💯 💯 💯

DaveSys911 (approved):
LGTM

PR #209, opened 2022-06-08
Refactor bundle building

jeniawhite:
Apply changes to bundle building refactoring that was done in this PR:
- https://github.com/elastic/csp-security-policies/pull/102

- https://github.com/elastic/security-team/issues/3652


go.mod
```diff
@@ -214,6 +225,7 @@ replace (
 	github.com/cucumber/godog => github.com/cucumber/godog v0.8.1
 	github.com/dop251/goja => github.com/andrewkroh/goja v0.0.0-20190128172624-dd2ac4456e20
 	github.com/dop251/goja_nodejs => github.com/dop251/goja_nodejs v0.0.0-20171011081505-adff31b136e6
+	github.com/elastic/csp-security-policies => github.com/jeniawhite/csp-security-policies v1.1.26
```
jeniawhite, line 228: Needs to be removed and version bumped when PR is merged

beater/cloudbeat.go
```diff
@@ -257,7 +252,10 @@ func (bt *cloudbeat) configUpdate(update *common.Config) error {
 		return fmt.Errorf("could not marshal to YAML: %w", err)
 	}
 
-	if err := csppolicies.HostBundleWithDataYaml("bundle.tar.gz", policies, y); err != nil {
+	bundle := csppolicies.CISKubernetesBundle()
+	bundle.With("data.yaml", y)
```
eyalkraft, line 256: Will this work?
```suggestion
	bundle := csppolicies.CISKubernetesBundle().With("data.yaml", y)
```

jeniawhite: The method expects a pointer in order to modify
```
cannot call pointer method With on "github.com/elastic/csp-security-policies/bundle".Bundlecompiler[InvalidMethodExpr](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal?utm_source%3Dgopls#InvalidMethodExpr)
```
Attempted to ref
```
bundle := (&csppolicies.CISKubernetesBundle()).With("data.yaml", y)
```
And got
```
invalid operation: cannot take address of csppolicies.CISKubernetesBundle() (value of type "github.com/elastic/csp-security-policies/bundle".Bundle)compiler[UnaddressableOperand](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal?utm_source%3Dgopls#UnaddressableOperand)
```
I can refactor the CSP policies code or do workarounds here (which are probably worse than having two lines), but I think that this code will be removed in the near future and there isn't much value in that, what do you think? @eyalkraft

eyalkraft: yep, let's keep it that way.

go.mod
```diff
@@ -214,6 +225,7 @@ replace (
 	github.com/cucumber/godog => github.com/cucumber/godog v0.8.1
 	github.com/dop251/goja => github.com/andrewkroh/goja v0.0.0-20190128172624-dd2ac4456e20
 	github.com/dop251/goja_nodejs => github.com/dop251/goja_nodejs v0.0.0-20171011081505-adff31b136e6
+	github.com/elastic/csp-security-policies => github.com/jeniawhite/csp-security-policies v1.1.28
```
eyalkraft, line 228: This must be changed before merging

eyalkraft (approved):

PR #217, opened 2022-06-15
Modify FS resource fields

uri-weisman:
Update FS resource fields according to ECS standards.
- https://github.com/elastic/security-team/issues/3889


resources/fetchers/file_system_fetcher_test.go
```diff
@@ -19,6 +19,9 @@ package fetchers
```
uri-weisman, line 19: Most definitely this test file has to be table-driven to reduce the repetition of code.

uri-weisman:
Related:
- https://github.com/elastic/csp-security-policies/pull/103

uri-weisman:
Screen Shot 2022-06-15 at 13 56 42

resources/fetchers/file_system_fetcher_test.go
```diff
@@ -110,11 +134,17 @@ func TestFileFetcherFetchDirectoryOnly(t *testing.T) {
 	cfg := FileFetcherConfig{
 		Patterns: filePaths,
 	}
-	factory := FileSystemFactory{}
+
+	osUserMock := &utils.MockOSUser{}
+	osUserMock.EXPECT().GetUserNameFromID(mock.Anything, mock.Anything).Return("", errors.New("err"))
```
ofiriro3, line 139: Why do we need to use mock here?

resources/utils/user.go
```diff
@@ -65,7 +65,18 @@ type Group struct {
 // lineFunc returns a value, an error, or (nil, nil) to skip the row.
 type lineFunc func(line []byte) (v interface{}, err error)
 
-func GetUserNameFromID(uid uint32, userFilePath string) (string, error) {
+type OSUser interface {
+	GetUserNameFromID(uid uint32, userFilePath string) (string, error)
+	GetGroupNameFromID(gid uint32, groupFilePath string) (string, error)
+}
+
+type OSUserUtil struct{}
```
ofiriro3, line 73: Wondering about this struct.

1. The function is a `pure` function - for the same input, it will always return the same output.
2. The function has no side effects.
3. The struct is stateless.
4. We do not modify any variable after assigning a value.

It leads me to think, that a struct is not necessary here and we can use functional programming in this case.
It also will reduce the need to mock this class every time we test file-system and will use the real functionality instead.

WDT ?

resources/utils/user_test.go
```diff
@@ -39,18 +39,23 @@ const (
 )
 
 type UserTestSuite struct {
```
ofiriro3, line 41: GJ on adding those tests!

ofiriro3 (changes requested):
Looks good!

There are some small things I wish to discuss :)

eyalkraft:
@ofiriro3 Please include your comments written

resources/fetchers/file_system_fetcher_test.go
```diff
@@ -39,18 +42,27 @@ func TestFileFetcherFetchASingleFile(t *testing.T) {
 		Patterns: filePaths,
 	}
 
+	osUserMock := &utils.MockOSUser{}
+	osUserMock.EXPECT().GetUserNameFromID(mock.Anything, mock.Anything).Return("root", nil)
```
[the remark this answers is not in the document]
ofiriro3: Relevant to all of the mock instances

resources/utils/user.go
uri-weisman: After a discussion we had, I decided to keep the current implementation.
Mocking the util in the unit tests is worth the overhead of creating an interface for a utility.
Unit tests suppose to test the class functionality, as the utility has unit tests of its own, it should be mocked in this use case.

resources/fetchers/file_system_fetcher_test.go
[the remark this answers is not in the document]
ofiriro3: I think that we can safely assume that our code will never run on windows.
So can we use this one ?

Can we use the 
```
//go:build !windows
// +build !windows
```

Like in the following example? 
https://github.com/elastic/beats/pull/30808/files#diff-304eb22795a7b56d19543ff398610f0845f05e42eb606ddd8953545521fcea4e


It's not a must, I wonder if we use mock without asserting on input, the test will be less robust.
I mean, regardless of the input it will always return the same value.

I'll approve it regardless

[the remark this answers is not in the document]
uri-weisman: Thank you @ofiriro3.
It's not only windows, the `staff` group which is relevant here, might vary among different UNIX systems.

PR #220, opened 2022-06-16
Remedy problems in automated tests.

yashtewari:
Building on the great work done by multiple contributors to these tests, this change includes:

1. #### Fix broken tests
Except for the ones [commented out here](https://github.com/elastic/cloudbeat/blob/a558c5cd79e91ee5b54527269cb688fff521963d/tests/product/tests/tests/process/process_test_cases.py#L1307), which need further investigation. The disabled test cases fail and break the cluster, preventing any test cases thereafter to run correctly.

2. #### A single command to run all test targets in succession
```
just run-test-targets
```
For each test target, cleanup and recreate the cluster, run the test target, and collect its logs and reports.

3. #### New separate markers for each test target
Described in more detail [here](https://github.com/elastic/security-team/issues/4167#issuecomment-1160676832), these markers allow for the test targets to be run separately. This is important because the tests are not idempotent, so they can't all be run together.

4. #### A log collection script
```
just collect-logs <target>
```
The tests run in a pod in the K8S cluster, and make changes to the same cluster, including in manifests like `kube-apiserver.yaml`, `etcd.yaml`, `kube-scheduler.yaml`, `kube-controller-manager.yaml` which can render the cluster unavailable for some time. This would be OK if everything goes as expected and the cluster becomes healthy again, but if it doesn't, there would be no way to get the logs from inside the test pod to get some insight on what happened.

This script is a "best effort" collector which gets the logs as and when it can, so the files in the generated `test-logs` directory will be as latest as possible.

5. #### More assertions/failures to clearly indicate what is going wrong in a test
For example, the difference between: for a given rule, the evaluation not being as expected, or the evaluation not being present at all.
Screenshot 2022-06-28 at 8 38 24 AM

Implemented during https://github.com/elastic/security-team/issues/4167

fixes: https://github.com/elastic/security-team/issues/4167

### DIFF
--- a/JUSTFILE
+++ b/JUSTFILE
@@ -121,6 +121,7 @@ delete-local-helm-cluster:
 cleanup-create-local-helm-cluster target: delete-local-helm-cluster create-kind-cluster build-cloudbeat load-cloudbeat-image
   just deploy-local-tests-helm {{target}}
 
+# TODO(DaveSys911): Move scripts out of JUSTFILE: https://github.com/elastic/security-team/issues/4291
 test-pod-status:
   #!/usr/bin/env sh
 
--- a/JUSTFILE
+++ b/JUSTFILE
@@ -1,6 +1,6 @@
 # Refactor via kustomize https://kustomize.io/
 
-image-tag := `git branch --show-current`
+image_tag := `git branch --show-current`
 
 create-kind-cluster:
   kind create cluster --config deploy/k8s/kind/kind-config.yml
@@ -20,6 +20,7 @@ load-cloudbeat-image:
   kind load docker-image cloudbeat:latest --name kind-mono
 
 build-cloudbeat:
+  GOOS=linux go mod vendor
   GOOS=linux go build -v && docker build -t cloudbeat .
 
 deploy-cloudbeat:
@@ -43,7 +44,7 @@ delete-cloudbeat-debug:
 build-deploy-eks-cloudbeat: build-cloudbeat publish-image-to-ecr deploy-eks-cloudbeat
 
 publish-image-to-ecr:
-  aws ecr get-login-password --region us-east-2 | docker login --username AWS --password-stdin 704479110758.dkr.ecr.us-east-2.amazonaws.com && docker tag cloudbeat 704479110758.dkr.ecr.us-east-2.amazonaws.com/cloudbeat:{{image-tag}} && docker push 704479110758.dkr.ecr.us-east-2.amazonaws.com/cloudbeat:{{image-tag}}
+  aws ecr get-login-password --region us-east-2 | docker login --username AWS --password-stdin 704479110758.dkr.ecr.us-east-2.amazonaws.com && docker tag cloudbeat 704479110758.dkr.ecr.us-east-2.amazonaws.com/cloudbeat:{{image_tag}} && docker push 704479110758.dkr.ecr.us-east-2.amazonaws.com/cloudbeat:{{image_tag}}
 
 deploy-eks-cloudbeat:
   kubectl delete -f deploy/eks/cloudbeat-ds.yml -n kube-system & kubectl apply -f deploy/eks/cloudbeat-ds.yml -n kube-system
@@ -80,23 +81,28 @@ expose-ports:
 
 #### TESTS ####
 
-TESTS_RELEASE := "cloudbeat-tests"
-TIMEOUT := "1200s"
+TEST_POD := 'test-pod-v1'
+TESTS_RELEASE := 'cloudbeat-test'
+TEST_LOGS_DIRECTORY := 'test-logs'
+POD_STATUS_UNKNOWN := 'Unknown'
+POD_STATUS_PENDING := 'Pending'
+POD_STATUS_RUNNING := 'Running'
+TIMEOUT := '1200s'
 
 patch-cb-yml-tests:
   kubectl kustomize deploy/k8s/kustomize/test > tests/deploy/cloudbeat-pytest.yml
 
 build-pytest-docker:
-  cd tests; docker build -t cloudbeat-test .
+  cd tests; docker build -t {{TESTS_RELEASE}} .
 
 load-pytest-kind:
-  kind load docker-image cloudbeat-test:latest --name kind-mono
+  kind load docker-image {{TESTS_RELEASE}}:latest --name kind-mono
 
 deploy-tests-helm:
   helm upgrade --wait --timeout={{TIMEOUT}} --install --values tests/deploy/values/ci.yml --namespace kube-system {{TESTS_RELEASE}}  tests/deploy/k8s-cloudbeat-tests/
 
-deploy-local-tests-helm:
-  helm upgrade --wait --timeout={{TIMEOUT}} --install --values tests/deploy/values/local-host.yml --namespace kube-system {{TESTS_RELEASE}}  tests/deploy/k8s-cloudbeat-tests/
+deploy-local-tests-helm target:
+  helm upgrade --wait --timeout={{TIMEOUT}} --install --values tests/deploy/values/local-host.yml --set testData.marker={{target}} --namespace kube-system {{TESTS_RELEASE}}  tests/deploy/k8s-cloudbeat-tests/
 
 purge-tests:
 	helm del {{TESTS_RELEASE}} -n kube-system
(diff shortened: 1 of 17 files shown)


JUSTFILE
```diff
@@ -105,8 +110,71 @@ gen-report:
...
+
+test-pod-status:
+  #!/usr/bin/env sh
+
+  if [ ${STATUS=`kubectl get pod -n kube-system test-pod-v1 --template {{{{.status.phase}}`} ]; then
+    echo $STATUS
+  else
+    echo {{POD_STATUS_UNKNOWN}}
+  fi
+
```
gurevichdmitry, line 131: I think we need to avoid scripting inside justfile. The better way is to create helper utils lib and execute scripts from the lib.

JUSTFILE
```diff
@@ -80,23 +80,28 @@ expose-ports:
 
 #### TESTS ####
```
gurevichdmitry, line 81: TESTS command should be documented in tests/readme.md file

tests/commonlib/kubernetes.py
```diff
@@ -179,7 +187,83 @@ def delete_resources(self, resource_type: str, **kwargs):
     def patch_resources(self, resource_type: str, **kwargs):
         """
         """
-        return self.dispatch_patch[resource_type](**kwargs)
```
gurevichdmitry, line 182: minimal docstrings should be provides.
We are working with PEP8 standards, you can find more information [PEP8 - Documentation strings](https://peps.python.org/pep-0008/#documentation-strings)

tests/commonlib/kubernetes.py
```diff
@@ -179,7 +187,83 @@ def delete_resources(self, resource_type: str, **kwargs):
...
+            if not deleted:
+                raise ValueError(f'could not delete Pod: {kwargs}')
+
+            pod = self.create_patched_resource(resource_type, patch_body)
+
+            return pod
+
+        else:
+            return self.dispatch_patch[resource_type](**kwargs)
+
```
gurevichdmitry, line 206: we need to simplify logic because python depends on indentations and _if inside if_ is less readable.
in this case we can use 'early termination' technique:

`if resource__type != RESOURCE_POD:
    return self.dispatch_patch[resource_type](**kwargs)`

`patch_body = kwargs.pop('body')`
`...`

tests/commonlib/kubernetes.py
```diff
@@ -179,7 +187,83 @@ def delete_resources(self, resource_type: str, **kwargs):
...
+            pod = self.create_patched_resource(resource_type, patch_body)
+
+            return pod
+
+        else:
+            return self.dispatch_patch[resource_type](**kwargs)
+    
+    def create_patched_resource(self, patch_resource_type, patch_body):
+        file_path = Path(__file__).parent / '../deploy/mock-pod.yml'
```
gurevichdmitry, line 208: missing docstring

tests/commonlib/kubernetes.py
```diff
@@ -179,7 +187,83 @@ def delete_resources(self, resource_type: str, **kwargs):
...
+                else:
+                    body[i] = val
+            
+            if len(patch) > len(body):
+                body += val[len(patch):]
+                
+        else:
+            raise ValueError(f'Invalid body {body} of type {type(body)}')
+        
+        return body
```
gurevichdmitry, line 267: Very complicated function that includes many if else statements and recursion.
Is it possible to improve?

tests/commonlib/kubernetes.py
```diff
@@ -179,7 +187,83 @@ def delete_resources(self, resource_type: str, **kwargs):
...
 
     def list_resources(self, resource_type: str, **kwargs):
         """
```
gurevichdmitry, line 269: docstring is missing

tests/product/tests/test_kube_api_rules.py
```diff
@@ -10,31 +10,32 @@
...
-        *cis_5_2_5.values(),
-        *cis_5_2_6.values(),
+        # *cis_5_1_3.values(),
+        # *cis_5_1_5.values(),
+        # *cis_5_1_6.values(),
+        # *cis_5_2_2.values(),
+        # *cis_5_2_3.values(),
+        # *cis_5_2_4.values(),
+        # *cis_5_2_5.values(),
+        # *cis_5_2_6.values(),
```
gurevichdmitry, line 24: Please add also brief explanation why test cases are commented

tests/product/tests/test_kube_api_rules.py
```diff
@@ -10,31 +10,32 @@
...
         *cis_5_2_7.values(),
         # *cis_5_2_8.values(),
         # *cis_5_2_9.values(),
         # *cis_5_2_10.values(),
```
gurevichdmitry, line 28: Please add also brief explanation why test cases are commented

tests/product/tests/test_kube_api_rules.py
```diff
@@ -10,31 +10,32 @@
...
-        *cis_5_2_5.keys(),
-        *cis_5_2_6.keys(),
+        # *cis_5_1_3.keys(),
+        # *cis_5_1_5.keys(),
+        # *cis_5_1_6.keys(),
+        # *cis_5_2_2.keys(),
+        # *cis_5_2_3.keys(),
+        # *cis_5_2_4.keys(),
+        # *cis_5_2_5.keys(),
+        # *cis_5_2_6.keys(),
```
gurevichdmitry, line 38: Please add also brief explanation why test cases are commented

tests/product/tests/test_kube_api_rules.py
```diff
@@ -10,31 +10,32 @@
...
         *cis_5_2_7.keys(),
         # *cis_5_2_8.keys(),
         # *cis_5_2_9.keys(),
```
gurevichdmitry, line 41: Please add also brief explanation why test cases are commented

JUSTFILE
DaveSys911: @gurevichdmitry @yashtewari  please note we already have it.These two dirs `scripts/ `& `.ci/scripts` .
 If you could please either generate a task and comment a Todo to the justfile, or move the bahs logic into one of these two dirs.

tests/product/tests/test_kube_api_rules.py
yashtewari: That was just for debugging, updated now.

tests/product/tests/test_kube_api_rules.py
yashtewari: That was just for debugging, updated now.

tests/product/tests/test_kube_api_rules.py
yashtewari: That was just for debugging, updated now.

tests/product/tests/test_kube_api_rules.py
yashtewari: That was just for debugging, updated now.

tests/commonlib/kubernetes.py
```diff
@@ -179,7 +187,83 @@ def delete_resources(self, resource_type: str, **kwargs):
...
+                continue
+            
+            patched_body = self.patch_resource_body(yml_resource, patch_body)
+            created_resource = self.create_from_dict(patched_body, **relevant_metadata)
+
+            done = self.wait_for_resource(resource_type=resource_type, status_list=["RUNNING", "ADDED"], **relevant_metadata)
+            if done:
+                patched_resource = created_resource
+
+            break
```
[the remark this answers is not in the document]
yashtewari: It's correct! At this point it should stop trying whether or not the patch was successful.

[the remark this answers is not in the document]
gurevichdmitry: If you have several k8s_resources with relevant data it will stop after first "successful" iteration and other resources will not be patched. Logically, without lines 220-221, you starting **for** loop and finishes after first iteration.

oren-zohar:
fixes: https://github.com/elastic/security-team/issues/4167

JUSTFILE
DaveSys911: @yashtewari if u can comment a Todo with task number https://github.com/elastic/security-team/issues/4291

tests/commonlib/kubernetes.py
yashtewari: Discussed that this can be refactored at a later time @gurevichdmitry

gurevichdmitry: new [task](https://github.com/elastic/security-team/issues/4295) is created

PR #221, opened 2022-06-16
Pre merge cloudbeat integration

gurevichdmitry:
- remove old test files
- delete rule markers
- create new cloudbeat integration test


tests/product/tests/conftest.py
```diff
@@ -60,7 +47,7 @@ def clean_test_env(data):
             k8s_client.get_resource(resource_type=resource_type, **relevant_metadata)
             k8s_client.delete_resources(resource_type=resource_type, **relevant_metadata)
             deleted = k8s_client.wait_for_resource(resource_type=resource_type, status_list=["DELETED"], **relevant_metadata)
```
oren-zohar, line 49:
```suggestion
           k8s_client.wait_for_resource(resource_type=resource_type, status_list=["DELETED"], **relevant_metadata)
```

tests/product/tests/conftest.py
```diff
@@ -60,7 +47,7 @@ def clean_test_env(data):
...
-            print(f"{resource_type} deleted: {deleted}")
+            # print(f"{resource_type} deleted: {deleted}")
```
oren-zohar, line 50: delete

tests/integration/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+        k8s.delete_from_yaml(get_k8s_yaml_objects(file_path=file_path))
+    k8s.start_agent(yaml_file=file_path, namespace=cloudbeat_agent.namespace)
+    time.sleep(5)
+    yield k8s, api_client, cloudbeat_agent
+    k8s_yaml_list = get_k8s_yaml_objects(file_path=file_path)
+    k8s.delete_from_yaml(yaml_objects_list=k8s_yaml_list)  # stop agent
+
+
+@pytest.fixture(name='fixture_data')
+def fixture_data(start_stop, k8s, cloudbeat_agent):
```
oren-zohar, line 35: why use `start_stop` as a fixture here, instead of just calling the function?

tests/integration/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+from pathlib import Path
+import pytest
+from commonlib.io_utils import get_k8s_yaml_objects
+
+
+DEPLOY_YML = "../../deploy/cloudbeat-pytest.yml"
+
+
+@pytest.fixture(scope='module', name='start_stop')
+def fixture_start_stop(k8s, api_client, cloudbeat_agent):
```
oren-zohar, line 14: I think the name could be more indicative, something like `start_cloudbeat`.

tests/integration/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    This fixture starts cloudbeat/agent on test module setup and
```
oren-zohar, line 16: only cloudbeat, right?

tests/integration/tests/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+    @param api_client: Docker api / FileSystem api
+    @param cloudbeat_agent: Cloudbeat configuration
+    @return: Kubernetes object, Api client, Cloudbeat config
+    """
+    file_path = Path(__file__).parent / DEPLOY_YML
+    if k8s.get_agent_pod_instances(agent_name=cloudbeat_agent.name,
+                                   namespace=cloudbeat_agent.namespace):
+        k8s.delete_from_yaml(get_k8s_yaml_objects(file_path=file_path))
+    k8s.start_agent(yaml_file=file_path, namespace=cloudbeat_agent.namespace)
+    time.sleep(5)
```
oren-zohar, line 28: we should avoid sleep in general, can you find an alternative?

tests/configuration.py
```diff
@@ -33,5 +34,5 @@
 docker.use_docker = bool(strtobool(os.getenv('USE_DOCKER', 'True')))
 
 # Printing all environment keys
-for key, value in sorted(os.environ.items()):
-    print(f'{key}: {value}')
+# for key, value in sorted(os.environ.items()):
+#     print(f'{key}: {value}')
```
oren-zohar, line 38: we should avoid commenting out code blocks like that, can be removed

tests/conftest.py
```diff
@@ -46,7 +49,7 @@ def api_client():
     @return: Client (docker / FsClient).
     """
     docker_config = configuration.docker
-    print(f"Config use_docker value: {docker_config.use_docker}")
+    # print(f"Config use_docker value: {docker_config.use_docker}")
```
oren-zohar, line 52: remove

oren-zohar (changes requested):

tests/integration/tests/conftest.py
gurevichdmitry: This fixture performs also start and stop cloudbeat, therefore if we need more indicative name it may be fixture_start_stop_cloudbeat

tests/integration/tests/conftest.py
gurevichdmitry: yes

tests/configuration.py
gurevichdmitry: removed

tests/conftest.py
gurevichdmitry: removed

tests/product/tests/conftest.py
gurevichdmitry: deleted

tests/integration/tests/conftest.py
gurevichdmitry: Because start_stop fixture allows to start cloudbeat before test actions and stop it after test execution, while function call requires additional logic implementation.
Fixture_data called several times during the test run, but start_stop only once because of scope definition.
More about fixtures may be found [here](https://docs.pytest.org/en/6.2.x/fixture.html)

DaveSys911 (approved):
LGTM

tests/integration/tests/conftest.py
gurevichdmitry: sleep replaced by waiting for resource function

oren-zohar: @gurevichdmitry there's still sleep here?

tests/integration/tests/conftest.py
gurevichdmitry: docstring is updated

tests/integration/tests/conftest.py
gurevichdmitry: no, may be you check outdated file

PR #232, opened 2022-06-21
Opa logger rename fields

amirbenun:
Fixes https://github.com/elastic/security-team/issues/4168

### DIFF
--- a/beater/bundle/bundle.go
+++ b/beater/bundle/bundle.go
@@ -42,7 +42,7 @@ var (
             }
         },
         "decision_logs": {
-            "console": true
+            "console": false
         }
     }`
 )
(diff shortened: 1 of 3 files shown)


amitkanfer:
👑

beater/bundle/bundle.go
```diff
@@ -42,7 +42,7 @@ var (
             }
         },
         "decision_logs": {
-            "console": true
+            "console": false
```
amirbenun, line 45: The decision logs are not affected by the logger changes that are part of the PR.
IMO it's OK to disable the decision logs to the console since we already have the decisions printed as a debug log in `events_creator.go`.
WDYT @amitkanfer @eyalkraft ?

eyalkraft: I agree.

PR #236, opened 2022-06-22
Feature 3893/update cloudbeat dependencies

amirbenun:
Resolves https://github.com/elastic/security-team/issues/3893
Resolves https://github.com/elastic/security-team/issues/4071


config/config.go
```diff
@@ -54,7 +54,7 @@ var DefaultConfig = Config{
 	Period: 4 * time.Hour,
 }
 
-func New(cfg *common.Config) (Config, error) {
```
yashtewari, line 57: What’s the reason for changing `New` -> `NewConfig`?

Go likes it when the packagename+typename reads like this for public functions, so in usage it becomes `config.New` (vs `config.NewConfig`)

Examples:
1. https://pkg.go.dev/reflect#New
2. https://pkg.go.dev/k8s.io/client-go/kubernetes#New

oren-zohar (approved):

resources/fetchers/file_system_fetcher.go
```diff
@@ -24,11 +24,12 @@ import (
 	"strconv"
 	"syscall"
 
+	"github.com/elastic/cloudbeat/resources/utils"
```
yashtewari, line 27: I would use import grouping like 
```
import (
     standard-library-imports

     same-org-imports (github.com/elastic/...)

     third-party-imports
)
```

yashtewari (approved):
Looks good! Left some minor comments

PR #240, opened 2022-06-23
Disable OPA decision logs

amirbenun:
Related to https://github.com/elastic/security-team/issues/4168


eyalkraft (approved):

evaluator/opa.go
```diff
@@ -39,6 +39,19 @@ type OpaEvaluator struct {
 	bundleServer *http.Server
 }
 
+var opaConfig = `{
```
ofiriro3, line 42: It would be nice if we will able to enable the decision logs while debugging with a flag or something

amirbenun: We don't have the infra for that at the moment so I think it's out of scope.

PR #249, opened 2022-06-28
Add elastic common file data to cloudbeat events

uri-weisman:
Relates to:
- https://github.com/elastic/security-team/issues/3788


.github/workflows/unit-test.yml
```diff
@@ -78,6 +78,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
+      - uses: actions/setup-go@v3
```
uri-weisman, line 81: @oren-zohar  - this makes the linter issues go away.
```
  Error: could not import github.com/elastic/cloudbeat/pipeline (-: could not load export data: no export data for "github.com/elastic/cloudbeat/pipeline") (typecheck)
```

beater/cloudbeat.go
```diff
@@ -22,7 +22,7 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/elastic/cloudbeat/pipeline" //nolint: typecheck
+	"github.com/elastic/cloudbeat/pipeline"
```
uri-weisman, line 25: Due to the workflow update, we can remove this comment as well.

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 )
 
 require (
+	github.com/djherbis/times v1.5.0
```
uri-weisman, line 35: Go pkg to support file time attributes (ctime, mtime, atime) for most platforms.

transformer/events_creator.go
```diff
@@ -65,12 +65,13 @@ func (t *Transformer) CreateBeatEvents(ctx context.Context, eventData evaluator.
 			Meta:      t.eventMetadata,
 			Timestamp: timestamp,
 			Fields: mapstr.M{
-				"resource":    resource,
-				"resource_id": resMetadata.ID,   // Deprecated - kept for BC
-				"type":        resMetadata.Type, // Deprecated - kept for BC
-				"cycle_id":    eventData.CycleId,
-				"result":      finding.Result,
-				"rule":        finding.Rule,
+				resMetadata.Type: eventData.GetElasticCommonData(),
```
uri-weisman, line 68: @kfirpeled  - As this field has no mapping, ES index it dynamically.
I guess that we are fine with it as the purpose of this change is the ability to match data across multiple features.
Plz, correct me if I'm wrong.

uri-weisman:
Screen Shot 2022-06-28 at 13 56 47

resources/fetchers/file_system_fetcher.go
```diff
@@ -40,18 +41,23 @@ const (
 	GroupFile      = "/hostfs/etc/group"
 )
 
-type FileSystemResource struct {
+type EvalFSResource struct {
 	Name    string `json:"name"`
 	Mode    string `json:"mode"`
-	Gid     uint32 `json:"gid"`
-	Uid     uint32 `json:"uid"`
+	Gid     string `json:"gid"`
```
kfirpeled, line 47: I see here schema change which raises some concerns for me that I wanna make sure we've covered.

1. We should check if this breaks any existing elasticsearch mapping
2. In case this is the format of the input for the OPA rules, we need to check it is being evaluated correctly.

uri-weisman: @kfirpeled 
1. The raw resource isn't being mapped [(index_mapping)](https://github.com/elastic/integrations/blob/main/packages/cloud_security_posture/data_stream/findings/fields/findings.yml#L48-L50)
2. Currently, we don't use `Uid` & `Gid` in our OPA code.

transformer/events_creator.go
uri-weisman: > I'm not sure I'm following the code example here, What are the values for `resMetadata.Type`? do we make sure it is only `file` & `process`?
>

At the moment, only the file & process fetchers implement the `GetElasticCommonData` and their types are `file` and `process` accordingly.
Once other fetchers will support it we can modify this logic and add a switch case or something similar.

Regarding the ECS mapping, it seems that it's not being applied by default [(discussion)](https://discuss.elastic.co/t/ecs-fields-not-mapping-correctly-automatically/224190/3), and the mapping needed to be imported explicitly [(fields.ecs.yml)](https://github.com/elastic/ecs/blob/8.3/generated/beats/fields.ecs.yml).

resources/fetchers/file_system_fetcher.go
```diff
@@ -165,3 +178,47 @@ func getFSSubType(fileInfo os.FileInfo) string {
...
+	if t.HasChangeTime() {
+		cd["Ctime"] = t.ChangeTime()
+	}
+
+	return nil
+}
+
+func enrichFromFileResource(cd mapstr.M, data EvalFSResource) {
+	cd["UID"] = data.Uid
+	cd["GID"] = data.Uid
```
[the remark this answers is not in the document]
uri-weisman: Good catch

PR #250, opened 2022-06-29
Rerun beater interface on reconfiguration

amirbenun:
Resolves https://github.com/elastic/security-team/issues/3876


rerun/launcher.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = s.waitForUpdates()
+	s.log.Error("Beater starter is stopping: %w", err)
+	return err
+}
+
+func (s *launcher) Stop() {
```
yashtewari, line 119: Nitpicking 🙈 `s`->`l`

rerun/launcher.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+}
+
+// configUpdate applies incoming reconfiguration from the Fleet server to the beater config,
+// and recreate the beater with the new values.
+func (s *launcher) configUpdate(update *config.C) error {
+	s.log.Info("Got config update")
+
+	err := s.latest.MergeWithOpts(update, ucfg.ReplaceArrValues)
```
yashtewari, line 175: Since we're doing a merge and not a replace, given that the incoming configuration is only partial, maybe the deleted tests from the earlier "merge configuration" logic are still useful as a double check?

Unrelated, did you explore having the integration send the whole configuration?

yashtewari (approved):
Looks awesome

rerun/launcher.go
yashtewari: > Regarding the first comment, not sure what you meant.

I was talking about some deleted tests, but I see now that they are replaced by analogous tests. Nevermind!

PR #254, opened 2022-07-06
Add ECS fields to process & FS resources

uri-weisman:
The work has been done as part of the ECS mapping epic:

- https://github.com/elastic/security-team/issues/3788
- https://github.com/elastic/security-team/issues/3787


resources/fetchers/process_fetcher.go
```diff
@@ -40,15 +43,65 @@ const (
 	CMDArgumentMatcher  = "\\b%s[\\s=]\\/?(\\S+)"
 	ProcessResourceType = "process"
 	ProcessSubType      = "process"
+	userHz              = 100
```
jeniawhite, line 46: The value of HZ varies across kernel versions and hardware platforms.
There is an assumption here for certain conditions that might not be respected between different deployments and versions.
I think that we can use https://man7.org/linux/man-pages/man3/sysconf.3.html and pull the value of `_SC_CLK_TCK` in order to reduce this assumption.

uri-weisman: Thank you for raising this up @jeniawhite .
It is a borrowed implementation from [go-sysinfo](https://github.com/elastic/go-sysinfo), we decided to deliver it with this known limitation while opening 2 new tickets to the go-sysinfo repo.
- https://github.com/elastic/go-sysinfo/issues/122
- https://github.com/elastic/go-sysinfo/issues/12

Once those tasks will be delivered, possibly by us, we'll change our implementation to use the go-sysinfo pkg.

jeniawhite (approved):

DaveSys911 (approved):
LGTM

.github/workflows/unit-test.yml
```diff
@@ -78,6 +78,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
```
DaveSys911, line 83: Could you please set go version from using the env var like the unit tests job?

DaveSys911 (changes requested):

resources/fetchers/kube_provider.go
```diff
@@ -80,6 +80,8 @@ func (r K8sResource) GetMetadata() fetching.ResourceMetadata {
 	}
 }
 
+func (r K8sResource) GetElasticCommonData() any { return nil }
```
oren-zohar, line 83: shouldn't this sit in the kube_fetcher?

uri-weisman: Seems fine to me to reside next to the rest of the `Resource interface`
```
type Resource interface {
	GetMetadata() ResourceMetadata
	GetData() any
	GetElasticCommonData() any
}
```

If you feel this is not the place we can move them all.

.github/workflows/unit-test.yml
uri-weisman: Fixed

DaveSys911 (approved):

Issue #256, opened 2022-07-07
Write a script that replaces agent's Cloudbeat 


ofiriro3:
**Motivation**
As a developer, I would like to quickly replace the running cloudbeat within an agent.

**In-Depth**
Many times during the development cycle we would like to develop/test our Cloudbeat code while it runs within an agent.
Currently, we can use `make PackageAgent` (https://github.com/elastic/cloudbeat/pull/222) but most of the time we don't want to build a new agent, we just want to replace our cloudbeat with the current running cloudbeat.

This task was created to ease the developing cycle of the EKS integration but it is relevant for every developing cycle that includes the agent.

**Definition of done**
- Running script that replaces cloudbeat for all running agent
- The script supported both arm and amd processors.

PR #257, opened 2022-07-07
Adding script for replacing running Cloudbeat within an agent

ofiriro3:
- Closes: https://github.com/elastic/cloudbeat/issues/256


scripts/remote_replace_cloudbeat.sh
```diff
@@ -0,0 +1,35 @@
+# This script will stop any Cloudbeat process within an agent and will replace the Cloudbeat artifacts with your local built artifacts
+
+# Prerequisites:
+# a. Create the Cloudbeat artifacts using the mage command (for example - "DEV=true PLATFORMS=linux/amd64 SNAPSHOT=true mage -v package")
+# b. The running elastic-agents were built with "DEV=TRUE" flag (https://github.com/elastic/security-team/blob/main/docs/cloud-security-posture-team/Onboarding/cloudbeat-ec2.md#:~:text=Build%20the%20agent%20(We%20are%20building%20with%20DEV%3Dtrue%2C%20so%20we%20can%20load%20cloudbeat.tar.gz%20without%20signature%20verification))
+
+#!/bin/bash
+ARCH=`uname -a | rev | cut -d " " -f 1 | rev` # This should return arm64 on m1 and x86_64 on regular mac
+PREFIX="/usr/share/elastic-agent/data/"
+SUFFIX="/install/cloudbeat-8.4.0-SNAPSHOT-linux-${ARCH}"
```
oren-zohar, line 10: maybe we can use `defaultBeatVersion` instead here?

scripts/remote_replace_cloudbeat.sh
```diff
@@ -0,0 +1,35 @@
...
+DISTRIBUTION_LOCAL_FOLDER="../build/distributions"
+CLOUDBEAT_IMAGE="cloudbeat-8.4.0-SNAPSHOT-linux-$ARCH"
```
oren-zohar, line 12: smae here

scripts/remote_replace_cloudbeat.sh
```diff
@@ -0,0 +1,35 @@
...
+
+PODS=`kubectl -n kube-system get pod -l app=elastic-agent -o name`
+for P in $PODS
+do
+    POD=`echo $P | cut -d '/' -f 2`
+    FOLDER=`kubectl -n kube-system exec $POD -- ls $PREFIX`
+    # Kill Cloudbeat process
+    CLOUDBEAT_PID=`kubectl -n kube-system exec "$POD" -- pidof cloudbeat`
+    kubectl -n kube-system exec "$POD" -- kill -9 "${CLOUDBEAT_PID}"
```
oren-zohar, line 21: wouldn't that will trigger a restart of cloudbeat?

oren-zohar: so maybe we should copy it before killing it?

scripts/remote_replace_cloudbeat.sh
```diff
@@ -0,0 +1,35 @@
...
+    kubectl -n kube-system exec "$POD" -- rm -r "${PREFIX}${FOLDER}${SUFFIX}"
+done
+
+# After the script finishes its work, you need to make Cloudbeat run again.
+# This can be achieved by:
+# a. Open Kibana and press the menu button on the top left of the screen.
+# b. Select "Fleet".
+# c. Press the "agent policies" tab and select the relevant policy.
+# d. Select the "..." next to the CIS Kubernetes benchmark integration and select `edit integration`.
+# e. Change the `integration name` and press on `save integration`.
```
oren-zohar, line 35: I think this can be automated by copying the command that was used to run cloudbeat in the first place from running `ps -aux | grep cloudbeat`, right?

scripts/remote_replace_cloudbeat.sh
```diff
@@ -0,0 +1,35 @@
+# This script will stop any Cloudbeat process within an agent and will replace the Cloudbeat artifacts with your local built artifacts
+
+# Prerequisites:
+# a. Create the Cloudbeat artifacts using the mage command (for example - "DEV=true PLATFORMS=linux/amd64 SNAPSHOT=true mage -v package")
+# b. The running elastic-agents were built with "DEV=TRUE" flag (https://github.com/elastic/security-team/blob/main/docs/cloud-security-posture-team/Onboarding/cloudbeat-ec2.md#:~:text=Build%20the%20agent%20(We%20are%20building%20with%20DEV%3Dtrue%2C%20so%20we%20can%20load%20cloudbeat.tar.gz%20without%20signature%20verification))
+
```
oren-zohar, line 6: is there a way to overcome this Prerequisite? Does this basically mean we will only be able to use this script if we compiled the agent beforehand?

oren-zohar (changes requested):

scripts/remote_replace_cloudbeat.sh
ofiriro3: If we will use an existing agent that we weren't compiled with DEV=TRUE, so the agent will throw an error that it can't validate the cloudbeat certificate.

scripts/remote_replace_cloudbeat.sh
ofiriro3: I tried to do it as Amir once suggested on his script, but since we are deleting the artifacts this command cannot be run.

`/usr/share/elastic-agent/data/elastic-agent-bbd93f/install/cloudbeat-8.4.0-SNAPSHOT-linux-x86_64/cloudbeat -E setup.ilm.enabled=false -E setup.template.enabled=false -E management.enabled`

This installation path is the old artifact, and if we would not delete it we will run the agent again and again.
And obviously, if the path is deleted as part of the script we can't run it. 

Upsetting :/

scripts/remote_replace_cloudbeat.sh
ofiriro3: Same as the other comment suggested.

oren-zohar: why can't we run the new cloudbeat artifact with the same parmas from here?

scripts/remote_replace_cloudbeat.sh
oren-zohar: yeah I know I'm just trying to think of a way we can overcome this. If we could find a way to that this script would be amazing, let's talk about it next week

scripts/remote_replace_cloudbeat.sh
oren-zohar: > I tried to do it as Amir once suggested on his script

and what was the result? it didn't work?

> This installation path is the old artifact, and if we would not delete it we will run the same old cloudbeat again and again.
And obviously, if the path is deleted as part of the script we can't run it.

but here we are killing it before deleting it so it could trigger a re-run of the old beat. 
I am just suggesting we replace the old beat before killing the current one if that's possible

oren-zohar (changes requested):

scripts/remote_replace_cloudbeat.sh
ofiriro3: Because the artifacts folder is being created from the tar.gz you copy to the folder

scripts/remote_replace_cloudbeat.sh
ofiriro3: - yeah it didn't work, and didn't work for him as well since there is no artifact folder.
- The thing is, we want the cloudbeat process to be killed, so it will extract the tar.gz and init the process. Currently we don't know how to do the extraction ourselves.
Wondering if it is a good practice to replace running directory while the process still running - I think it can lead to a lot of problems:/

DaveSys911:
Thank you for opening this issue

scripts/remote_replace_cloudbeat.sh
ofiriro3: Was changed according to our call.

The kill was moved to the end of the script.

oren-zohar (approved):
Approved, but I would add somewhere in the script that after the process is killed it restarts automatically.

PR #259, opened 2022-07-07
support bundle loading per FW

oren-zohar:
Adding bundle configuration support per framework (Vanilla and EKS)

### DIFF
--- a/evaluator/bundle.go
+++ b/evaluator/bundle.go
@@ -74,7 +74,7 @@ func StartServer(ctx context.Context, cfg config.Config) (*http.Server, error) {
 
 // loadBundle loads the relevant bundle based on the config.
 func loadBundle(cfg config.Config, log *logp.Logger) csppolicies.Bundle {
-	bundle := csppolicies.Bundle{}
+	var bundle csppolicies.Bundle
 
 	if cfg.Type == config.InputTypeEKS {
 		bundle = csppolicies.CISEksBundle()
--- a/evaluator/bundle.go
+++ b/evaluator/bundle.go
@@ -35,15 +35,19 @@ var (
 )
 
 func StartServer(ctx context.Context, cfg config.Config) (*http.Server, error) {
-	bundle := csppolicies.CISKubernetesBundle()
+	log := logp.NewLogger("cloudbeat_bundle_server")
 
+	bundle := loadBundle(cfg, log)
+
+	// if no config is provided, all rules will be activated
 	if len(cfg.Streams) > 0 {
-		dataYml, err := cfg.DataYaml()
+		activatedRules, err := cfg.GetActivatedRules()
 		if err != nil {
 			return nil, fmt.Errorf("could not marshal to YAML: %w", err)
 		}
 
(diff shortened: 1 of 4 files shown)


evaluator/bundle.go
```diff
@@ -35,15 +35,19 @@ var (
 )
 
 func StartServer(ctx context.Context, cfg config.Config) (*http.Server, error) {
-	bundle := csppolicies.CISKubernetesBundle()
+	log := logp.NewLogger("cloudbeat_bundle_server")
+
```
ofiriro3, line 39: nit - no space is needed here

evaluator/bundle.go
```diff
@@ -69,3 +71,18 @@ func StartServer(ctx context.Context, cfg config.Config) (*http.Server, error) {
 
 	return srv, nil
 }
+
+// loadBundle loads the relevant bundle based on the config.
+func loadBundle(cfg config.Config, log *logp.Logger) csppolicies.Bundle {
```
ofiriro3, line 76: Wondering about this method.

Do we wish to throw an error if we don't receive any of the two?

WDT?

evaluator/bundle_test.go
```diff
@@ -94,17 +94,10 @@ func (s *BundleTestSuite) TestCreateServerWithDataYaml() {
...
     streams:
-      - data_yaml:
-          activated_rules:
-            cis_k8s:
-              - a
-              - b
-              - c
-              - d
-              - e
+      - data_yaml
```
ofiriro3, line 97: Was the test changed on purpose?

evaluator/bundle_test.go
```diff
@@ -170,3 +163,77 @@ func (s *BundleTestSuite) TestCreateServerWithDataYaml() {
...
+
+	time.Sleep(time.Second * 2)
+	for _, test := range tests {
+		s.Run(test.name, func() {
+			server, err := StartServer(context.Background(), test.cfg)
+			s.NoError(err)
+
+			target := ServerAddress + test.path
+			client := &http.Client{}
+			res, err := client.Get(target)
```
ofiriro3, line 230: Not a clastic unit test, as far as I understand it uses a real HTTP client. 
I would have mocked it, but you relied on others' work.

Definitely not a must.

ofiriro3 (changes requested):
Good job. 🦑 

Few small comments + I would love to hear your thoughts on one of the questions I left

evaluator/bundle_test.go
oren-zohar: yeah the test was wrong

evaluator/bundle.go
oren-zohar: good point. 
I had the impression that we would want to have vanilla as default in any case the config has some misconfiguration as we had before.
Additionally, keeping it like that will be backward compatible with the previous config format, which is still what we have today, and will not break the integration test. We can update this function to be more strict once we have working integration between the config and cloudbeat.

ofiriro3 (approved):
🚀

PR #261, opened 2022-07-10
fetchers support by cloud provider (Vanilla/EKS)

oren-zohar:
Main changes in this PR:
- [ ] ~Adding `suppertedFetchers` that is set per Cloud Provider~
- [x] One deployment yml with both fetchers configured, loading them by the type of the stream `(cloudbeat/vanilla`/`cloudbeat/eks`)
- [x] Testing refactoring across our beat to support this change (updating ymls)
- [x] `Manager`-> `FetchersManager` (name collided with beat's `Manager` and was too generic)

** Many of the changes here will become obsolete once we get the fetcher config from Kibana
____
- Closes: https://github.com/elastic/security-team/issues/3041


resources/fetchersManager/data_test.go
```diff
@@ -109,7 +110,7 @@ func TestDataTestSuite(t *testing.T) {
 func (s *DataTestSuite) SetupTest() {
 	s.ctx = context.Background()
 	s.opts = goleak.IgnoreCurrent()
-	s.registry = NewFetcherRegistry(s.log)
+	s.registry = NewFetcherRegistry(s.log, config.DefaultConfig)
```
ofiriro3, line 113: I think we can remove the config, you didn't use it at the end

resources/fetchersManager/data_test.go
```diff
@@ -15,11 +15,12 @@
 // specific language governing permissions and limitations
 // under the License.
 
-package manager
+package fetchersManager
```
ofiriro3, line 18: I prefer to revert the file.

I know that we might need to do it again in the future.

But we don't know when/how we will implement the validation - until then, I think we should not assume anything about the implementation.

resources/fetchersManager/factory.go
```diff
@@ -40,7 +40,7 @@ func newFactories() factories {
 	return factories{m: make(map[string]fetching.Factory)}
 }
 
-func (fa *factories) ListFetcherFactory(name string, f fetching.Factory) {
+func (fa *factories) SetFetcherFactory(name string, f fetching.Factory) {
```
ofiriro3, line 43: Can we change to `RegisterFetcher`/`RegisterFetcherInFactody`?

I think that if we use a known pattern we should stick with the conventions 🥇

resources/fetchersManager/registry.go
```diff
@@ -15,21 +15,23 @@
...
 	"context"
 	"fmt"
+	"github.com/elastic/cloudbeat/config"
 	"github.com/elastic/cloudbeat/resources/fetching"
 	"github.com/elastic/elastic-agent-libs/logp"
 )
 
 type FetchersRegistry interface {
-	Register(key string, f fetching.Fetcher, c ...fetching.Condition) error
+	Register(fetcherName string, f fetching.Fetcher, c ...fetching.Condition) error
```
ofiriro3, line 29: 🚀

resources/fetchersManager/registry.go
```diff
@@ -42,21 +44,39 @@ type registeredFetcher struct {
 	c []fetching.Condition
 }
 
-func NewFetcherRegistry(log *logp.Logger) FetchersRegistry {
-	return &fetchersRegistry{
+func NewFetcherRegistry(log *logp.Logger, cfg config.Config) FetchersRegistry {
```
ofiriro3, line 47: Can you remove the config paramater?

resources/fetchersManager/registry.go
```diff
@@ -42,21 +44,39 @@ type registeredFetcher struct {
...
+
+	return nil
 }
 
 // Register registers a Fetcher implementation.
-func (r *fetchersRegistry) Register(key string, f fetching.Fetcher, c ...fetching.Condition) error {
-	r.log.Infof("Registering new fetcher: %s", key)
-	if _, ok := r.reg[key]; ok {
-		return fmt.Errorf("fetcher key collision: %q is already registered", key)
+func (r *fetchersRegistry) Register(fetcherName string, f fetching.Fetcher, c ...fetching.Condition) error {
```
ofiriro3, line 73: Do `Register` now need to stay public method?
I guess it can go private, isn't it?

resources/fetchersManager/registry_test.go
```diff
@@ -15,17 +15,16 @@
 // specific language governing permissions and limitations
 // under the License.
```
ofiriro3, line 17: Same as the other test files

ofiriro3 (changes requested):
Very good job.

Really like that you cleaned/improved that code that you touched!

Few comments to address

beater/cloudbeat.go
```diff
@@ -63,7 +63,7 @@ func New(b *beat.Beat, cfg *agentconfig.C) (beat.Beater, error) {
 	reloader := launcher.NewListener(ctx, log)
 	validator := &validator{}
 
-	s, err := launcher.New(ctx, log, reloader, validator, b, NewCloudbeat, cfg)
+	s, err := launcher.New(ctx, log, reloader, validator, NewCloudbeat, cfg)
```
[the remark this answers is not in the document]
oren-zohar: we didn't use it

beater/cloudbeat.go
```diff
@@ -72,8 +72,8 @@ func New(b *beat.Beat, cfg *agentconfig.C) (beat.Beater, error) {
 	return s, nil
 }
 
-// New creates an instance of cloudbeat.
-func NewCloudbeat(b *beat.Beat, cfg *agentconfig.C) (beat.Beater, error) {
+// NewCloudbeat creates an instance of cloudbeat.
+func NewCloudbeat(_ *beat.Beat, cfg *agentconfig.C) (beat.Beater, error) {
```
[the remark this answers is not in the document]
oren-zohar: what wildcard?

config/config.go
```diff
@@ -37,8 +37,17 @@ const (
 	InputTypeEKS = "cloudbeat/eks"
 )
 
+type Fetcher struct {
+	Name string `config:"name"` // Name of the fetcher
+}
+
+type Fetchers struct {
```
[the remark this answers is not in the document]
oren-zohar: sure, what's confusing about it?

resources/fetchersManager/factory.go
oren-zohar: register implies registry and could be very confusing, we already have `RegisterFetcher` function in the registry so I think you can see why it can be confusing with our current registry. What convention?

beater/cloudbeat.go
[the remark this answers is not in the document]
ofiriro3: _

resources/fetchersManager/factory.go
```diff
@@ -58,28 +58,6 @@ func (fa *factories) CreateFetcher(log *logp.Logger, name string, c *agentconfig
 	return factory.Create(log, c, ch)
 }
 
-func (fa *factories) RegisterFetchers(log *logp.Logger, registry FetchersRegistry, cfg config.Config, ch chan fetching.ResourceInfo) error {
```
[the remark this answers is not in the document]
oren-zohar: wasn't removed, but moved to a place that makes more sense to me. We have the factory that should handle the factories and we have the fetcher registry that handles registration fetchers, other implementation details shouldn't change these two facts. And regarding newcomers that will find this change complex, from my own experience here, the old design really made it hard for me.

> I think that the idea was that the factory will have one function that will register all of the fetchers.

exactly, why should the factory care about the registry job?

> The author of the original code tries to encapsulate all the registration in a single line of code, so it will be easier to use the factory.

what does that mean? a single line of code? easier to use the factory? why should we use the factory to register the fetchers in the first place?

resources/fetchersManager/registry_test.go
oren-zohar: not sure I can revert everything here but I'll try

config/config.go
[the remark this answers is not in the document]
ofiriro3: I would expect Fetcher to be an interface for a fetcher instance.

So `fetchers` seems to me like something related. 
I guess something like -> FetchersConfiguration/ FetchersSetup something that will impliy that we will use it to init the fetchers and not the fetchers themselves.

resources/fetchersManager/factory.go
ofiriro3: This is the closest example that I cloud find:

https://medium.com/@geoffreykoh/implementing-the-factory-pattern-via-dynamic-registry-and-python-decorators-479fc1537bbe

I don't mind finding the middle ground - 
`AddFetcher`, `EnlistFetcher`.

Set gives the impression of replacing changing a value from `a` to `b`

ofiriro3 (approved):
Please change the `SetFetcherFactory` to `RegisterFactory`

PR #263, opened 2022-07-11
Tests-ci-v2

DaveSys911:


.github/workflows/Periodic-CI.yml
```diff
@@ -0,0 +1,237 @@
...
+    - id: Nightly
+      if: github.event_name == 'schedule' && github.event.schedule == '00 02 * * *'
+      env:
+        TEST_TARGETS: '"[\"pre_merge\", \"file_system_rules\", \"k8s_object_rules\", \"process_api_server_rules\", \"process_controller_manager_rules\", \"process_etcd_rules\", \"process_kubelet_rules\", \"process_scheduler_rules\"]"'
+      run: echo "TEST_MATRIX=${TEST_TARGETS}" >> $GITHUB_ENV
+    - id: Weekly
+      if: github.event_name == 'schedule' && github.event.schedule != '00 02 * * 6'
+      env:
+        TEST_TARGETS: '[\"pre_merge\", \"file_system_rules\", \"k8s_object_rules\", \"process_api_server_rules\", \"process_controller_manager_rules\", \"process_etcd_rules\", \"process_kubelet_rules\", \"process_scheduler_rules\"]'
+      run: echo "TEST_MATRIX=${TEST_TARGETS}" >> $GITHUB_ENV
```
gurevichdmitry, line 43: According to the **if** statement weekly is defined by event 'schedule' and **not** '00 02 * * 6'.
What did you meant?

gurevichdmitry (commented):
Please check in id: Weekly "if" statement definition.

.github/workflows/cloudbeat-ci.yml
```diff
@@ -82,63 +85,66 @@ jobs:
...
-    name: Integration Tests
+  Test_Matrix:
+    name: ${{ matrix.test-target }}-tests
+    needs: Build
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    strategy:
+      matrix:
+        test-target: [pre_merge, file_system_rules, k8s_object_rules, process_api_server_rules, process_controller_manager_rules, process_etcd_rules, process_kubelet_rules, process_scheduler_rules]
+      fail-fast: false
```
gurevichdmitry, line 96: Pre merge integration tests should execute only **pre_merge** marker tests as required.
Functional tests may be executed but not as required.

.github/workflows/Periodic-CI.yml
```diff
@@ -0,0 +1,237 @@
+name: Periodic-CI
```
gurevichdmitry, line 1: all file names are lowercase. Please rename Periodic-CI.yml to be on the same convention.

yashtewari (commented):
Finally at the finish line! I just want to mention a couple of things:
- I remember getting some unexpected exit codes from the `helm test` commands even when all the tests run and pass, so you might want to double check that when depending on the exit codes to decide pass/fail.
- The command
```
just collect-logs <test-target>
```
will keep dumping logs from the test pod into the file `test-logs/<test-target>.log` up until the point that tests are done running. This would be an improvement over the pipe to `tee test.log` that you currently have which will only dump logs at the _end_ of the test run. It's a known issue that in case a test breaks the cluster (which isn't that unlikely), then you lose access to the logs which are in the test pod, behind the broken Kube API, so in that case with `tee` you can end up with no logs whatsoever.

Not critical for this PR though! Looks great overall

DaveSys911:
> Finally at the finish line! I just want to mention a couple of things:
> 
> * I remember getting some unexpected exit codes from the `helm test` commands even when all the tests run and pass, so you might want to double check that when depending on the exit codes to decide pass/fail.
> * The command
> 
> ```
> just collect-logs <test-target>
> ```
> 
> will keep dumping logs from the test pod into the file `test-logs/<test-target>.log` up until the point that tests are done running. This would be an improvement over the pipe to `tee test.log` that you currently have which will only dump logs at the _end_ of the test run. It's a known issue that in case a test breaks the cluster (which isn't that unlikely), then you lose access to the logs which are in the test pod, behind the broken Kube API, so in that case with `tee` you can end up with no logs whatsoever.
> 
> Not critical for this PR though! Looks great overall

@yashtewari @Dima Good catch. After examining it a bit more carefully, I noticed a certain gap in our shell invocation logic which might lead to inconsistent results. So the main points are:
1. Explicit & consistent shell type
We don't set the shell type consistently & explicitly. It's a mix of bash, sh & whatever the system default is(as for most of us, bash for GH).

2. Just's interference with exit codes:
We didn't take into consideration how just might affect the exit codes- Sometimes just can get in the way and return something different than expected.

3. Intentional exit code control:
When daisy-chaining various commands, some more elaborate error handling needs to be in place. for example:
```
  #!/usr/bin/env bash
  helm test {{TESTS_RELEASE}} --namespace kube-system --kube-context kind-kind-mono --timeout {{TESTS_TIMEOUT}} --logs 2>&1 | tee test.log
  result_code=${PIPESTATUS[0]}
  SUMMARY=$(cat test.log | sed -n '/summary/,/===/p')
  echo "summary<<EOF" >> "$GITHUB_ENV"
  echo "$SUMMARY" >> "$GITHUB_ENV"
  echo "EOF" >> "$GITHUB_ENV"
  exit $result_code
```

4. Keep it consistent between our local envs & ci- 
No need to elaborate here.I guess we might get 3 different outcomes on some runs.

Will piggy bank the existing todo & ticket To document these https://github.com/elastic/security-team/issues/4291, will prob generate some follow up task.

.github/workflows/publish-test-results.yml
```diff
@@ -0,0 +1,73 @@
...
+          updatePr: actions
+          collapseSummary: false
+          summary: suites
+          summaryTableType: markdown
+          copyLatest: true
+          bucket: csp-allure-reports
+          prefix: allure_reports/cloudbeat/${{ github.event.workflow_run.head_branch }}
+          ignoreMissingResults: true
+
+#Todo comment allure results if pr gh ticket https://github.com/elastic/security-team/issues/3936
```
oren-zohar, line 58: this todo is not very clear. Is it referring to a ticket that summarizes what needs to be done? if so we can remove the comment and link this line from the ticket. If not, let's remove it since it is not clear and will probably stay here for a while.

.github/workflows/cloudbeat-ci.yml
```diff
@@ -82,63 +82,67 @@ jobs:
...
-  Integration_tests:
-    name: Integration Tests
+  Test_Matrix:
+    name: ${{ matrix.test-target }}-tests
+    needs: Build
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    strategy:
+      matrix:
+        # test-target: [pre_merge, file_system_rules, k8s_object_rules, process_api_server_rules, process_controller_manager_rules, process_etcd_rules, process_kubelet_rules, process_scheduler_rules]
```
oren-zohar: I guess we can remove this

.github/workflows/Periodic-CI.yml
```diff
@@ -0,0 +1,229 @@
...
+        required: true
+        description: 'Specify test markers to run'
+        default: '[\"pre_merge\", \"file_system_rules\", \"k8s_object_rules\", \"process_api_server_rules\", \"process_controller_manager_rules\", \"process_etcd_rules\", \"process_kubelet_rules\", \"process_scheduler_rules\"]'
+
+  schedule:
+    # Nightly job. Running every day at 2 am
+    - cron: '00 02 * * *'
+
+    # Weekly job running every saturday at 2 am
+    - cron: '00 02 * * 6'
```
oren-zohar, line 17: will it mean that we're going to trigger the same two jobs on Saturday? also, this is 2 AM UTC which is 5 AM for us, right?

.github/workflows/Periodic-CI.yml
DaveSys911: These are two fields coming from the payload json.
github.event_name holds the event type. Push, pr etc. while github.event.schedule holds the cron data.

It allows us to identify a schedule trigger. Differentiate  between two different schedules to determine which tests should run on weekly vs nightly for example.
the `!=` is fixed

oren-zohar (changes requested):

JUSTFILE
```diff
@@ -113,6 +113,9 @@ gen-report:
 run-tests:
   helm test {{TESTS_RELEASE}} --namespace kube-system
 
+run-tests-ci:
+  helm test {{TESTS_RELEASE}} --namespace kube-system --kube-context kind-kind-mono --logs 2>&1 | tee test.log
+
```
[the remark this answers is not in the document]
DaveSys911: Increased helm timeout to 60m & workflow for periodic-ci to 90m.
Also added some logic to control the access code, short term should be ok.

Long term, note https://github.com/elastic/cloudbeat/pull/263#issuecomment-1180357960 - some work needs to be done there.

.github/workflows/cloudbeat-ci.yml
DaveSys911: Added pre_merge & file_system rules. Only pre_merge will be set as mandatory check in gh branch rules.

.github/workflows/publish-test-results.yml
DaveSys911: Summary with link back is a good idea since this one is critical & complex.

Since tickets get re-shuffled often, I'd like to have that Todo remain there until it's 100% confirmed that  pr input is handled both secure & successfully.

I'll add a more descriptive one-liner & a link back to a detailed summary on the ticket.

.github/workflows/Periodic-CI.yml
DaveSys911: So the purpose here is to have something quickly pluggable even if dirty to serve as a reference, to make it easy to add tests to CI as quickly as possible. So it's implemented with the intent that they will be different in the upcoming weeks.

I guess weekly can contain daily checks as a subset, So we can avoid the redundant overlap.

gurevichdmitry (approved):
LGTM, impressive work!

oren-zohar (approved):

PR #266, opened 2022-07-11
fix publish reports syntax issues

DaveSys911:
Small fix for publish reports
Updated code owners. @gurevichdmitry is in both ops & automation, so it's a bit of a deadlock.


.github/CODEOWNERS
```diff
@@ -3,11 +3,11 @@
...
 # Infra & ops related
 /deploy/ @elastic/csp-ops 
-/.github/workflows  @elastic/csp-ops @elastic/csp-automation
+/.github/workflows  @elastic/csp-ops
 .ci/  @elastic/csp-ops
 /scripts/ @elastic/csp-ops
-Makefile  @elastic/csp-ops @elastic/csp-automation
-JUSTFILE  @elastic/csp-ops @elastic/csp-automation
+Makefile  @elastic/csp-ops
+JUSTFILE  @elastic/csp-ops
```
oren-zohar, line 10:
```suggestion
Makefile  @elastic/cloudbeat
JUSTFILE  @elastic/cloudbeat
```

oren-zohar (changes requested):
There is a problem with you and Dima owning too many files, which results in every PR depending on your review, even though it is a trivial change. It is not necessary for you to review every change to a makefile or justfile. We are all capable of reviewing them and there are integration tests in place to prevent any changes from breaking.

.github/CODEOWNERS
DaveSys911: Actually this change expands them to Me, @gurevichdmitry, and you.
I understand the inconvenience, but the files host logic is not currently validated as part of the pre_merge(our release pipeline for example).

Makefile for example is used for the release almost exclusively.
Both btw are barely changed by devs, so I don't see it as that much of a bottleneck.

https://github.com/elastic/cloudbeat/blob/main/Makefile
https://github.com/elastic/cloudbeat/commits/main/JUSTFILE

i prefer not to make this change on a "drive by". I prefer to add the missing pre-merge checks and separate the critical bits that should be reviewed by csp-ops, away from these file.

oren-zohar (approved):

Issue #234, opened 2022-06-22
Cloudbeat - Support EKS data yaml schema


ofiriro3:
**Motivation**
As a developer, I would the cloudbeat to be able to read the new data.yaml provided by the EKS integration.

**In-Depth**
As part of developing the EKS integration, we will need to support the new EKS data.yaml schema.
Currently, the data.yaml will look like the following:
```yaml
activated_rules:
  cis_k8s:
    - cis_1_1_1
    - cis_1_1_2
  cis_eks:
    - cis_3_1_1
    - cis_3_1_2
```

Cloudbeat currently performing a scheme validation that doesn't support the `cis_eks` section.

**Related tasks**
https://github.com/elastic/security-team/issues/4160

**Definition of done**
- Cloudbeat is able to load EKS-related data.yaml

ofiriro3:
This file is probably related to the fix

https://github.com/elastic/cloudbeat/blob/main/config/config.go

PR #267, opened 2022-07-12
EKS activated rules support

oren-zohar:
Adds support to the EKS activated rules field in the data.yaml schema 
____
- closes: https://github.com/elastic/cloudbeat/issues/234

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,7 +34,8 @@ const DefaultNamespace = "default"
 const ResultsDatastreamIndexPrefix = "logs-cloud_security_posture.findings"
 
 const (
-	InputTypeEKS = "cloudbeat/eks"
+	InputTypeVanillaK8s = "cloudbeat/vanilla"
+	InputTypeEKS        = "cloudbeat/eks"
 )
 
 type Config struct {
--- a/config/config.go
+++ b/config/config.go
@@ -56,7 +56,8 @@ type DataYaml struct {
 }
 
 type Benchmarks struct {
-	CISK8S []string `config:"cis_k8s" yaml:"cis_k8s" json:"cis_k8s"`
+	CisK8s []string `config:"cis_k8s,omitempty" yaml:"cis_k8s,omitempty" json:"cis_k8s,omitempty"`
+	CisEKS []string `config:"cis_eks,omitempty" yaml:"cis_eks,omitempty" json:"cis_eks,omitempty"`
 }
 
 var DefaultConfig = Config{
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -430,10 +430,9 @@ func (s *ConfigTestSuite) TestConfigPeriod() {
 
 func (s *ConfigTestSuite) TestActivatedRulesFrameWork() {
 	var tests = []struct {
-		config                    string
-		expectedActivatedRules    []string
-		expectedEksActivatedRules []string
-		expectedType              string
+		config                 string
+		expectedActivatedRules []string
+		expectedType           string
 	}{
 		{
 			`
@@ -446,7 +445,6 @@ streams:
           - b
 `,
 			[]string{"a", "b"},
-			nil,
 			"cloudbeat/vanilla",
 		},
 		{
@@ -459,7 +457,6 @@ streams:
           - a
           - b
 `,
-			nil,
 			[]string{"a", "b"},
 			"cloudbeat/eks",
 		},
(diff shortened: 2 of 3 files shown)


config/config_test.go
```diff
@@ -427,3 +427,53 @@ func (s *ConfigTestSuite) TestConfigPeriod() {
...
+
+	for _, test := range tests {
+		cfg, err := config.NewConfigFrom(test.config)
+		s.NoError(err)
+
+		c, err := New(cfg)
+		s.NoError(err)
+
+		s.Equal(test.expectedType, c.Type)
+		if test.expectedType == "cloudbeat/eks" {
```
ofiriro3, line 473: Can the test checks both benchmarks at the same time?

For example:
Instead of having `expectedActivatedRules []string`, we can have `expectedEKSActivatedRules []string` and `expectedVanillaActivatedRules []string`

And then in the assert, we will assert them both.

That way I think that test will "lock" the behavior even better.

WDT?

config/config.go
```diff
@@ -56,7 +56,8 @@ type DataYaml struct {
 }
 
 type Benchmarks struct {
-	CISK8S []string `config:"cis_k8s" yaml:"cis_k8s" json:"cis_k8s"`
+	CisK8s []string `config:"cis_k8s,omitempty" yaml:"cis_k8s,omitempty" json:"cis_k8s,omitempty"`
```
ofiriro3, line 59: Amazing!

evaluator/bundle_test.go
```diff
@@ -208,7 +208,7 @@ func (s *BundleTestSuite) TestCreateServerWithFetchersConfig() {
 					{
 						DataYaml: &config.DataYaml{
```
ofiriro3, line 209: Can you add test cases/tests here as well? (for eks yaml)

ofiriro3 (changes requested):
Nice work brosky :) 

Please check it with OPA, and address the comments(mostly on tests) 🥇

config/config_test.go
oren-zohar: asserting them both can result in false positive in the case where both of the configuration resulted only EKS or Vanilla

ofiriro3:
```suggestion
		func (s *ConfigTestSuite) TestActivatedRulesFrameWork() {
	var tests = []struct {
		config                    string
		expectedActivatedRules    []string
		expectedEksActivatedRules []string
		expectedType              string
	}{
		{
			`
type: cloudbeat/vanilla
streams:
  - data_yaml:
      activated_rules:
        cis_k8s:
          - a
          - b
`,
			[]string{"a", "b"},
			nil,
			"cloudbeat/vanilla",
		},
		{
			`
type: cloudbeat/eks
streams:
  - data_yaml:
      activated_rules:
        cis_eks:
          - a
          - b
`,
			nil,
			[]string{"a", "b"},
			"cloudbeat/eks",
		},
	}

	for _, test := range tests {
		cfg, err := config.NewConfigFrom(test.config)
		s.NoError(err)

		c, err := New(cfg)
		s.NoError(err)

		s.Equal(test.expectedType, c.Type)
		s.Equal(test.expectedActivatedRules, c.Streams[0].DataYaml.ActivatedRules.CisK8s)
		s.Equal(test.expectedEksActivatedRules, c.Streams[0].DataYaml.ActivatedRules.CisEKS)
	}
}
```

ofiriro3: What do you think about this option?

oren-zohar: it's the same, I see no point in changing it, just leave it

evaluator/bundle_test.go
```diff
@@ -129,13 +129,32 @@ func (s *BundleTestSuite) TestCreateServerWithDataYaml() {
 			"valid config from string", "/bundles/bundle.tar.gz", "200 OK", configWithDataYaml,
 		},
 		{
-			"valid config struct", "/bundles/bundle.tar.gz", "200 OK", config.Config{
+			"valid config struct", "/bundles/bundle.tar.gz", "200 OK",
+			config.Config{
+				Type: config.InputTypeVanillaK8s,
+				Streams: []config.Stream{
```
ofiriro3, line 135: Can you add a similar test to EKS ? (with ActivatedRules)

config/config_test.go
ofiriro3: I don't think that they are the same option, since the first option doesn't check that the other configuration is actually `Nil` and not an empty array or anything else.

It will just "lock" the behavior

oren-zohar: oh this is what you mean by lock ok

ofiriro3 (approved):
👍

PR #269, opened 2022-07-13
fixing the common.yml.tmpl

ofiriro3:
- fixes : https://github.com/elastic/security-team/issues/3041

### DIFF
--- a/_meta/config/cloudbeat.common.yml.tmpl
+++ b/_meta/config/cloudbeat.common.yml.tmpl
@@ -1,32 +1,51 @@
 cloudbeat:
   # Defines how often an event is sent to the output
   period: 4h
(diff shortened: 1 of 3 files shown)


eyalkraft:
currently looks like a local problem

eyalkraft:
The problem is that our build process (specifically `make PackageAgent`) changes the files:
```
	modified:   cloudbeat.reference.yml
	modified:   cloudbeat.yml
	modified:   docs/fields.asciidoc
```

Are you familiar with this behavior?

eyalkraft:
The problem is probably that I wasn't running with this change :)

ofiriro3:
@eyalkraft updated us that this fix solves his problem, he just had to solve something on his local machine

Issue #238, opened 2022-06-23
[BUG] Weird message in cloudbeat logs


uri-weisman:
**Describe the bug**
**"failed to find message"** appears multiple times in the cloudbeat dataset in the agent logs.
Most likely that some of the logs are not following the ECS convention and this results in Kibana failing to parse them [(discussion).](https://discuss.elastic.co/t/message-failed-to-find-message-in-kibana-logs/210522/2)

**Preconditions**
1. Run an elastic stack 
VERSION: 8.3.0
BUILD: 53457
COMMIT: ba70bdf9fe5d28e832c40b7dbdade1e582133af1

2. Install the CIS k8s integration.
3. Install elastic agent.

**To Reproduce**
1. Navigate to the elastic-agent logs.
2. Enable debug mode.
3. Observe the logs and watch for "failed to find message".

**Screenshots**
<img width="1207" alt="Screen Shot 2022-06-23 at 9 09 33" src="https://user-images.githubusercontent.com/68195305/175238772-de677851-f065-410f-a743-2948f62bb587.png">

eyalkraft:
@oren-zohar fixed by https://github.com/elastic/cloudbeat/pull/271

JordanSh:
QA Cycle 1: passed
OS: MacOS
Browser: Chrome
Kibana Version: 8.4 latest BC1
SHA: 58f7eaf0f8dc3c43cbfcd393e587f155e97b3d0d

uri-weisman:
In **debug mode** this message still appears.
Reproduced by:
1. Setting the log level of an agent to `debug` -> apply changes.
2. Filter by cloudbeat dataset.

<img width="1216" alt="Screen Shot 2022-08-15 at 18 01 24" src="https://user-images.githubusercontent.com/68195305/184660617-e62dc305-246b-42ee-9b94-386f3ac59d23.png">

oren-zohar:
**DoD update**
- [ ] make sure we don't miss anything else here

olegsu:
> @oren-zohar fixed by #271

There are more cases?

oren-zohar:
> > @oren-zohar fixed by #271
> 
> There are more cases?

We should ensure we don't see them anymore, @uri-weisman saw a few in debug mode, you can sync with him

kfirpeled:
@oren-zohar , @tehilashn is it still 8.5 candidate? if not, can we move it?

oren-zohar:
[...]

PR #271, opened 2022-07-13
Add message to our findings

eyalkraft:
closes https://github.com/elastic/security-team/issues/3786

Before:
Screen Shot 2022-07-13 at 18 13 34

Screen Shot 2022-07-13 at 18 14 18

After:
Screen Shot 2022-07-13 at 20 27 30

### DIFF
--- a/transformer/events_creator.go
+++ b/transformer/events_creator.go
@@ -19,6 +19,7 @@ package transformer
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/elastic/beats/v7/libbeat/beat"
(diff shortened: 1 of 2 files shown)


amitkanfer:
👑

uri-weisman (approved):
NIce! I guess that we need to update the findings mapping as well.

oren-zohar:
Does it close this?
- https://github.com/elastic/cloudbeat/issues/238

eyalkraft:
@oren-zohar It does for findings but I'm afraid maybe there are some other log messages cloudbeat produces the don't have a message field? wdyt?

oren-zohar:
I'll keep an eye on it for the next couple of days, if we don't see it anymore I'll close it. We can always re-open it.

Issue #270, opened 2022-07-13
Autodetect region


ofiriro3:
**Motivation**
As a developer, I would the cloudbeat to auto-detect its AWS region. 

**In-Depth**
Currently, our integration could not detect what is the region it operates from.
In this task, we will make cloudbeat autodetect what region it operates from.

**Related tasks**

**Definition of done**
- Cloudbeat is able to detect its AWS region.

PR #273, opened 2022-07-13
Add autodetect region to EKS

ofiriro3:
- Fixes: https://github.com/elastic/cloudbeat/issues/270


resources/fetchers/ecr_factory_test.go
```diff
@@ -104,3 +120,18 @@ default_region: us1-east
 		s.Equal(test.expectedRegex[1], ecrFetcher.PodDescribers[1].FilterRegex.String())
 	}
 }
+
+func CreateSdkConfig(config aws.ConfigAWS, region string) awssdk.Config {
```
ofiriro3, line 124: Maybe move it to a different place

DaveSys911:
Allure link: http://csp-allure-reports.s3.amazonaws.com/allure_reports/cloudbeat/autodetect-region/index.html

resources/providers/awslib/config_provider.go
```diff
@@ -18,23 +18,29 @@
...
 import (
-	"github.com/aws/aws-sdk-go-v2/aws/external"
+	"context"
+	awssdk "github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/elastic/beats/v7/x-pack/libbeat/common/aws"
 )
 
 type ConfigGetter interface {
-	GetConfig() Config
+	InitializeAWSConfig(ctx context.Context, cfg aws.ConfigAWS) (awssdk.Config, error)
```
oren-zohar, line 27: A few points about this interface:
- it only provides the AWS config, so it's not generic as the name suggests
- ideally, the interface and the implementation should be kept in separate files
- the implementation name is not consistent with the interface (getter vs provider)

I would expect each FW to implement the interface on its own level. This will also will allow us to avoid awkward naming situations like this. So instead of `ConfigProvider` that implements the `ConfigGetter` interface for the AWS use case, I would suggest a more general interface `ConfigProvider`, that will have `InitializeConfig` function instead of specific InitializeAWSConfig, and `AwsConfigProvider` will implement it.

resources/providers/awslib/config_provider.go
```diff
@@ -18,23 +18,29 @@
...
 }
 
 type ConfigProvider struct {
+	MetadataProvider MetadataGetter
```
oren-zohar, line 31: same. Getter vs Provider inconsistency.

resources/providers/awslib/ec2_metadata_getter_mock.go
```diff
@@ -0,0 +1,103 @@
...
+package awslib
+
+import (
+	context "context"
+
+	aws "github.com/aws/aws-sdk-go-v2/aws"
+
+	ec2metadata "github.com/aws/aws-sdk-go-v2/aws/ec2metadata"
+
+	mock "github.com/stretchr/testify/mock"
```
oren-zohar, line 29: redundant import naming?

resources/providers/awslib/ec2_metadata_getter_mock.go
```diff
@@ -0,0 +1,103 @@
...
+// MockMetadataGetter is an autogenerated mock type for the MetadataGetter type
+type MockMetadataGetter struct {
+	mock.Mock
+}
+
+type MockMetadataGetter_Expecter struct {
+	mock *mock.Mock
+}
+
+func (_m *MockMetadataGetter) EXPECT() *MockMetadataGetter_Expecter {
```
oren-zohar, line 41: I see we use `_Expecter` suffix in other tests this seems very out of go functions naming convention, right? this comment is relevant to all of the functions named here, but maybe I'm missing something.

resources/providers/awslib/ec2_metadata_provider.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/ec2metadata"
+)
+
+type Ec2MetadataProvider struct {
+}
+
+type Metadata = ec2metadata.EC2InstanceIdentityDocument
+
+type MetadataGetter interface {
```
oren-zohar, line 31:
```suggestion
type MetadataProvider interface {
```
I think it should be consistent across the repo, either use providers or getters and since we use mostly providers we should stick to it.

oren-zohar (changes requested):
Nice work

resources/providers/awslib/ec2_metadata_getter_mock.go
ofiriro3: This file is auto-generated by our mocking framework -https://github.com/vektra/mockery

resources/providers/awslib/ec2_metadata_getter_mock.go
ofiriro3: It's part of the auto-generated file - but I'll change

resources/fetchers/ecr_factory.go
```diff
@@ -42,12 +41,14 @@ func init() {
...
 		KubernetesProvider: providers.KubernetesProvider{},
 		IdentityProvider:   awslib.GetIdentityClient,
+		AwsConfigProvider:  awslib.ConfigProvider{MetadataProvider: awslib.Ec2MetadataProvider{}},
 	})
 }
 
 type ECRFactory struct {
 	KubernetesProvider providers.KubernetesClientGetter
 	IdentityProvider   func(cfg awssdk.Config) awslib.IdentityProviderGetter
+	AwsConfigProvider  awslib.ConfigGetter
```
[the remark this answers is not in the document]
oren-zohar: I was talking about the naming inconsistency, see my other comments as they explain in-depth.

resources/providers/awslib/ec2_metadata_getter_mock.go
oren-zohar: ignore my comments regarding the auto-generated file

resources/providers/awslib/config_provider.go
oren-zohar: > If I would think that this interface will be common to a few structs, my answer would have been deferred.

This is exactly the case I'm referring to, a generic provider interface for different providers, such as the EKS. I wouldn't expect to have all of the implementations in the same file as the interface.

> > the implementation name is not consistent with the interface (getter vs provider)
> 
> Totally agree - 💯 . Do you have a suggestion?

I would stick to _provider_

> Sounds like a good idea.
But I wonder, what is mutual to every config provider? 
The awsConfig provider will return different properties than the azureConfig provider and the GCPConfig provider and so on.
So if the **output** of the interface and the **input** of the interface will be the diffenert, we will use `interface{}` as the type of both input and output.
And then I wonder if there is any point in using the generalization.

They provide `Config` which can be extended. If they don't have anything in common, then I think there's no need for an interface in the first place, right?, just implement AwsConfigProvider.

ofiriro3: I think the usage of interfaces in these kinds of cases is very important since we are using third-party code (AWS libraries).

The interface creates segregation between the concrete AWS lib implementation (interface segregation principle) and it makes our code more testable.
The usage of an interface here, let me mock the provider so I'll be able to test the config provider and the ECR/ELB/EKS - factories.
Without using the interface it will enforce the test use the real AWS ec2 metadata provider.

Other than that, using interfaces makes our code less coupled to any concrete implementation.

oren-zohar: the reasons behind using an interface here are clear to me, my point was about the use of a generic one, i.e. `configProvider` versus a specific one, i.e. `AwsConfigProvider`, sorry for not being clear.

oren-zohar (changes requested):
Looks good, still need to rename `config_getter_mock.go`

oren-zohar (approved):
💯

PR #287, opened 2022-07-18
separate coverage job

oren-zohar:
Separates the coverage job from the unit tests since coverall website is not stable, and when it's down it will fail our unit tests job. It also should not be a required job.


DaveSys911 (approved):
Lgtm

PR #294, opened 2022-07-24
Add benchmark id field to schema

orouz:
Screen Shot 2022-07-24 at 12 25 38


oren-zohar (approved):
Should add tests as part of the `opa.go` tests, but there aren't any tests for it. If you want to add a test to `opa.go` that would be great but it's definitely too much to ask for such a small change

Issue #145, opened 2022-05-15
[BUG] Cloudbeat cluster leader election not working


jeniawhite:
**Describe the bug**
When deploying a three-node kind cluster, I've noticed that all of the three cloudbeat instances are sending Kubernetes findings and fetching data.

**Preconditions**
Three-node kind cluster.

**To Reproduce**
1. Deploy a three-node kind cluster.
2. Deploy elastic agent with cloudbeat daemonset.
3. Go to cloud posture -> findings -> filter k8s_object.
4. Look at the agent that is sending the findings and you will see that you are getting them from all instances.

**Expected behavior**
Only the leader instance should send the findings for k8s_object resources.

**Screenshots**
<img width="1074" alt="Screen Shot 2022-05-15 at 12 27 58" src="https://user-images.githubusercontent.com/12246907/168466277-529e3f99-6db5-40c7-860a-54c2103f8072.png">
<img width="1450" alt="Screen Shot 2022-05-15 at 12 30 24" src="https://user-images.githubusercontent.com/12246907/168466282-ffb56ad9-6b9e-4b77-975a-153954845389.png">

kfirpeled:
@jeniawhite, can you say if this happens when there's more than one master?
On our system flows we haven't defined a cluster to check it, worth adding a dedicated test for this one.

eyalkraft:
@kfirpeled leader election lease and k8s master node are unrelated.

The reason this bug is effecting the user is the decision we made to include `node id` in the resource id of kube-api resources. If we would have used only their uuid or their uuid combined with the cluster id all the collected findings for the same resources would have been squashed into a single latest finding per resource (per rule).

@amitkanfer This raises another concern for the case of the leader changing. All our kube-resource ids would change as a result (since the node Id used in their resource id is the id of the node where the leader agent was running on). Since we keep stale findings for a while this will result with a spike in resource number and then a decrease back to the real number.

Generally seems a bit problematic to have the resource Id of some Pod change based on who fetched the pod's info (this is not even related to where the pod is running but where the agent is running.

tehilashn:
This sounds very similar to https://github.com/elastic/security-team/issues/4125 although the outcome is that all the agents are sending k8s objects. When working on https://github.com/elastic/security-team/issues/4125 let's verify that it is indeed the same.

uri-weisman:
FIxed by:
- https://github.com/elastic/cloudbeat/pull/295

ofiriro3:
[...]

PR #295, opened 2022-07-24
Add leader election mechanism to cloudbeat

uri-weisman:
Adds a leader election mechanism to cloudbeat, introducing a new leader lease.

Fixes:
- https://github.com/elastic/security-team/issues/4125
- https://github.com/elastic/cloudbeat/issues/145


beater/cloudbeat.go
```diff
@@ -198,8 +207,8 @@ func (bt *cloudbeat) Run(b *beat.Beat) error {
 	}
 }
 
-func initRegistry(log *logp.Logger, cfg config.Config, ch chan fetching.ResourceInfo) (fetchersManager.FetchersRegistry, error) {
-	registry := fetchersManager.NewFetcherRegistry(log)
+func initRegistry(log *logp.Logger, cfg config.Config, ch chan fetching.ResourceInfo, le leaderelection.ElectionManager) (fetchersManager.FetchersRegistry, error) {
+	registry := fetchersManager.NewFetcherRegistry(log, le)
```
eyalkraft, line 211: It doesn't make a lot of sense to change these signatures for leader election

leaderelection/leaderelection.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	go m.leader.Run(m.ctx)
+	m.log.Infof("started leader election", "for", leaderElectionConf.Lock.Describe(), "id", leaderElectionConf.Lock.Identity())
+
+	return nil
+}
+
+func (m *Manager) IsLeader() bool {
+	return m.leader.IsLeader()
```
eyalkraft, line 89: so will this work properly what the manages is shared by many gorutines?
I think it will.
previous version seems to be not concurrent safe.

eyalkraft (changes requested):
The current design is a bit problematic since it seems like the leader election process manages data.Run which makes no sense.
Try and figure out a way to have the conditions mechanism to access it without propagating it that way.

leaderelection/leaderelection.go
```diff
@@ -0,0 +1,154 @@
...
+func currentPodID() string {
+	pod := os.Getenv(PodNameEnvar)
+
+	return lastPart(pod)
+}
+
+func lastPart(s string) string {
+	parts := strings.Split(s, "-")
+	if len(parts) == 0 {
+		return ""
```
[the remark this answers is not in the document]
uri-weisman: Let's generate a UUID to be used as an identifier in case the `pod.id` or env var doesn't exist...

leaderelection/leaderelection.go
```diff
@@ -0,0 +1,154 @@
...
+						m.log.Error(err)
+					}
+				})
+			},
+		},
+	}
+}
+
+func currentPodID() string {
+	pod := os.Getenv(PodNameEnvar)
```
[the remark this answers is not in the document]
uri-weisman: Let's generate a UUID to be used as an identifier in case the pod.id or env var doesn't exist...

leaderelection/k8s_leaderelection.go
```diff
@@ -0,0 +1,54 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+// Config is put into a different package to prevent cyclic imports in case
+// it is needed in several locations
+
```
uri-weisman, line 20: An abstraction layer on top of the leader election util, was created in order to be mockable in unit tests.
In the end, I decided not to use it as I couldn't mock the callbacks, yet, I still find it valuable for the future in case someone will want to mock this service.

leaderelection/leaderelection.go
```diff
@@ -0,0 +1,154 @@
...
+	return nil
+}
+
+func (m *Manager) IsLeader() bool {
+	return m.leader.IsLeader()
+}
+
+func (m *Manager) getConfig(onNewLeaderCb func(ctx context.Context) error) le.LeaderElectionConfig {
+	id := fmt.Sprintf("%s_%s", LeaderLeaseName, currentPodID())
+	ns, err := kubernetes.InClusterNamespace()
```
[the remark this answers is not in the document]
uri-weisman: I noticed that both the agent and the native k8s go client implementations use the `default` namespace as a fallback.
WDYT?

leaderelection/leaderelection.go
```diff
@@ -0,0 +1,184 @@
...
+	return manager
+}
+
+func (m *Manager) IsLeader() bool {
+	return m.leader.IsLeader()
+}
+
+// Run leader election is blocking until a leader is being elected or timeout has reached.
+func (m *Manager) Run() error {
+	go m.leader.Run(m.ctx)
```
eyalkraft, line 96: This should be an argument ctx and not the m.ctx
I'm not sure m.ctx is needed

leaderelection/leaderelection.go
```diff
@@ -0,0 +1,184 @@
...
+		return nil, err
+	}
+
+	initOnce.Do(func() {
+		var leConfig le.LeaderElectionConfig
+		var leader K8SLeaderElectionService
+
+		block := make(chan bool)
+		leConfig, err = buildConfig(ctx, log, kubeClient, block)
+		leader, err = NewK8sLeaderElector(leConfig)
```
eyalkraft, line 73: a bit confusing because of the abstraction layer... not sure it's worth it if we're not using it

eyalkraft: yeah, pretty sure the best approach for now will be to get rid of it

leaderelection/leaderelection.go
```diff
@@ -0,0 +1,184 @@
...
+			},
+			OnStoppedLeading: func() {
+				log.Infof("leader election lock LOST, id: %v", id)
+				// leaderelection.Run stops in case it has stopped holding the leader lease, we re-run the manager
+				// to keep following leader status.
+				go manager.leader.Run(ctx)
+				log.Infof("re-running leader elector")
+			},
+			OnNewLeader: func(identity string) {
+				log.Infof("leader election lock has been acquired, id: %v", identity)
```
eyalkraft, line 150: acquired by another cloudbeat

eyalkraft: right? because otherwise `OnStartedLeading` is called

leaderelection/leaderelection_test.go
```diff
@@ -0,0 +1,113 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package leaderelection
+
```
eyalkraft, line 19: hmm ok I see now you do use the abstraction layer so keep it.

eyalkraft (approved):
Small changes required.
Remember to test also reconfiguring the integration.

leaderelection/leaderelection.go
uri-weisman: Both will be called, first the `OnNewLeader` and then the `OnStartedLeading` cb.

beater/cloudbeat.go
```diff
@@ -135,13 +144,17 @@ func NewCloudbeat(_ *beat.Beat, cfg *agentconfig.C) (beat.Beater, error) {
...
 		resourceCh:  resourceCh,
+		leader:      le,
 	}
 	return bt, nil
 }
 
 // Run starts cloudbeat.
 func (bt *cloudbeat) Run(b *beat.Beat) error {
 	bt.log.Info("cloudbeat is running! Hit CTRL-C to stop it.")
+	if err := bt.leader.Run(); err != nil {
```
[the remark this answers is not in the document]
uri-weisman: I already have a timeout, `FirstLeaderDeadline`, that should handle the scenario you mentioned.
If you don’t like the approach I can pass a context with a timeout, but I won’t pass it to the leader election util as it should run as long as cloudbeat runs.

leaderelection/leaderelection_test.go
uri-weisman: Not really, removed.

beater/cloudbeat.go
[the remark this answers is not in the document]
eyalkraft: yeah, sounds good 👍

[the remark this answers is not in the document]
eyalkraft: even keep `FirstLeaderDeadline` as an additional case that's ok

tehilashn:
@uri-weisman - cloudbeat 8.4 branch was already created (https://github.com/elastic/cloudbeat/tree/8.4 ) so we should backport this one

DaveSys911 (approved):
LGTM - if you can please check it works on eks as well.

leaderelection/leaderelection.go
eyalkraft: add an `if identity == id` and write different and proper log messaged for these to cases (acquired by another, acquired by me)

leaderelection/leaderelection.go
```diff
@@ -141,10 +141,6 @@ func buildConfig(ctx context.Context, log *logp.Logger, client k8s.Interface, bl
 			},
 			OnStoppedLeading: func() {
 				log.Infof("leader election lock LOST, id: %v", id)
-				// leaderelection.Run stops in case it has stopped holding the leader lease, we re-run the manager
-				// to keep following leader status.
-				go manager.leader.Run(ctx)
-				log.Infof("re-running leader elector")
```
eyalkraft: I'm not super sure about this..
Removing this means we're missing the edge case of when cloudbeat is losing the leadership for some reason which is not container restart/cloudbeat reconfiguration (for example, network issues).
@uri-weisman wdyt exactly happened when this went crazy on your test?
were the leaderelector started multiple times?

yashtewari:
Note that currently the [cloudbeat design document](https://github.com/elastic/security-team/blob/main/docs/cloud-security-posture-team/cloudbeat-design.mdx) briefly describes leader election; IMO it would be worthwhile to update it with more details of these changes when they are merged.

leaderelection/leaderelection.go
```diff
@@ -56,6 +56,9 @@ type Manager struct {
 
 // NewLeaderElector acts as a singleton
 func NewLeaderElector(ctx context.Context, log *logp.Logger, cfg config.Config) (ElectionManager, error) {
+	// the launcher is creating a new beater instance every new configuration.
+	// therefore, we must reset callOnce every new beater to stale the fetching cycle until we have the leader information.
```
eyalkraft, line 60:
```suggestion
	// therefore, we must reset callOnce every new beater to stall the fetching cycle until we have the leader information.
```

eyalkraft (approved):
Looks good, has to be thoroughly tested 👍

PR #298, opened 2022-07-24
Fix-pre-merge-ci

DaveSys911:
- Remove tmate actions
- Extend test run time to 55 mins to accommodate API tests
- add concurrency to limit executions to one per pr


oren-zohar:
> add concurrency to limit executions to one per pr

What are the reasons for doing so?

oren-zohar (approved):

PR #302, opened 2022-07-26
Fix broken k8s_object_rules tests.

yashtewari:
Most of the problems occur in resource-matching logic that tries to find the right ES event to compare the expected evaluation against, which is harder to do for K8S resources than other resource types.

Inserting a per-testcase UUID into the K8S object's metadata that is being tested allows for much simpler resource-matching than the recursion logic in the [`dict_contains`](https://github.com/elastic/cloudbeat/blob/298e0d2e1cb0c7414f16bf5ea287f1bb5bc4850b/tests/commonlib/utils.py#L87) utility.

Resolves: https://github.com/elastic/security-team/issues/4312

### DIFF
--- a/tests/product/tests/test_k8s_objects_rules.py
+++ b/tests/product/tests/test_k8s_objects_rules.py
@@ -3,12 +3,12 @@ Kubernetes CIS rules verification.
 This module verifies correctness of retrieved findings by manipulating audit and remediation actions
 """
 from datetime import datetime
+import uuid
 
 import pytest
 
-# from product.tests.data.k8s_object.k8s_object_rules import *
 from product.tests.data.k8s_object import k8s_object_rules as k8s_tc
-from commonlib.utils import get_ES_evaluation, get_resource_identifier
+from commonlib.utils import get_ES_evaluation
 from commonlib.framework.reporting import skip_param_case, SkipReportData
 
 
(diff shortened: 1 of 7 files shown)


tests/product/tests/test_k8s_objects_rules.py
```diff
@@ -92,13 +87,19 @@ def test_kube_resource_patch(elastic_client, test_env, rule_tag, resource_type,
         raise ValueError(
             f'Could not patch resource type {resource_type}:'
             f' {relevant_metadata} with patch {resource_body}')
-    
+
+    def match_resource(eval_resource):
+        try:
+            return eval_resource.metadata.labels.test_resource_id == test_resource_id
+        except AttributeError:
+            return False
```
oren-zohar, line 95: this function can be implemented outside of this function

oren-zohar (changes requested):
Looks good, one comment

tests/product/tests/test_k8s_objects_rules.py
```diff
@@ -82,6 +72,11 @@ def test_kube_resource_patch(elastic_client, test_env, rule_tag, resource_type,
 
     assert resource, f"Resource {resource_type} not found"
 
+    test_resource_id = str(uuid.uuid4())
+    labels = resource_body['metadata'].get('labels', {})
+    labels['test_resource_id'] = test_resource_id
+    resource_body['metadata']['labels'] = labels
```
oren-zohar, line 78:
```suggestion
    metadata.setdefault('labels', {})
    test_resource_id = str(uuid.uuid4())
    metadata['labels']['test_resource_id'] = test_resource_id
```

tests/product/tests/test_k8s_objects_rules.py
```diff
@@ -3,12 +3,13 @@
 This module verifies correctness of retrieved findings by manipulating audit and remediation actions
 """
 from datetime import datetime
+import uuid
 
 import pytest
 
 # from product.tests.data.k8s_object.k8s_object_rules import *
```
oren-zohar, line 10: delete

oren-zohar (changes requested):

oren-zohar (approved):

gurevichdmitry (approved):
LGTM, great work!

PR #305, opened 2022-07-26
Removing public ecr from ecr fetcher

ofiriro3:
- fixes: https://github.com/elastic/security-team/issues/4035


oren-zohar (approved):
LGTM 🌮

PR #318, opened 2022-08-01
Launcher recovery on beater panic

amirbenun:


yashtewari (approved):
Nice

launcher/launcher.go
```diff
@@ -132,6 +132,11 @@ func (l *launcher) runBeater() error {
 	l.wg.Add(1)
 	go func(beater beat.Beater) {
 		l.log.Info("Launcher is running the new created Beater")
+		defer func() {
```
uri-weisman, line 135: Very nice!
Do we count on the agent to re-run cloudbeat or do we want to apply a similar logic of our own?

yashtewari: At the moment `type launcher` functionality is reserved for re-instantiating `type cloudbeat` with configuration updates.

It makes more sense to propagate the error up to the Agent and let the Agent restart the cloudbeat binary as a whole. This lets the Agent do its own handling on the error (at least conceptually), and it's a more complete restart.

uri-weisman: That sounds right.
Is it actually happening? propagating an error to the agent results in a cloudbeat restart?

yashtewari: Yes, you can observe a cloudbeat restart loop on the Agent if you return an error from the `Run` method.

Issue #235, opened 2022-06-22
[BUG] New os user/group doesn't reflect in Cloudbeat


uri-weisman:
**Describe the bug**
Remediation that requires creating a new OS user/group won't be presented to the user even though it took effect.

**Preconditions**
A Containerized cloudbeat using a Kind cluster.

**To Reproduce**
1. Verify CIS rule 1.1.12 is failing the evaluation.
2. Apply remediation:
4. SSH to host (kind cluster)
5. `groupadd etcd`
6. `useradd -g etcd etcd`
7. `chown etcd:etcd /var/lib/etcd/`
8. Wait a cycle for new results to be presented
9. CIS rule 1.1.12 is still failing the evaluation.

**Expected behavior**
CIS rule 1.1.12 is should pass as we apply the right remediation.

**Investigation conclusions**
1. group and user info can be found in `/etc/group` & `/etc/passwd` respectively.
2. In order to propagate this data to cloudbeat we mount those files to the elastic-agent.
3. In the case of a missing user/group we use the `useradd`/`groupadd` commands.
4. Those commands edit the mentioned files and modify their inodes.
5. When you mount an individual file into a container, what you are mounting is the inode of the file on the FS ([discussion](https://unix.stackexchange.com/questions/44373/single-bind-mounted-file-gets-out-of-sync-in-linux)).
As a result, the files inside the container are not synced.

**Proposed solution**
Mounting an entire directory will solve the issue because the directory's pointer to the inode gets updated. 
Having said that, we don't want to mount the entire `etc` directory, therefore, we can create a new folder to mount and symlink the mentioned files.

**Screenshots**

`useradd` cmd modify `/etc/passwd` file inode.

<img width="147" alt="Screen Shot 2022-06-22 at 17 31 43" src="https://user-images.githubusercontent.com/68195305/175076199-d51562d5-21ce-460d-8444-bdf70bef62e8.png">

**Related issue:** 
- #118

tinnytintin10:
I have prioritized this as a 0 @tehilashn the definition of done should be that remediations should show up as passing and there should be a test developed for this.

yashtewari:
Very thorough, Uri! This helped me while working on
- https://github.com/elastic/security-team/issues/4311

[...]

PR #320, opened 2022-08-02
Fix broken file_system_rules tests.

yashtewari:
Resolves https://github.com/elastic/security-team/issues/4311

### DIFF
--- a/tests/commonlib/io_utils.py
+++ b/tests/commonlib/io_utils.py
@@ -2,14 +2,16 @@
 This module provides input / output manipulations on streams / files
 """
 
-from datetime import datetime
-import os
 import io
 import json
+import os
 import shutil
-from pathlib import Path
+import subprocess
 import yaml
+
+from datetime import datetime
 from munch import Munch, munchify
+from pathlib import Path
 
 
 def get_events_from_index(elastic_client, index_name: str, rule_tag: str, time_after: datetime) -> list[Munch]:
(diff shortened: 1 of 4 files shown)


DaveSys911 (approved):
approved as a temp solution until https://github.com/elastic/cloudbeat/issues/235  is resolved.

tests/commonlib/io_utils.py
```diff
@@ -129,17 +131,62 @@ def exec_command(container_name: str, command: str, param_value: str, resource:
...
-            if len(uid_gid) != 2:
-                raise Exception(
-                    "User and group parameter shall be separated by ':' ")
-            shutil.chown(path=resource, user=uid_gid[0], group=uid_gid[1])
+            try:
+                uid, gid = param_value.split(':')
+            except ValueError as exc:
+                raise Exception("User and group parameter shall be separated by ':' ") from exc
+
+            FsClient.add_users_to_node([uid, gid], in_place=True)
```
gurevichdmitry, line 139: @yashtewari, add_users_to_node will be executed always. Can we solve this in more elegant way? For example to ask either user /group exists and if not to call the function.

tests/commonlib/io_utils.py
```diff
@@ -129,17 +131,62 @@ def exec_command(container_name: str, command: str, param_value: str, resource:
...
+            shutil.copyfile(host_groups_file, temp_groups_file)
+
+            for user in users:
+                # These commands fail silently for users/groups that exist.
+                subprocess.run(['groupadd', user, '-P', '/tmp'], capture_output=True)
+                subprocess.run(['useradd', user, '-g', user, '-P', '/tmp'], capture_output=True)
+                subprocess.run(['useradd', user], capture_output=True) # For container to get around chmod check.
+
+            FsClient.in_place_copy(temp_users_file, host_users_file)
+            FsClient.in_place_copy(temp_groups_file, host_groups_file)
```
gurevichdmitry, line 177: @yashtewari why do we need a new method that just copies the content?
shutil has several functions to copy data:
copy()
copy2()
copyfile()
I think we can find appropriate function for our case.

tests/commonlib/io_utils.py
```diff
@@ -129,17 +131,62 @@ def exec_command(container_name: str, command: str, param_value: str, resource:
...
+            host_groups_file = Path('/hostfs/etc/group')
+
+            temp_etc = Path('/tmp/etc')
+            temp_etc.mkdir(parents=True, exist_ok=True)
+
+            temp_users_file = temp_etc / 'passwd'
+            temp_groups_file = temp_etc / 'group'
+
+            shutil.copyfile(host_users_file, temp_users_file)
+            shutil.copyfile(host_groups_file, temp_groups_file)
```
gurevichdmitry, line 168: Seems like we perform same operations for host_users_file and host_group_file. We can decrease number of lines by executing for loop on list of paths

gurevichdmitry (commented):
@yashtewari - great job.
Please review my comments.

tests/commonlib/io_utils.py
yashtewari: We can, but it would be over optimization. We have much more critical optimizations to do at the moment, compared to saving a few milliseconds here.

tests/commonlib/io_utils.py
yashtewari: A loop of two iterations doesn't seem much better/cleaner to me than just running those two iterations.

tests/commonlib/io_utils.py
yashtewari: All of these methods replace the existing file at the destination, which is undesirable for testing this flow. Check out [this comment](https://github.com/elastic/security-team/issues/4311#issuecomment-1202182095) for more on why we want to test the flow with an in-place copy.

tests/commonlib/io_utils.py
```diff
@@ -129,17 +131,62 @@ def exec_command(container_name: str, command: str, param_value: str, resource:
...
+            shutil.chown(path=resource, user=uid, group=gid)
         elif command == 'unlink':
             if not Path(param_value).is_dir():
                 Path(param_value).unlink()
         else:
             raise Exception(
                 f"Command '{command}' still not implemented in test framework")
+    
+    @staticmethod
+    def add_users_to_node(users: list, in_place: bool):
```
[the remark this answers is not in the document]
yashtewari: Agreed, with the caveat that it's one of the lower priorities things to clean up. At the moment we never test whether or not a certain user or group exists. It's more of a pre-condition for some of the tests, and it wouldn't corrupt any of the current tests if a given user or group was already there.

gurevichdmitry (approved):
@yashtewari, approved.
Please open task for optimization that is not critical for now and may be improved later.

PR #324, opened 2022-08-03
Add integration test for leader election

uri-weisman:
This PR introduces a new integration test that should verify the leader election mechanism.
At the moment, it will run on a single node cluster, later we plan to modify the CI to support a multi-node cluster.


tests/integration/tests/conftest.py
```diff
@@ -44,10 +43,32 @@ def fixture_data(start_stop_cloudbeat, k8s, cloudbeat_agent):
...
                                        namespace=cloudbeat_agent.namespace)
     nodes = k8s.get_cluster_nodes()
-    return pods, nodes
+    leader_node = get_cluster_leader(k8s, cloudbeat_agent, pods)
+
+    return pods, nodes, leader_node
+
+
+def get_cluster_leader(k8s, cloudbeat_agent, pods):
+    lease_name = "cloudbeat-cluster-leader"
```
gurevichdmitry, line 58: These tests shall be relevant for cloudbeat and elastic agent. If lease_name will not be the same please find better solution.

tests/integration/tests/conftest.py
```diff
@@ -44,10 +43,32 @@ def fixture_data(start_stop_cloudbeat, k8s, cloudbeat_agent):
...
+                          namespace=cloudbeat_agent.namespace)
+
+    lease_info = k8s.get_resource(resource_type="Lease", name=lease_name, namespace=cloudbeat_agent.namespace)
+    lease_holder_identity = lease_info.spec.holder_identity
+    holder_id = lease_holder_identity.split("_")[-1]
+
+    leader_node = ""
+    for pod in pods:
+        if holder_id in pod.metadata.name:
+            leader_node = pod.spec.node_name
```
gurevichdmitry, line 72: after finding leader_node are we required to continue 'for' loop? If not, you can stop iteration by break command

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -55,6 +55,55 @@ def test_elastic_index_exists(elastic_client, match_type):
...
+                                           query=query,
+                                           size=1000,
+                                           sort=sort)
+    # checking that k8s_objects are being sent only by the leader node.
+    for resource in result['hits']['hits']:
+        assert leader_node == resource['_source']['agent']['name'], \
+            f"Multiple agents send k8s_objects"
+
+
+def build_es_query(match_type):
```
gurevichdmitry, line 106: 1. This function shall be a part of commonlib -> elastic wrapper.
2. Missing docstring -> please provide function description.

tests/integration/tests/conftest.py
```diff
@@ -44,10 +43,32 @@ def fixture_data(start_stop_cloudbeat, k8s, cloudbeat_agent):
...
     @param cloudbeat_agent: config object
-    @return: pods, nodes in cluster
+    @return: pods, nodes, leader_node in cluster
     """
     # pylint: disable=W0612,W0613
     pods = k8s.get_agent_pod_instances(agent_name=cloudbeat_agent.name,
                                        namespace=cloudbeat_agent.namespace)
     nodes = k8s.get_cluster_nodes()
-    return pods, nodes
+    leader_node = get_cluster_leader(k8s, cloudbeat_agent, pods)
```
gurevichdmitry, line 52: Why do you need leader information _before_ test case execution? You can just pass k8s object to the test function and retrieve the data inside the test case.

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -22,10 +22,10 @@ def test_cloudbeat_pod_exist(fixture_data):
     :param fixture_data: (Pods list, Nodes list)
     :return:
     """
-    pods, nodes = fixture_data
+    pods, nodes, leader_node = fixture_data
```
gurevichdmitry, line 25: leader_node is not used in cloudbeat_pod_exist function.

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -55,6 +55,55 @@ def test_elastic_index_exists(elastic_client, match_type):
...
+@pytest.mark.dependency(depends=["test_cloudbeat_pod_exist"])
+def test_leader_election(fixture_data, elastic_client):
+    """
+    This test verifies that findings of all types are sending to elasticsearch
+    :param elastic_client: Elastic API client
+    :return:
+    """
+
+    query, sort = build_es_query("k8s_object")
+    pods, nodes, leader_node = fixture_data
```
gurevichdmitry, line 90: so in this case fixture data is unpacked - pods and nodes are not used in this test function.

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -55,6 +55,55 @@ def test_elastic_index_exists(elastic_client, match_type):
...
+        time.sleep(1)
+
+    assert len(result) > 0, \
+        f"The findings of type {match_type} not found"
+
+
+@pytest.mark.pre_merge
+@pytest.mark.order(4)
+@pytest.mark.dependency(depends=["test_cloudbeat_pod_exist"])
+def test_leader_election(fixture_data, elastic_client):
```
gurevichdmitry, line 82: instead of passing fixture data you can pass k8s client and retrieve leader node information.

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -55,6 +55,55 @@ def test_elastic_index_exists(elastic_client, match_type):
...
+    """
+    This test verifies that findings of all types are sending to elasticsearch
+    :param elastic_client: Elastic API client
+    :return:
+    """
+
+    query, sort = build_es_query("k8s_object")
+    pods, nodes, leader_node = fixture_data
+    assert leader_node != "", \
+        f"The Leader node could not be found"
```
gurevichdmitry, line 92: f-string does not have any interpolated variables-> f usage is not required in this case.

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -55,6 +55,55 @@ def test_elastic_index_exists(elastic_client, match_type):
...
+
+    time.sleep(CONFIG_TIMEOUT)  # Wait for all pods to send resources to ES
```
gurevichdmitry, line 94: I'd like to prevent usage of time.sleep. Can we implement more elegant solution?

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -55,6 +55,55 @@ def test_elastic_index_exists(elastic_client, match_type):
...
+
+    result = elastic_client.get_index_data(index_name=elastic_client.index,
+                                           query=query,
+                                           size=1000,
+                                           sort=sort)
+    # checking that k8s_objects are being sent only by the leader node.
+    for resource in result['hits']['hits']:
+        assert leader_node == resource['_source']['agent']['name'], \
+            f"Multiple agents send k8s_objects"
```
gurevichdmitry, line 103: Why did you put assert inside for loop? Assert will break execution of 'for' loop on failure with a common message and you will not know where is the issue.
So if by design the test shall be interrupted, please format failure string (line 103) and provide more information using variables.
If you want to continue and iterate over all resources, you need to collect them to a list/dictionary (instead of assert), and then to perform assert once to check if a list/dictionary is empty.

gurevichdmitry (changes requested):
@uri-weisman, looks great.
1. Please use pylint and fix suggestions.
2. Several functions shall be moved to common lib and redesigned.

tests/integration/tests/conftest.py
uri-weisman: It will be the same for both.

tests/integration/tests/test_cloudbeat_integration.py
uri-weisman: I'm fine with stopping the test execution, I'll make the log more comprehensive.

tests/integration/tests/test_cloudbeat_integration.py
uri-weisman: I would prefer a different solution but I couldn't think of a way to verify that all agents finished sending their findings.
We can check that an agent sent at least a single finding but it would not ensure that all of them were sent.

tests/commonlib/elastic_wrapper.py
```diff
@@ -39,8 +40,8 @@ def get_total_value(data: dict) -> int:
         @param data: Data dictionary from elasticsearch
         @return: Total Value integer
         """
-        ret_value = data.get('hits', {})\
-            .get('total', {})\
+        ret_value = data.get('hits', {}) \
```
uri-weisman, line 43: Probably has been modified as part of running pylint

tests/integration/tests/test_cloudbeat_integration.py
```diff
@@ -94,5 +72,37 @@ def test_elastic_index_exists(elastic_client, match_type):
...
+    time.sleep(CONFIG_TIMEOUT)  # Wait for all pods to send resources to ES
+
+    result = elastic_client.get_index_data(index_name=elastic_client.index,
+                                           query=query,
+                                           size=1000,
+                                           sort=sort)
+    # checking that k8s_objects are being sent only by the leader node.
+    for resource in result['hits']['hits']:
+        assert leader_node == resource['_source']['agent']['name'], \
+            f"Multiple agents send k8s_objects, leader: {leader_node}, resource: {resource['_source']}"
```
gurevichdmitry, line 107: nit, failure message will display very long string with resource['_source'], because the last one returns big dictionary. Did you tried to simulate failure and check how it seems in the report? 
The simplest way to do that is set leader_node = "some value". Please verify that the message is readable.
if not you can  print resource['_source']['agent']['name'] and then to attach resource to allure report ( example can be found in test_elastic_index_exists test case).

gurevichdmitry (commented):
@uri-weisman, great job.
Can you please give a try to replace sleep timeout. If it is difficult we need to open a task for test improvement, and then we can close this PR.

tests/integration/tests/test_cloudbeat_integration.py
uri-weisman: Sure, but it won't ensure you that all agents finished sending their findings (k8s-related findings in particular).
Having said that, it might be sufficient.

tests/integration/tests/test_cloudbeat_integration.py
uri-weisman: TY @gurevichdmitry, the message is readable.

tests/commonlib/utils.py
```diff
@@ -110,3 +112,43 @@ def resource_identifier(resource):
...
+    @return: true if all agents finished sending findings in the configured timeout
+    """
+    required_cycles = 2
+    start_time = time.time()
+    prev_cycle_id = ""
+    curr_cycle_id = ""
+    active_agents = 0
+    num_cycles = 0
+
+    while num_cycles < required_cycles:
```
gurevichdmitry, line 133: @uri-weisman, num_cycles is iterated only if new cycle_id is retrieved. In case cloudbeat stops sending findings this will lead to infinite loop.

gurevichdmitry (changes requested):
@uri-weisman, cool implementation.
Sorry for delayed review.
It seems that we have infinite loop if no findings are sent isn't it?

tests/commonlib/utils.py
uri-weisman: Do we have a timeout per test?
Nevertheless, I can add a timeout to this function.

gurevichdmitry: currently test framework does not support timeouts per tests.

uri-weisman: I've added a timeout to the function

gurevichdmitry (approved):
@uri-weisman great work! LGTM

PR #335, opened 2022-08-14
DRA job to produce both snapshot & staging artifact by default

DaveSys911:
https://github.com/elastic/security-team/issues/4292
example of successful run 
https://internal-ci.elastic.co/blue/organizations/jenkins/cloudbeat%2Fcloudbeat-mbp/detail/PR-335/14/pipeline

DRA publishing seems to work https://artifacts-snapshot.elastic.co/cloudbeat/8.5.0-d3a61940/summary-8.5.0-SNAPSHOT.html

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ pipeline {
     ES_LOG_LEVEL = "${params.ES_LOG_LEVEL}"
     DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
     DOCKER_REGISTRY = 'docker.elastic.co'
-    WORKFLOW = "${params.build_type}"
+    // WORKFLOW = "${params.build_type}"
   }
   options {
     timeout(time: 2, unit: 'HOURS')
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -33,7 +33,7 @@ pipeline {
     booleanParam(name: 'test_ci', defaultValue: false, description: 'Enable test')
     booleanParam(name: 'release_ci', defaultValue: true, description: 'Enable build the release packages')
     booleanParam(name: 'its_ci', defaultValue: true, description: 'Enable async ITs')
-    choice(name: 'build_type', choices: ['snapshot', 'staging'], description: 'Choose Snapshot or Staging Build type(Default: Snapshot)')
+    choice(name: 'build_type', choices: ['both', 'snapshot', 'staging'], description: 'Choose Snapshot or Staging Build type(Default: Both)')
```
gurevichdmitry, line 36: nit, 'both' option is not used as expression { params.build_type != 'xxx' } and this trick will work because 2 options are implemented 'snapshot' and 'staging'. Adding additional option will require more logic to be implemented.

gurevichdmitry (approved):
LTGM

Jenkinsfile
DaveSys911: Yes, there are only 2 types of dra builds though. Snapshot & Staging.
So i either want both,one of them or none(and then we don't run that stage).

Issue #242, opened 2022-06-27
[BUG] Cloudbeat deployed in fleet manager elastic-agent period change doesn't work


jeniawhite:
**Describe the bug**
In order to change the period of cloudbeat, I've attempted to run the following script:
- https://github.com/elastic/cloudbeat/blob/main/scripts/remote_edit_config.sh
 
After editing the file in the elastic agent and restarting the go process of cloudbeat inside the agent (via an edit to the integration), I did not see any new findings being sent.

**Preconditions**
Cloud environment with 8.3 BC5 deployed
Local fleet managed elastic agent

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Run the script
2. Edit the integration (change the name or add description)
3. Fleet server should reboot the elastic agent's cloudbeat process (you will see a new PID)

**Expected behavior**
I should be able to see new findings in the new time period that was set

amirbenun:
Update:
It seems like cloudbeat process doesn't restart anymore after a configuration change (as it used to do).
Therefore, in order the new configuration to apply we need to make sure the process restarts by running:
```
kill -9 `pidof cloudbeat`
```

The next time cloudbeat restarts it will read the updated file and act accordingly.

PR #337, opened 2022-08-14
Remote config editing fixes

amirbenun:
Resolves https://github.com/elastic/cloudbeat/issues/242

This PR changes the way the `remote_edit_config.sh` works.
In order to find the configuration file path:
Instead of assuming the architecture and the version being run remote, we iteratively find the next folder name by its expected prefix.


ofiriro3:
Can you please apply these changes to the `remote_replace_cloudbeat.sh` script as well?

amirbenun:
Verified on a multi-node cluster.

Issue #336, opened 2022-08-14
[BUG] Leader election doesn't work correctly on a used environment


ofiriro3:
**Describe the bug**
When a Kubernetes cluster has an old Cloudbeat `lease`, the next cycle of any `Cloudbeat` instances will contain no K8s objects.

**Short background**
Every time interval(15s), each of the running Cloudbeat instances tries to acquire the `lease` object.
The Cloudbeat that acquired the `lease` will be the only Cloudbeat instance from this `daemonset` that will fetch the K8s object.
More can be found here - [leader election](https://github.com/elastic/security-team/issues/4125)

**In-depth**
When degrading cloudbeat instances, the old `lease` object is not deleted. 
Then when a new deployment is starting, no Cloudbeat instance will fetch K8s objects since the lease is still "attached" to the old Cloudbeat instance that was deleted in the last deployment.

When the new deployment is starting, at some point the `lease` will be acquired by one of the new Cloudbeat instances - but sometimes no Cloudbeat instance acquires the `lease` before the starting of the first cycle. 

**Preconditions**
- A running Kubernetes cluster

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Run the CSP integration on your Kubernetes cluster (Spawn a new agent with the CSP integration).
2. Delete the agent.
3. Run the agent again.
4. Go to the findings page, and you will see that no `K8s` objects were fetched in this cycle.

**Expected behavior**
One of the `daemonset` Cloudbeat instances should fetch the K8s objects.

**Desktop (please complete the following information):**
 - OS: Mac darwin
 - Browser: Chrome
 - Kibana Version: commit 0011c16d2b4b19b169bd801766cb4db8792a2adc

**Additional context**
I discuss this bug with @uri-weisman and he will investigate more.

eyalkraft:
Thanks for opening @ofiriro3!

Some relevant questions:
1. Does this reproduce if you wait 15+ seconds between the termination of the agent to the deployment of the next one?
2. How Is the UX effected by this behavior? In this case the user already has k8s objects findings (and latest). 
3. Given there is an old but valid lease when cloudbeat starts, What are the suggested solutions? Deleting the lease on a shutting down cloudbeat is possible but will probably only result with another race (the lease of a new holder is deleted by an old terminating agent, etc…)

ofiriro3:
Hi, @eyalkraft thanks for addressing this issue so fast.

1. Yes.
[...]

PR #338, opened 2022-08-15
Modify leader election mechanism block timeout

uri-weisman:
- Fixes: https://github.com/elastic/cloudbeat/issues/336
- Improve leader election retry mechanism.


leaderelection/leaderelection.go
```diff
@@ -91,25 +88,19 @@ func (m *Manager) IsLeader() bool {
...
 
-	select {
-	case <-m.block:
-		m.log.Infof("new leader has been elected")
-	case <-timeoutCtx.Done():
-		m.log.Warnf("timeout - no leader has been elected for %s", FirstLeaderDeadline.String())
-	}
-
-	return nil
+	<-timeoutCtx.Done()
```
amirbenun, line 99: You can use `time.After` as the context has no true usage.

amirbenun: Following question:
Does the call to `go m.leader.Run(ctx)` move the `ctx` context to "done" when it finishes?
If so, we can maintain the select syntax we had before, and add the `time.After` as another case.

leaderelection/leaderelection.go
```diff
@@ -158,11 +149,6 @@ func buildConfig(ctx context.Context, log *logp.Logger, client k8s.Interface, bl
 				} else {
 					log.Infof("leader election lock has been acquired by another pod, id: %v", identity)
 				}
-				// called every new beater
-				callOnce.Do(func() {
```
amirbenun, line 162: `callOnce` can be completely removed from the file.

amirbenun: Same goes for `retryOnce`

amirbenun: Can you explain why do we need `initOnce`?
What happens after a reconfiguration event from the fleet?

amirbenun (changes requested):
A few changes and questions

leaderelection/leaderelection.go
uri-weisman: As described in the leaderelection file,`retryOnce` is used to re-run the manager in case the leader stopped holding the lease, for example, the pod still exists but the leader has lost the lease due to connectivity issues.

uri-weisman: `initOnce` is been used to implement a singleton, we want only a single instance of the leader election to exist upon cloudbeat creation.
After a reconfiguration event we call `NewLeaderElector` but the same instance is returned.
A singleton is needed in this case as multiple leader election instances can cause races, especially when we count on the leader election data to trigger the fetching cycle.

leaderelection/leaderelection.go
uri-weisman: It will not change the status of our ctx, as a new ctx is created in the leaderelection package, and canceling it will only be propagated down the context hierarchy.

leaderelection/leaderelection.go
amirbenun: The `retryOnce` is not required since cloudbeat should try to acquire the lease over and over again and there is no reason to stop after two times.

leaderelection/leaderelection.go
amirbenun: So let's make sure that's indeed the case, and if so, there is no advantage of using `time.After`, instead we can just use `time.Sleep`

leaderelection/leaderelection.go
```diff
@@ -35,82 +35,82 @@ import (
...
+	time.Sleep(FirstLeaderDeadline)
+	m.log.Infof("stop waiting after %s for a leader to be elected", FirstLeaderDeadline)
+
 	return nil
 }
 
-func buildConfig(ctx context.Context, log *logp.Logger, client k8s.Interface, block chan bool, callOnce *sync.Once) (le.LeaderElectionConfig, error) {
-	podId, err := currentPodID(log)
+func (m *Manager) Stop() {
+	if m.cancelFunc != nil {
```
amirbenun, line 105: Add error log in case `cancelFunc` not set

leaderelection/leaderelection.go
```diff
@@ -141,55 +140,57 @@ func buildConfig(ctx context.Context, log *logp.Logger, client k8s.Interface, bl
...
-					go manager.leader.Run(ctx)
-					log.Infof("re-running leader elector")
-				})
+				// to keep following leader status if the context is not cancelled.
+				m.log.Infof("leader election lock LOST, id: %v", id)
+				defer m.wg.Done()
+
+				select {
+				case <-ctx.Done():
+					m.log.Info("Context is cancelled - should not re-run leader election")
```
amirbenun, line 153: lower case

deploy/kustomize/base/daemonset.yml
```diff
@@ -38,12 +38,10 @@ spec:
...
                   fieldPath: spec.nodeName
-            # This is done for leader election on local env
-            # GAP: https://github.com/elastic/cloudbeat/issues/162
             - name: POD_NAME
-              value: cloudbeat-leader
-              # fieldRef:
-              #   fieldPath: metadata.name
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
```
amirbenun, line 44: Do we use that?

leaderelection/leaderelection.go
```diff
@@ -141,55 +140,57 @@ func buildConfig(ctx context.Context, log *logp.Logger, client k8s.Interface, bl
...
 				// leaderelection.Run stops in case it has stopped holding the leader lease, we re-run the manager
-				// to keep following leader status.
-				retryOnce.Do(func() {
-					log.Infof("leader election lock LOST, id: %v", id)
-					go manager.leader.Run(ctx)
-					log.Infof("re-running leader elector")
-				})
+				// to keep following leader status if the context is not cancelled.
+				m.log.Infof("leader election lock LOST, id: %v", id)
+				defer m.wg.Done()
```
amirbenun, line 149: Add a comment that the function gets called even if cloudbeat wasn't the leader

amirbenun (commented):
Nice!

deploy/kustomize/base/daemonset.yml
uri-weisman: Yes, when running standalone cloudbeat.

leaderelection/leaderelection_test.go
```diff
@@ -23,29 +23,116 @@ import (
...
+	s.manager.Stop()
+
+	// Verify no goroutines are leaking. Safest to keep this on top of the function.
+	goleak.VerifyNone(s.T(), s.opts)
+}
+
+func (s *LeaderElectionTestSuite) TestManager_RunWaitForLeader() {
+	sTime := time.Now()
+	err := s.manager.Run(context.Background())
+	elapsed := time.Since(sTime)
```
amirbenun, line 84: `Run` function is running asynchronously and gets the leader in the background.
In other words, I don't think you really measure the amount of time we waited for a leader.

leaderelection/leaderelection_test.go
```diff
@@ -23,29 +23,116 @@ import (
...
+// Verify that when a pre-configured lease exists, eventually, the leader-manager will try to gain control if the
+// lease is not being renewed.
+func (s *LeaderElectionTestSuite) TestManager_RunWithExistingLease() {
+	podId := "this_pod"
+	os.Setenv(PodNameEnvar, podId)
+
+	holderIdentity := LeaderLeaseName + "_another_pod"
+	lease := generateLease(&holderIdentity)
+	s.manager.kubeClient = k8sFake.NewSimpleClientset(lease)
+	err := s.manager.Run(context.Background())
```
amirbenun, line 100: Again, `Run` returns immediately, how does it take the lease? I don't see where you waiting for it to change

leaderelection/leaderelection_test.go
```diff
@@ -23,29 +23,116 @@ import (
...
+		metav1.GetOptions{},
+	)
+
+	s.NoError(err)
+	s.Contains(*updatedLease.Spec.HolderIdentity, podId)
+}
+
+// Verify that after the lease is lost we re-run the leader-election manager.
+// After waiting for a FirstLeaderDeadline seconds we should be holding the lease again as it has not been renewed.
+func (s *LeaderElectionTestSuite) TestManager_ReRun() {
```
amirbenun, line 115: Nice!

amirbenun (commented):
Nice tests! Maybe we can add a unit test that run n managers in parallel and makes sure only on is the leader

leaderelection/leaderelection.go
amirbenun: Nitpicking, I would invert the `if` statement (so the more interesting part is not indented in)

DaveSys911 (approved):
approving @elastic/csp-ops `daemonset.yml`

amirbenun (approved):

PR #340, opened 2022-08-15
Bug fix - init kube watchers upon reconfiguration

uri-weisman:
The launcher interface is responsible for receiving config updates from Fleet.
When we get a new config, the launcher is stopping the existing cloudbeat and creates a new one with the new config.
The kube-fetcher is being re-created as part of this process, yet, the k8s watchers are not initiated because the sync object is global.
This PR solves this issue by creating a new sync object every time we re-create the kube-fetcher.


ofiriro3:
Nice change !

eyalkraft (approved):
Nice

amirbenun:
Nice catch @uri-weisman! Maybe worth adding a unit test where multiple fetchers initialize properly.

PR #346, opened 2022-08-17
feat: ecs event spec

olegsu:
Related to https://github.com/elastic/security-team/issues/3782
image

### DIFF
--- a/evaluator/opa.go
+++ b/evaluator/opa.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"time"
 
 	"github.com/elastic/cloudbeat/config"
 	"github.com/elastic/cloudbeat/resources/fetching"
@@ -34,6 +35,8 @@ import (
 	"github.com/open-policy-agent/opa/sdk"
 	"github.com/sirupsen/logrus"
 )
+
+var now = func() time.Time { return time.Now().UTC() }
 
 type OpaEvaluator struct {
 	log            *logp.Logger
@@ -115,7 +118,6 @@ func (o *OpaEvaluator) Eval(ctx context.Context, resourceInfo fetching.ResourceI
 		Result:         fetcherResult,
 		ActivatedRules: o.activatedRules,
 	})
-
 	if err != nil {
 		return EventData{}, fmt.Errorf("error running the policy: %v", err)
 	}
@@ -159,6 +161,7 @@ func (o *OpaEvaluator) decode(result interface{}) (RuleResult, error) {
 	}
 
 	err = decoder.Decode(result)
+	opaResult.Metadata.CreatedAt = now()
 	return opaResult, err
 }
 
--- a/evaluator/opa_test.go
+++ b/evaluator/opa_test.go
@@ -1,0 +1,70 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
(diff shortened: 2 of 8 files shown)


evaluator/opa_test.go
```diff
@@ -0,0 +1,70 @@
...
+package evaluator
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestOpaEvaluator_decode(t *testing.T) {
```
eyalkraft, line 27: Nice!

evaluator/opa.go
```diff
@@ -156,6 +159,7 @@ func (o *OpaEvaluator) decode(result interface{}) (RuleResult, error) {
 	}
 
 	err = decoder.Decode(result)
+	opaResult.Metadata.CreatedAt = now()
```
eyalkraft, line 162: Since it's only used once that could maybe be 
```suggestion
	opaResult.Metadata.CreatedAt = time.Now().UTC()
```

eyalkraft (changes requested):
Good job!
small changes required

evaluator/opa.go
olegsu: The reason for this was to have the ability to overwrite the value to a predictable one in the [test](https://github.com/elastic/cloudbeat/pull/346/files#diff-eb9ab193881d8642e921206e9c7c64f27a626defd0e3cc3a2f2b8e3773e46c0cR59)

resources/ecs/event.go
```diff
@@ -0,0 +1,93 @@
...
+	KindEvent         = "event"
+	KindMetric        = "metric"
+	KindState         = "state"
+	KindPipelineError = "pipeline_error"
+	KindSignal        = "signal"
+)
+
+// Event According to https://www.elastic.co/guide/en/ecs/current/ecs-event.html
+// event.ingested property is not part of this struct as the fleet server setting it
+type Event struct {
```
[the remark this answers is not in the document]
olegsu: Great, I was looking this one

eyalkraft (approved):
Looks good!

olegsu:
A quick update following https://github.com/elastic/cloudbeat/pull/346#discussion_r947981964

To support the `ecs` struct tags properly, we need to add to all the structs that we use `ecs` tag.  Later, implement the decoder that knows how to convert the struct into `mapstr.M` as required in the `beat.Event` (https://github.com/elastic/beats/blob/main/libbeat/beat/event.go#L44)
Similar to what is been done by [packetbeat](https://github.com/elastic/beats/blob/89bcc33a9a90cbacceaa61fceec0e8d23073574c/packetbeat/pb/event.go)

The migration to `ecs` tags can be done just after this will be merged as it mostly a refactor without any changes in terms of the structure of the outgoing event.

To close, the `Event` struct that was [removed](https://github.com/elastic/cloudbeat/pull/346#discussion_r947826187.) will be part of this PR

@eyalkraft @uri-weisman

eyalkraft:
Thanks @olegsu! This PR looks great.
Can you please open a follow-up ticket regarding using the `ecs` tag and implementing the decoder, including the information from your last comment  and including what is the motivation for us to do so? (pros vs cons)
Thanks!

gurevichdmitry (approved):
@olegsu, utils.py file approved. LGTM

olegsu:
> Thanks @olegsu! This PR looks great. Can you please open a follow-up ticket regarding using the `ecs` tag and implementing the decoder, including the information from your last comment and including what is the motivation for us to do so? (pros vs cons) Thanks!

Donehttps://github.com/orgs/elastic/projects/705/views/47?filterQuery=iteration%3A%22sprint+15%22

Issue #348, opened 2022-08-23
Fixing Kustomize CI 


ofiriro3:
**Motivation**
We would like to fix the broken unit tests regarding the new `Kustomize` feature that is now fully supported.

**In-Depth**
After emerging the EKS support using the `Kustomize` feature, some of the `strict` checks were broken.
Since it wasn't a mandatory check in the CI, the content was merged regardless of the failed tests.

**Definition of done**
- Fix the broken CI tests.
- Consider making these tests mandatory for the CI to pass.

**Related tasks/epics**
- https://github.com/elastic/cloudbeat/issues/332

oren-zohar:
@DaveSys911 any reason why we don't have these checks as required in the first place?

PR #351, opened 2022-08-23
fixing ci issues

ofiriro3:
- Fixes: https://github.com/elastic/cloudbeat/issues/348


.github/workflows/unit-test.yml
```diff
@@ -62,8 +62,9 @@ jobs:
           kubectl: 1.18.4
           kubeval: v0.16.1
           command: |
-            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/k8s/*.yml
-            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/k8s/kustomize/base/cloudbeat*.yml
+            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/k8s/*.yml --ignore-missing-schemas
```
oren-zohar, line 65:
```suggestion
            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/k8s/*.yml
```

.github/workflows/unit-test.yml
```diff
@@ -62,8 +62,9 @@ jobs:
...
+            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/eks/*.yml --ignore-missing-schemas
+            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/kustomize/base/*.yml --ignore-missing-schemas
```
oren-zohar, line 68: I would add the overlays too, wdyt?
```
kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/kustomize/overlays/cloudbeat-eks/*.yml --ignore-missing-schemas
kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/kustomize/overlays/cloudbeat-vanilla/*.yml --ignore-missing-schemas
kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/kustomize/overlays/test/*.yml --ignore-missing-schemas
```

oren-zohar (changes requested):

.github/workflows/unit-test.yml
ofiriro3: The reason that I put the `--ignore-missing-schemas` is that `kubeval` cannot evaluate a non-k8s schema object. 

Therefore, in a case of a type such that `eksctl.io/v1alpha5` the `kubeval` will throw an error since it cannot evaluate it.

I would not want to limit the content in the directory to k8s only. Then we wouldn't able to use `kind` .yaml in this directory.

For example -> `kind.x-k8s.io/v1alpha4`;

+ I see that it does not check recursively I am fixing it.

ofiriro3 (commented):
Commented + add recursive glob matching

.github/workflows/unit-test.yml
oren-zohar: specifically for what I commented on, there's no need for it

.github/workflows/unit-test.yml
oren-zohar: they will result in only warnings in our case, but will still prevent any breaking change (errors) no?

.github/workflows/unit-test.yml
```diff
@@ -62,8 +62,9 @@ jobs:
           kubectl: 1.18.4
           kubeval: v0.16.1
           command: |
-            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/k8s/*.yml
-            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/k8s/kustomize/base/cloudbeat*.yml
+            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/k8s/**/*.yml --ignore-missing-schemas
+            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/eks/*.yml --ignore-missing-schemas
+            kubeval --strict ${{ env.K8S_MANIFEST_DIR }}/kustomize/base/*.yml --ignore-missing-schemas
```
oren-zohar, line 67: maybe we can use `directories` flag instead of glob matching here
```suggestion
            kubeval --strict --directories ${{ env.K8S_MANIFEST_DIR }} --ignore-missing-schemas --ignored-filename-patterns kustomization
```

oren-zohar (changes requested):
we need to check all of the files, and since warnings wouldn't fail the check I don't mind having them. Working with the `directories` flag will ensure we will check all of the files inside `deploy` folder and will make it easier for maintenance (one we change the structure inside we don't need to update this check)

.github/workflows/unit-test.yml
ofiriro3: Yes, you are absolutely right, but I changed the code the first time you looked at it.
When you first review the code, it ran only on the parent directory(`deploy/k8s`) so no issues there, now it works recursively on all of the directories so without adding the flag it won't work.

.github/workflows/unit-test.yml
ofiriro3: not sure, I tried to remove essential fields for the deployment and still I get only warnings, it takes a really nasty change to make it notifies of an error.

In general, I think that since they are partial files, they should not be tested, you will always get a warning on these.
But I guess it is a matter of perspective, if you still think it is better doing it, I can do it.

.github/workflows/unit-test.yml
ofiriro3: That is a nice idea, but do we really want to test the entire directory?

Adv
- Maintenance - totally agree
- Can prevent future bugs

Dsv
- There are some files that I think should not be validated by the `kubeval` (the overlays files for example) - these files will always produce warnings, maybe they can even produce errors and still works. I just the solution is a bit too generalized.

That being said, I don't think it's so critical - if you think it's the best way after reading what I wrote - NP let me know and I'll change it.

oren-zohar: yes I would go with what I suggested since it will also prevent us from breaking the overlays. I don't mind the warning as I explained above.

oren-zohar (approved):

oren-zohar:
@ofiriro3 Please wait with the merge until they fix the 8.4 conflicts, also let's backport this into 8.4.

PR #360, opened 2022-08-26
Fix OPA logging structure and target

amirbenun:
Fixes https://github.com/elastic/security-team/issues/4168


amirbenun:
With the change:
image

oren-zohar:
@amirbenun the logs are still parsed incorrectly, is this PR could also handle it?

amirbenun:
Yes, right now the log message is still escaped but the default is to have no logs at all.
The logs are escaped since they are printed to the stdout instead of to the dedicated file.
In order to fix that we will have to implement OPA's `logging.Logger` interface in a way that delegates to the beats logger.
Do we want to have it right away or should I create a ticket? What do you think? @oren-zohar @yashtewari

oren-zohar:
> Yes, right now the log message is still escaped but the default is to have no logs at all.
> The logs are escaped since they are printed to the stdout instead of to the dedicated file.
> In order to fix that we will have to implement OPA's `logging.Logger` interface in a way that delegates to the beats logger.
> Do we want to have it right away or should I create a ticket? What do you think? @oren-zohar @yashtewari

@amirbenun If you think it's a quick fix we can do it here then we can update the current ticket, otherwise opening a separate ticket sounds good

_meta/config/cloudbeat.common.yml.tmpl
```diff
@@ -47,4 +47,6 @@ cloudbeat:
         patterns: [
           "/hostfs/etc/kubernetes/kubelet/kubelet-config.json",
           "/hostfs/var/lib/kubelet/kubeconfig",
-        ]
\ No newline at end of file
+        ]
+  evaluator:
```
DaveSys911, line 51: Please add it to cloudbeat.common.yml.tmpl & be sure to run `mage check`

DaveSys911 (approved):
See comment about adding change to cloudbeat reference tmpl

config/config_test.go
```diff
@@ -133,237 +133,6 @@ func (s *ConfigTestSuite) TestRuntimeCfgExists() {
 	}
 }
 
-func (s *ConfigTestSuite) TestConfigUpdate() {
```
oren-zohar, line 136: What is the reason for removing the tests here?

oren-zohar (changes requested):

config/config_test.go
amirbenun: The `Update` function is not being used anymore

oren-zohar (approved):

PR #272, opened 2022-07-13
Update deployment yamls

eyalkraft:
It seems like elastic-package was updated to support https and certificates:
- https://github.com/elastic/elastic-package/pull/789

If you use cloudbeat with elastic-package setup, these changes are required in order for things to work.

PR #367, opened 2022-08-30
feat: support latest elastic-packge stack deployment with ssl

olegsu:
# Motivation
Latest version of `elastic-package` setup for both Kibana and ES using SSL, HTTP disabled.
This is not compatible with the [vanilla](https://github.com/elastic/cloudbeat/tree/main/deploy/kustomize/overlays/cloudbeat-vanilla) deployment we have (`just deploy-cloudbeat`)

# Proposal
This PR follows what @eyalkraft [started](https://github.com/elastic/cloudbeat/pull/272/files). 
1. Update manifests with the HTTPS protocol on both Kibana and ES
2. Inject the certificate into the pod
3. When creating a `kind` cluster it will connect the cluster master node to the same network that created by elastic-package
4. Update to the latest version of elastic-package

## Flow
1. `just elastic-stack-up`
2. `just create-kind-cluster`
3. `just build-deploy-cloudbeat`


oren-zohar (approved):
LGTM 🚀

PR #369, opened 2022-08-30
Update pywin32 version

oren-zohar:
**In this PR**
- Update `docker` version in the `pyproject.toml` in order to support the newer `pywin32` version
- Update `poetry.lock`
- Force recreate: false to Allure report (it will edit the original instead of commenting a new one)
___
- Closes: https://github.com/elastic/security-team/issues/4072


.github/workflows/publish-test-results.yml
```diff
@@ -78,9 +78,9 @@ jobs:
         if: ${{ github.event.workflow_run.event == 'pull_request' }}
         with:
           header: Test Results
-          recreate: true
+          recreate: false
```
gurevichdmitry, line 81: @oren-zohar, according to [plugin's documentation](https://github.com/marocchino/sticky-pull-request-comment) **recreate** parameter is optional, and indicates if previous comment should be removed before creating a new comment. Only true is allowed. Just skip this option when you don't need it. Thus if not used can be removed.
But if I understand correctly, if option is **false**, each execution will create a new comment with the results link.

gurevichdmitry (changes requested):
@oren-zohar, please check comment.

gurevichdmitry (approved):

PR #370, opened 2022-08-31
Dependency review workflow

oren-zohar:
Adding [dependency review](https://github.com/actions/dependency-review-action) step, see it in action [here](https://github.com/oren-zohar/cloudbeat/actions/runs/2993517967)


DaveSys911 (approved):
LGTM

PR #374, opened 2022-09-05
Make tests more concurrent.

yashtewari:
Add the range option to tests so they can be split up and run in concurrent parts using a test matrix.

### DIFF
--- a/.github/workflows/cloudbeat-ci.yml
+++ b/.github/workflows/cloudbeat-ci.yml
@@ -92,14 +92,92 @@ jobs:
           mv /tmp/.buildx-cache-new /tmp/.buildx-cache
 
   Test_Matrix:
-    name: ${{ matrix.test-target }}-tests
+    name: ${{ matrix.test-target }}-${{ matrix.range }}-tests
     needs: Build
     runs-on: ubuntu-latest
     timeout-minutes: 55
     strategy:
       matrix:
-        # test-targets chosen by stabillity and runtime.Only pre_merge will be set as mandatory
-        test-target: [pre_merge, file_system_rules, k8s_object_rules, process_scheduler_rules, process_api_server_rules, process_controller_manager_rules, process_etcd_rules, process_kubelet_rules]
+        include:
+          - test-target: pre_merge
+          - test-target: file_system_rules
+            range: '0..15'
+          - test-target: file_system_rules
+            range: '15..30'
+          - test-target: file_system_rules
+            range: '30..45'
+          - test-target: file_system_rules
+            range: '45..60'
+          - test-target: file_system_rules
+            range: '60..'
+          - test-target: k8s_object_rules
+            range: '0..6'
+          - test-target: k8s_object_rules
+            range: '6..12'
+          - test-target: k8s_object_rules
+            range: '12..18'
+          - test-target: k8s_object_rules
+            range: '18..'
+          - test-target: process_scheduler_rules
+            range: '0..3'
+          - test-target: process_scheduler_rules
+            range: '3..'
+          - test-target: process_api_server_rules
+            range: '0..5'
+          - test-target: process_api_server_rules
+            range: '5..10'
+          - test-target: process_api_server_rules
+            range: '10..15'
+          - test-target: process_api_server_rules
+            range: '15..20'
+          - test-target: process_api_server_rules
+            range: '20..24'
+          - test-target: process_api_server_rules
+            range: '24..28'
+          - test-target: process_api_server_rules
+            range: '28..32'
+          - test-target: process_api_server_rules
+            range: '32..36'
+          - test-target: process_api_server_rules
+            range: '36..40'
+          - test-target: process_api_server_rules
+            range: '40..44'
+          - test-target: process_api_server_rules
+            range: '44..48'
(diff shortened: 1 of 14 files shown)


.github/workflows/cloudbeat-ci.yml
```diff
@@ -90,14 +90,92 @@ jobs:
...
     runs-on: ubuntu-latest
     timeout-minutes: 55
     strategy:
       matrix:
-        # test-targets chosen by stabillity and runtime.Only pre_merge will be set as mandatory
-        test-target: [pre_merge, file_system_rules, k8s_object_rules, process_scheduler_rules, process_api_server_rules, process_controller_manager_rules, process_etcd_rules, process_kubelet_rules]
+        include:
+          - test-target: pre_merge
+          - test-target: file_system_rules
+            range: '0..15'
```
oren-zohar, line 102: Instead of sending the range indices, I think we can send the number per group, or the number of groups, and do the splitting in `pytest_generate_tests`.

oren-zohar (changes requested):

.github/workflows/cloudbeat-ci.yml
yashtewari: In this model, one `pytest` invocation will only run one part of the test suite, so you would not only have to give the number of parts, but also the index of the part that should be run. Something like `parts: 4, part_index: 2`.

I prefer the range option over this. It's nice to have direct control on the range, in fact I've already been using the range option to run parts of tests for other tasks. The more fundamental implementation of range can also be used to build abstractions. For example, the kind of utility you're suggesting can be implemented as another layer on top of the range option.

oren-zohar: Yes, I understand what you are saying, but this approach will require more maintenance since if we wish to add tests we must edit the grouping again, as opposed to defining the number of groups in advance, and handling it in the splitting layer. It is also more error-prone, as it's hard to review this kind of change, where we will define the entire ranges like that. But again, I might be missing something here so I think we should go over it offline if possible.

DaveSys911: @yashtewari I agree with your point about adding an abstraction for the ranges.
Perhaps we can add something basic before this pr, to avoid hardcoding.

What if we use a script to generate the JSON payload for the matrix job?

That way we won't have to hardcode the ranges for each suite in the CI, Just decide on the convention or "interface" for the testing job. You can see something similar done with PowerShell [here]( https://purple.telstra.com/blog/github-action-test-matrices-part2).  Just need to agree on the parameters.

yashtewari: @DaveSys911 that's a pretty cool way to sidestep Github Actions limitations. Let's split the work into a new task.

> @oren-zohar: this approach will require more maintenance since if we wish to add tests we must edit the grouping again

Not exactly. New tests go to the end of the list, where they will automatically be included into the last partition of the suite. Further partitioning only needs to be done if the last part starts taking too much time. But then again, there is no plan to add any tests whatsoever to these suites, although I accept that this is a hassle long-term to do for every suite we implement.

> @oren-zohar: It is also more error-prone

I agree. For now we'll have to be more careful with reviews.

Again, my point isn't that we shouldn't add the ability to define more abstract options for this feature. All I'm saying is that we should start with the most basic building block, which is range, because that is the most useful for building anything more advanced. I'll mention again that it's so useful that I'm already using it to tackle other problems in the test infra.

This discussion IMO is about advancements to this feature, which should come later. The main issue right now is that the matrix definition is verbose. But we don't need to wait on a 5x improvement in CI speed before addressing that issue. It can be addressed later.

oren-zohar:
Ultimately, it's obvious we shouldn't implement parallelization at the workflow file level, but at the testing infrastructure level, but since it's a really significant improvement, we should consider using it anyway. In that case, I still believe we shouldn't hard-code the parallelization configuration. I especially like @DaveSys911's suggestion. I would suggest creating a script that will generate the indices split from some parameter, like this for example:
```bash
SPLIT=5 && echo $(seq 0 "$SPLIT" ) | tr ' ' ',' | xargs -J % echo '{"file_system_rules": [' % ']}' | jq 'to_entries|map(.key) as $keys | (map(.value)|transpose) as $values | $values|map([$keys, .] | transpose | map( {(.[0]): .[1]} ) | add)'
```
here `SPLIT` will set how many parts should we split the test suite, and we could set it at the workflow env params easly. The result will be as follows:
```json
[
  {
    "file_system_rules": 0
  },
  {
    "file_system_rules": 1
  },
  {
    "file_system_rules": 2
  },
  {
    "file_system_rules": 3
  },
  {
    "file_system_rules": 4
  },
  {
    "file_system_rules": 5
  }
]
```
@yashtewari @DaveSys911 what do you think?

oren-zohar (approved):
@yashtewari will open a follow up task so we can prioritize and track

yashtewari:
@DaveSys911 the old required tests are blocking this PR. Could you force merge + set the new required tests when the checks pass? Thanks!

DaveSys911:
> @yashtewari will open a follow up task so we can prioritize and track

@oren-zohar I think it makes sense to invest in some design before we jump into the next task.

PR #376, opened 2022-09-06
fix: connect kind cluster to elastic-package network

olegsu:
Separate `create-kind-cluster` to `elastic-stack-connect-kind`.
This fix the dependency `cleanup-create-local-helm-cluster` has on `create-kind-cluster` but it should not be connected to the stack network.

Thanks @yashtewari


yashtewari (approved):
Nice 🙂

oren-zohar (approved):
LGTM

PR #379, opened 2022-09-08
chore: add test

olegsu:
Related to https://github.com/elastic/security-team/issues/4495

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/elastic/beats/v7/x-pack/libbeat/common/aws"
 	"github.com/elastic/elastic-agent-libs/config"
 	"github.com/elastic/elastic-agent-libs/logp"
 	"github.com/stretchr/testify/suite"
@@ -46,17 +47,15 @@ func (s *ConfigTestSuite) SetupTest() {
 }
 
 func (s *ConfigTestSuite) TestNew() {
-	var tests = []struct {
+	tests := []struct {
 		config                 string
 		expectedActivatedRules []string
 		expectedType           string
-		expectedAccessKey      string
-		expectedSecret         string
-		expectedSessionToken   string
+		expectedAWSConfig      aws.ConfigAWS
 	}{
 		{
 			`
-   type : cloudbeat/cis_k8s
+   type : cloudbeat/cis_eks
    streams:
     - runtime_cfg:
         activated_rules:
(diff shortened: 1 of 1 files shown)


config/config_test.go
```diff
@@ -261,3 +265,42 @@ streams:
 		s.Equal(test.expectedEksActivatedRules, c.Streams[0].RuntimeCfg.ActivatedRules.CisEks)
 	}
 }
+
+func (s *ConfigTestSuite) TestAWS() {
```
ofiriro3, line 269: Why not edit `TestNew`?

The convention that we usually use is the `TestFunctionName` - like the following suggests https://ieftimov.com/posts/testing-in-go-naming-conventions/#function-naming-conventions

If you found it better to have its own function I don't mind, but maybe edit the test name to

`TestNewConfigFrom` `TestNewConfigFromAwsConfig` or something similar

config/config_test.go
```diff
@@ -261,3 +265,42 @@ streams:
...
+	tests := []struct {
+		config   string
+		expected aws.ConfigAWS
+	}{
+		{
+			`
+type: cloudbeat/cis_k8s
```
ofiriro3, line 276: This `type` will compile obviously, but it's a bad example for someone who is trying to understand the `config.go` by going over its tests.

Please change it to `type: cloudbeat/cis_eks`

ofiriro3 (changes requested):

ofiriro3:
Good PR!

Thanks for adding the relevant tests :) 

As far as I know, we usually link the Github to the git issue this way.

- Fixes/Implements <Github issue link>

config/config_test.go
olegsu: Sounds right

config/config_test.go
olegsu: Oh, my mistake

olegsu: @ofiriro3 this is not that simple actually
`type: cloudbeat/cis_eks` tries to connect to aws [here](https://github.com/elastic/cloudbeat/blob/main/resources/providers/awslib/config_provider.go). It returns with an error.
To support this test properly we need to inject that so it can be mocked. 
Do you think this refactor should be done as part of this pr?

* The test updated.

ofiriro3: Yes, I think it should be part of this PR.

https://github.com/elastic/cloudbeat/blob/main/resources/providers/awslib/config_provider_test.go
Everything is mocked here already, it should be super easy.

ofiriro3 (approved):

PR #380, opened 2022-09-08
Load OPA bundle from FS

uri-weisman:
1. Release - package opa bundle with cloudbeat.
3. Load opa bundle from FS and remove bundle-server logic.
2. Update CI.

- https://github.com/elastic/security-team/issues/4496


DaveSys911:
@uri-weisman looks neat.I like the trick with build step.
I think the only thing missing is adding the bundle build to the mage scripts we use to produce the cloudbeat.tar.gz for release. you can find them [here](https://github.com/elastic/cloudbeat/blob/main/scripts/mage/package.go) & the[ main magefile](https://github.com/elastic/cloudbeat/blob/main/magefile)

.github/workflows/cloudbeat-ci.yml
```diff
@@ -34,6 +34,14 @@ jobs:
         with:
           go-version: ${{ env.GO_VERSION }}
 
+      - name: Fetch OPA version
```
DaveSys911, line 37: LGTM, if u can please add these steps to  the  nightly ci job as well
https://github.com/elastic/cloudbeat/blob/main/.github/workflows/Periodic-CI.yml

magefile.go
```diff
@@ -334,3 +337,13 @@ func PythonEnv() error {
 	_, err := mage.PythonVirtualenv(true)
 	return err
 }
+
+func BuildOpaBundle() error {
```
DaveSys911, line 341: Perhaps can be a good idea to add this function as a dependency to the build step
https://github.com/elastic/cloudbeat/blob/main/magefile.go?plain=1#L81. since without the rego bundle cb isn't really useable.

scripts/mage/package.go
```diff
@@ -21,13 +21,24 @@ import (
 	devtools "github.com/elastic/beats/v7/dev-tools/mage"
 )
 
+const opaBundle = "bundle.tar.gz"
+
 // CustomizePackaging modifies the device in the configuration files based on
 // the target OS.
 func CustomizePackaging() {
+	bundleDir := devtools.PackageFile{
```
DaveSys911, line 29: Could you perhaps add the correct opa bundle to our hermit config as well?
https://cashapp.github.io/hermit/usage/management/

DaveSys911 (approved):
Looks great. Have a few small comments & suggestions. Adding the changes to periodic ci is critical, the rest up to you.

evaluator/opa.go
```diff
@@ -67,15 +61,15 @@ var opaConfig = `{
 }`
 
 func NewOpaEvaluator(ctx context.Context, log *logp.Logger, cfg config.Config) (Evaluator, error) {
-	server, err := StartServer(ctx, cfg)
+	// provide the OPA configuration which specifies
+	// fetching policy bundle and logging decisions locally to the console
+	path, err := filepath.Abs("bundle.tar.gz")
```
DaveSys911, line 66: Looks good. If the bundle isn't present on the path, an error will register, right?

magefile.go
eyalkraft, line 341: I'd rather not have two different ways to build a bundle. (magefile and justfile).
This can cause problems if in the future someone only change one of them and they diverge.
What about consolidating the bundle building logic to a single place and then using it from wherever needed?

justfile
```diff
@@ -23,7 +24,12 @@ build-deploy-cloudbeat-debug: build-cloudbeat-debug load-cloudbeat-image deploy-
 load-cloudbeat-image:
   kind load docker-image cloudbeat:latest --name kind-mono
 
+build-opa-bundle:
+  CSP_POLICIES_PKG_DIR=$( go list -m -f {{{{.Dir}} {{cspPoliciesPkg}} ) && \
+  opa build -b ${CSP_POLICIES_PKG_DIR}/bundle -e ${CSP_POLICIES_PKG_DIR}/bundle/compliance
```
eyalkraft, line 29: Does that assume I have the opa binary locally?

.github/workflows/cloudbeat-ci.yml
eyalkraft: @DaveSys911 so we have to duplicate?
Isn't there a way to have these steps once and consume them from both jobs?

DaveSys911: @eyalkraft we have a task for it, but currently it does.

magefile.go
DaveSys911: Good catch, Our build scripts are a bit messy, working on consolidation. As a temporary step invoking mage command from justfile is a good solution.

justfile
DaveSys911: yeap, we added hermit to the repo. So opa can be added to it.
https://github.com/elastic/cloudbeat/pull/380#discussion_r971697163

scripts/mage/package.go
uri-weisman: Weird, they don't have the version we use - **opa-0.42.0**


`hermit install opa-0.42.0`
```
fatal:hermit: opa-0.42.0: https://github.com/cashapp/hermit-packages.git/opa.hcl: no version opa-0.42.0 found in versions (opa-0.30.1, opa-0.38.1, opa-0.43.0, opa-0.44.0) or channels (opa@0, opa@0.30, opa@0.38, opa@0.43, opa@0.44, opa@latest): unknown package
```

evaluator/opa.go
uri-weisman: That is correct.

PR #390, opened 2022-09-13
Delegate opa logger to logp lib

amirbenun:
Resolves https://github.com/elastic/security-team/issues/4168


evaluator/logger.go
```diff
@@ -0,0 +1,95 @@
...
+func (l *logger) GetLevel() logging.Level {
+	return toOpaLevel(l.lvl.Level())
+}
+
+func (l *logger) SetLevel(level logging.Level) {
+	l.lvl.SetLevel(toZapLevel(level))
+}
+
+func mapToArray(m map[string]interface{}) []interface{} {
+	ret := []interface{}{}
```
oren-zohar, line 60:
```suggestion
	var ret []interface{}
```
as a rule of thumb, we should try to use nil slice declaration wherever we can so an empty result will be comparable with nill, see an [exmaple](https://go.dev/play/p/V8pesYrwZBY)

evaluator/logger.go
```diff
@@ -0,0 +1,95 @@
...
+
+	return zap.FatalLevel
+}
+
+func toOpaLevel(l zapcore.Level) logging.Level {
+	if res, ok := zapToOpaLevelsMap[l]; ok {
+		return res
+	}
+
+	return logging.Level(4)
```
oren-zohar, line 94: what does 4 stand for? level.Debug is 3

oren-zohar (changes requested):
Looks good, need to generate license headers with `mage addLicenseHeaders`. Did you test it? can you please share how the logs look now?

amirbenun:
Fleet UI logs screenshot:
image

Note that "Bundle load failed..." is produced by OPA.

oren-zohar (approved):
🚀 🚀 🚀 🚀

PR #393, opened 2022-09-15
Fix `process_api_server` test cases.

yashtewari:
These cases used to break the cluster, mostly due to bad configuration values.


tests/product/tests/data/process/process_test_cases.py
```diff
@@ -855,16 +856,6 @@
 )]
 
 cis_1_2_23 = [(
```
oren-zohar, line 858: we still need a fail case for this rule

tests/product/tests/data/process/process_test_cases.py
```diff
@@ -1344,8 +1343,10 @@
     *cis_1_2_29,
     *skip_param_case(skip_list=[*cis_1_2_32],
                      data_to_report=SkipReportData(
-                         skip_reason="This case fails and breaks cluster")
-                     )
+                         url_title="security-team: #4981",
+                         url_link="https://github.com/elastic/security-team/issues/4981",
+                         skip_reason="Depends on this issue to be closed before fixing"
```
oren-zohar, line 1348: the issue is closed, plus per your [comment](https://github.com/elastic/security-team/issues/4193#issuecomment-1281924261):

> CIS 1.2.32: K8S won't accept dummy values for --tls-cipher-suites; valid cipher suites have to be provided.

So we is it on skip?

oren-zohar (changes requested):

oren-zohar (changes requested):
Looks good, let's just move the failing tests to skip and track them, then we can merge

PR #395, opened 2022-09-15
Fixed empty default range in justfile

kfirpeled:


yashtewari (commented):
Thanks Kfir. I'd prefer if instead of changing the defaults in the `JUSTFILE`, we supply the new correct arguments from where the particular jobs are being "called". So for example for the nightly jobs, let's [supply the argument on the job](https://github.com/elastic/cloudbeat/blob/92a10ff46b7adc12ae4d98e7dfba132ad181c785/.github/workflows/Periodic-CI.yml#L183), and not on the default.

The defaults are set to `range=''` in places that affect tests for which range has not been implemented, so those tests might break if you change that default. See how the range option [is set conditionally](https://github.com/elastic/cloudbeat/blob/155c2f34cc477f242d7f343367732e4863b1c7a8/tests/deploy/k8s-cloudbeat-tests/templates/cloudbeat-test-configmap.yml#L14) to prevent some tests from breaking.

Also, `..` is equivalent to `0..`. [In this snippet](https://github.com/elastic/cloudbeat/blob/155c2f34cc477f242d7f343367732e4863b1c7a8/tests/product/tests/conftest.py#L145), with `range=..`, the flow of control ends up doing nothing to the original list of test cases, which means all test cases are run.

DaveSys911:
@yashtewari His fix seems to play nice with test markers that don't require the range value, like "pre-merge" https://github.com/elastic/cloudbeat/actions/runs/3062791363/jobs/4944228080. I suggest we merge this as it fixes nightly, and iterate over it.

kfirpeled:
@yashtewari if you have another solution please go for it :)
Currently, our nightly is running only 6 tests and we are losing our visibility

- [nightly allure report](http://csp-allure-reports.s3.amazonaws.com/allure_reports/cloudbeat/main/index.html#)

yashtewari:
When we want to revive the nightly, we'll do it in the way described [in this comment](https://github.com/elastic/cloudbeat/pull/395#pullrequestreview-1110123843).

PR #398, opened 2022-09-18
Use GitHub token in setup-just action

oren-zohar:
updates extractions/setup-just@v1 action to use `GITHUB_TOKEN` in order to avoid [rate-limiting errors](https://github.com/elastic/cloudbeat/actions/runs/3078654006/jobs/4974428235#step:4:6) like here:
image

ref:
- https://github.com/extractions/setup-just#usage

### DIFF
--- a/.github/workflows/Periodic-CI.yml
+++ b/.github/workflows/Periodic-CI.yml
@@ -172,6 +172,8 @@ jobs:
         id: install
 
       - uses: extractions/setup-just@v1
+        env:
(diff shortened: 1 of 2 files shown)


DaveSys911 (changes requested):
The rate limiting issue isn't exclusive to the just action. It's per origin IP.
Try adding it as a global env var to both workflows

DaveSys911 (approved):
Edit: noticed it's a common issue with this action - approved

PR #399, opened 2022-09-19
fix depedency scanner on push event

DaveSys911:
Limits dependency reviews to pr only
https://github.com/elastic/cloudbeat/actions/runs/3050543387/jobs/4917744157


oren-zohar (approved):
LGTM

PR #406, opened 2022-09-20
Update csp-security-policies to v1.2.0

yashtewari:


jeniawhite (approved):

oren-zohar:
> Was there a reason to take it instead of 1.1 with the older OPA version?

Yes @jeniawhite, they a security issue in the previous OPA vesion

PR #422, opened 2022-09-29
feat: run pre-merge in multi node

olegsu:
**What does this PR do?**

* Update all the ci tests steps to run in a multi-node (2 nodes) cluster.
* Update all the required just commands to support `kind` as a parameter with `default=kind-multi`.

**Run Locally**
This is TL;DR of [Readme](https://github.com/elastic/cloudbeat/blob/1d8c64ad8b47e4a00ab7d576a963d19f537e363e/tests/README.md)
```bash
just create-kind-cluster
just build-cloudbeat
just load-cloudbeat-image
just deploy-tests-helm-ci pre_merge
just build-pytest-docker
just load-pytest-kind
just run-tests
```

**Related Issues**
Related: https://github.com/elastic/security-team/issues/4741


olegsu:
Although it passed the `pre_merge` is [flaky](https://github.com/elastic/cloudbeat/actions/runs/3151932372/jobs/5127145039).
image

tests/integration/tests/conftest.py
```diff
@@ -40,6 +41,7 @@ def fixture_start_stop_cloudbeat(k8s, api_client, cloudbeat_agent):
                           name=cloudbeat_agent.name,
                           status_list=['ADDED', 'MODIFIED'],
                           namespace=cloudbeat_agent.namespace)
+    time.sleep(5)
```
olegsu, line 44: Added to resolve flakiness of `k8s.wait_for_resource`
@gurevichdmitry thanks

yashtewari (approved):
Neat. Please make sure that the local run commands like
```
just run-test-targets
```
are still working with this change.

tests/integration/tests/conftest.py
olegsu: The flakiness might be related to the fact that now the cluster has 2 pods and not just one.
The `wait_for_resource` function returns after the first find.
I think we need something similar to `kubectl rollout status ds NAME` (https://github.com/kubernetes/kubectl/blob/master/pkg/polymorphichelpers/rollout_status.go#L95)

olegsu:
> Neat. Please make sure that the local run commands like
> 
> ```
> just run-test-targets
> ```
> 
> are still working with this change.

Hm, this command fails for me also on `main` branch

.github/workflows/cloudbeat-ci.yml
```diff
@@ -114,104 +114,154 @@ jobs:
       matrix:
         include:
           - test-target: pre_merge
```
DaveSys911, line 116: If there is some easy workaround to avoid hardcoding kind-mono as the default on multiple lines, would be a nice plus.

DaveSys911 (approved):
LGTM

tests/integration/tests/conftest.py
oren-zohar: please add an explanation in the code for why we need this wait, so we'll know how to handle it in the future if we want to remove it. Alternatively, we can use the Kubernetes SDK to list all the pods with elastic-agent in their names and wait for the number to be equal to the number of nodes we have, but this is a bit more complicated than this.

oren-zohar (changes requested):
Looks great, I'm just wondering if we're missing the point with having all of the functional tests running on mono-cluster

.github/workflows/cloudbeat-ci.yml
```diff
@@ -114,104 +114,154 @@ jobs:
...
+            kind: kind-multi
           - test-target: file_system_rules
             range: '0..5'
+            kind: kind-mono
```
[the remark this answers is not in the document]
olegsu: This is the "default" cluster. I just renamed if from `kind-config`.
Previously we have two kind
1. kind-config
2. kind-multi

justfile
```diff
@@ -4,8 +4,8 @@ kustomizeVanillaOverlay := "deploy/kustomize/overlays/cloudbeat-vanilla"
 kustomizeEksOverlay := "deploy/kustomize/overlays/cloudbeat-eks"
 cspPoliciesPkg := "github.com/elastic/csp-security-policies"
 
-create-kind-cluster:
-  kind create cluster --config deploy/k8s/kind/kind-config.yml --wait 30s
+create-kind-cluster kind='kind-mono':
```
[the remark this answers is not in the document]
olegsu: It feels like a breaking change, we used to run this command and to have just one node, wdyt?

.github/workflows/cloudbeat-ci.yml
olegsu: It can be done, as in [this](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations) example.
So we need to add more steps to create an environment variable if the `${{ matrix.kind }}` is not set. I like it more explicitly. Wdyt?

justfile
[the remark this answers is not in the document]
oren-zohar: will it break the tests? if not, why not then? this is the more common use case for our product, if we can test it in a multi-node cluster then we should. wdyt?

olegsu:
All the other tests should be moved to the multi node cluster as a follow-up task

oren-zohar (approved):
:shipit:

oren-zohar (approved):
LGTM

Issue #416, opened 2022-09-22
OPA decision logger attempting to send logs to bundle server


jeniawhite:
**Describe the bug**
I've investigated the logs and saw that when we turn off the decision logger (this is how we run by default), OPA's behavior in that case is to attempt and send the decision logs to the bundle server that is configured by us.
This is the configuration that we have right now (%t by default is false):
```
var opaConfig = `{
	"bundles": {
		"CSP": {
			"resource": "file://%s"
		}
	},
	"decision_logs": {
		"console": %t
	}
}`
```
You can see this in the documentation of OPA (https://www.openpolicyagent.org/docs/v0.13.5/configuration/#decision-logs).
It states the following: **"Name of the service to use to contact remote server. If no plugin is specified, and console logging is disabled, this will default to the first service name defined in the Services configuration"**.

If you will run cloudbeat with OPA you will be able to see that it attempts to send the logs to an unresolvable endpoint and having problems with it:
```
{"log.level":"info","@timestamp":"2022-09-22T10:18:11.574Z","log.origin":{"file.name":"process/stdlogger.go","file.line":54},"message":"cloudbeat stdout: \"{\\\"@timestamp\\\":\\\"2022-09-22T10:18:11Z\\\",\\\"log.level\\\":\\\"error\\\",\\\"log.logger\\\":\\\"opa_logger\\\",\\\"message\\\":\\\"Log encoding failed: upload chunk size too small.\\\",\\\"plugin\\\":\\\"decision_logs\\\",\\\"service.name\\\":\\\"cloudbeat\\\"}\\n\"","agent.console.name":"cloudbeat","agent.console.type":"stdout","ecs.version":"1.6.0"}
{"log.level":"info","@timestamp":"2022-09-22T10:18:14.189Z","log.origin":{"file.name":"process/stdlogger.go","file.line":54},"message":"cloudbeat stdout: \"{\\\"@timestamp\\\":\\\"2022-09-22T10:18:14Z\\\",\\\"log.level\\\":\\\"error\\\",\\\"log.logger\\\":\\\"opa_logger\\\",\\\"message\\\":\\\"log upload failed, server replied with not found.\\\",\\\"plugin\\\":\\\"decision_logs\\\",\\\"service.name\\\":\\\"cloudbeat\\\"}\\n\"","agent.console.name":"cloudbeat","agent.console.type":"stdout","ecs.version":"1.6.0"}
```

**To Reproduce**
[...]

PR #423, opened 2022-09-29
Opa debug logger

amirbenun:
Fixes https://github.com/elastic/cloudbeat/issues/416


evaluator/debug_logger/factory.go
```diff
@@ -0,0 +1,43 @@
...
+	return &plugin{
+		manager: m,
+		mtx:     sync.Mutex{},
+		config:  conf.(config),
+	}
+}
+
+func (Factory) Validate(_ *plugins.Manager, conf []byte) (interface{}, error) {
+	parsedConfig := config{}
+	return parsedConfig, util.Unmarshal(conf, &parsedConfig)
```
yashtewari, line 42: Nice

evaluator/debug_logger/plugin.go
```diff
@@ -0,0 +1,77 @@
...
+func (p *plugin) Start(ctx context.Context) error {
+	p.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateOK})
+	return nil
+}
+
+func (p *plugin) Stop(ctx context.Context) {
+	p.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateNotReady})
+}
+
+func (p *plugin) Reconfigure(ctx context.Context, conf interface{}) {
```
yashtewari, line 50: This is a no-op function right? Would be better for it to be explicitly no-op (empty function).

evaluator/debug_logger/plugin.go
```diff
@@ -0,0 +1,77 @@
...
+		p.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateErr})
+		return nil
+	}
+	fields := map[string]interface{}{}
+	err = util.UnmarshalJSON(eventBuf, &fields)
+	if err != nil {
+		p.manager.UpdatePluginStatus(PluginName, &plugins.Status{State: plugins.StateErr})
+		return nil
+	}
+	p.manager.ConsoleLogger().WithFields(fields).WithFields(map[string]interface{}{
```
yashtewari, line 73: I think it should be possible to inherit from the libbeat logger that the rest of cloudbeat uses, and use that instead. That would truly align logs from OPA with all the rest of the logs. What do you think?

evaluator/opa.go
```diff
@@ -51,11 +53,17 @@ var opaConfig = `{
 			"resource": "file://%s"
 		}
 	},
-	"decision_logs": {
-		"console": %t
-	}
+%s
```
yashtewari, line 56: This is getting a bit hard to read now, and is a good use case for the [templates package](https://pkg.go.dev/text/template).

evaluator/logger.go
```diff
@@ -83,7 +83,7 @@ func mapToArray(m map[string]interface{}) []interface{} {
 }
 
 func newLogger() logging.Logger {
-	lvl := zap.NewAtomicLevelAt(zapcore.InfoLevel)
+	lvl := zap.NewAtomicLevelAt(zapcore.DebugLevel)
```
yashtewari, line 86: I guess I'm out of sync here. What's the reason we decided to setup this logger instead of inheriting from the cloudbeat/libbeat logger?

yashtewari (commented):
Looks like the right direction for dealing with OPA events.

amirbenun:
Fleet UI doesn't show the entire log but only selected fields, with the change it looks like that:
image

We should define which of the hidden fields in the `EventV1` struct should be displayed:
```
type EventV1 struct {
	Labels       map[string]string       `json:"labels"`
	DecisionID   string                  `json:"decision_id"`
	Revision     string                  `json:"revision,omitempty"` // Deprecated: Use Bundles instead
	Bundles      map[string]BundleInfoV1 `json:"bundles,omitempty"`
	Path         string                  `json:"path,omitempty"`
	Query        string                  `json:"query,omitempty"`
	Input        *interface{}            `json:"input,omitempty"`
	Result       *interface{}            `json:"result,omitempty"`
	MappedResult *interface{}            `json:"mapped_result,omitempty"`
	Erased       []string                `json:"erased,omitempty"`
	Masked       []string                `json:"masked,omitempty"`
	Error        error                   `json:"error,omitempty"`
	RequestedBy  string                  `json:"requested_by,omitempty"`
	Timestamp    time.Time               `json:"timestamp"`
	Metrics      map[string]interface{}  `json:"metrics,omitempty"`

	inputAST ast.Value
}
```

oren-zohar:
@amirbenun make sure to check the allure report before merging

.github/workflows/unit-test.yml
```diff
@@ -36,9 +36,49 @@ jobs:
...
+
+      - name: Download coverage artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: coverage-file
+
+      - name: Send coverage
+        env:
+          COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: goveralls -coverprofile=covprofile -service=github
```
oren-zohar, line 81: is it covprofile or coverage-file?

oren-zohar (changes requested):

.github/workflows/unit-test.yml
amirbenun: It's the coverage result file it should read from

oren-zohar (changes requested):
Please add tests

amirbenun:
@oren-zohar noted that after the change our coverage dropped by 6%.
In this PR I also reconfigured goveralls tool, instead of automatically creating a coverage file, it takes an existing coverage file that we generate (in order to avoid running our unit tests twice).
It seems like there are differences in the different coverage calculations, the main reason for that is that with the current implementation mock files coverage is 0% whereas before we had some coverage on mock files.
The solution here is to exclude the mock files from the coverage report by adding the `_test` suffix.
Do you want me to create a task for that? @oren-zohar

oren-zohar (approved):
🚀 🚀 🚀

PR #441, opened 2022-10-03
feat: add cluster name to elb resource

olegsu:
**What does this PR do?**
Update `resource.id` for all AWS types to have a cluster name as part of the id
- `iam`
- `caas`
- `load-balancer`
- `container-registry`

**Related Issues**
- Related https://github.com/elastic/security-team/issues/4770

**Checklist**
- [ ] I have added necessary README/documentation (if appropriate)

### DIFF
--- a/transformer/common_data_provider.go
+++ b/transformer/common_data_provider.go
@@ -114,7 +114,9 @@ func (c CommonDataProvider) getNodeName() (string, error) {
 func (cd CommonData) GetResourceId(metadata fetching.ResourceMetadata) string {
 	switch metadata.Type {
 	case fetchers.ProcessResourceType, fetchers.FSResourceType:
-		return cd.generateUUID(metadata)
+		return uuid.NewV5(uuid_namespace, cd.clusterId+cd.nodeId+metadata.ID).String()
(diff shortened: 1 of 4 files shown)


transformer/common_data_provider.go
```diff
@@ -114,7 +114,9 @@ func (c CommonDataProvider) getNodeName() (string, error) {
 func (cd CommonData) GetResourceId(metadata fetching.ResourceMetadata) string {
 	switch metadata.Type {
 	case fetchers.ProcessResourceType, fetchers.FSResourceType:
-		return cd.generateUUID(metadata)
+		return uuid.NewV5(uuid_namespace, cd.clusterId+cd.nodeId+metadata.ID).String()
+	case fetching.ElbType, fetching.EcrType, fetching.EksType, fetching.RolePolicyType:
```
amirbenun, line 118: After having a glance at the corresponding fetchers, it seems like all of these values are used as `metadata.Subtype` which means you will never get into that case as your switch checks `metadata.Type`.

Instead, you should use:
`fetching.CloudLoadBalancer`
`fetching.CloudContainerRegistry`
`fetching.CloudContainerMgmt`
`fetching.CloudIdentity`

amirbenun (changes requested):

amirbenun (approved):

PR #449, opened 2022-10-06
Optimize test stack resources & cloudbeat cycle duration

DaveSys911:
**What does this PR do?**

- Increase cloudbeat's cycle duration to 30s
- Add cloudbeat resource limits
- Increase ES resource limit
**Related Issues**

**Checklist**
- [x] I have added necessary README/documentation (if appropriate)

### DIFF
--- a/tests/deploy/cloudbeat-pytest.yml
+++ b/tests/deploy/cloudbeat-pytest.yml
@@ -145,7 +145,7 @@ data:
       type: cloudbeat/cis_k8s
       streams:
           # Defines how often an event is sent to the output
-        - period: 5s
+        - period: 30s
           fetchers:
             - name: kube-api
             - name: process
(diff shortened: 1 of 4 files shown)


tests/deploy/cloudbeat-pytest.yml
```diff
@@ -296,6 +296,9 @@ spec:
           requests:
             cpu: 100m
             memory: 100Mi
+          limits:
+            cpu: 1000m
+            memory: 1Gi
```
gurevichdmitry, line 301: @DaveSys911 nit, we can fail to catch real performance problem defining memory limit to 1Gi.
I think we need to measure cpu and memory and define more real limit value for Cloudbeat.

gurevichdmitry (approved):
@DaveSys911 LGTM

tests/deploy/cloudbeat-pytest.yml
DaveSys911: I agree, this one is a very generous upper limit to avoid killing CI workers. We should decrease it.

Issue #444, opened 2022-10-03
[BUG] EKS integration - cloudbeat does not crash when updating with bad credentials


kfirpeled:
**Describe the bug**

When providing bad credentials to EKS, cloudbeat does not crash as expected.
Instead, it logs the following:

`
[elastic_agent.cloudbeat][error] Failed to update Beater config: could not create beater: could not retrieve user identity for ECR fetcher: operation error STS: GetCallerIdentity, https response error StatusCode: 403, RequestID: d132180b-670e-4895-961e-0d8fba171f69, api error SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.
`

**Preconditions**
Running 8.5.0-latest and eks deployment (containing two nodes) ready for use

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Set up kspm-eks integration in kibana with proper aws credentials (ACCESS_KEY+SECRET)
2. Deploy the elastic-agent
3. Wait for first results
4. Add '@' character in end of the SECRET
5. Save and deploy
6. Check the logs/ verify if cloudbeat crashes

**Expected behavior**
Cloudbeat should crash

**Additional context**
 - Kibana Version: 8.5.0 BC2 dc769f45a5a6dafb0a8c8f0c0cabcced4df45e11
 - elastic-agent version: BC2 elastic-agent:8.5.0-ca5dd693
 - https://github.com/elastic/cloudbeat/issues/388

oren-zohar:
need to verify three cases:
- [ ] no credentials are provided
- [ ] invalid credentials are provided
- [ ] credentials with not enough privileges are provided 

all should match our docs.

@olegsu can you please take it as part of the QA Cycle?

olegsu:
## No credentials
When no credentials are provided cloudbeat will use the ec2 attached IAM role and try using it when calling AWS APIs.
This means that it will continue reporting events based on the permissions it has (files, processes, k8s)

<img width="2546" alt="image" src="https://user-images.githubusercontent.com/23185631/195087644-dba06196-3b72-4f9d-a60c-a7643cb706d5.png">


## Update VS New Agent
The cloud beat crash depends on whether the configuration was invalid ( or wrong ) at the beginning or streamed from Kibana.
[...]

PR #457, opened 2022-10-18
feat: crash cloudbeat when update configuration is failed

olegsu:
**Motivation**
When the configuration is invalid the expected behavior is to crash (return an error to the beats framework) the cloudbeat process.
This works as expected when the cloudbeat starts. But, it does not crash when the configuration comes from Kibana as an update.
When cloudbeat receives an invalid update, it does not crash, nor keep working with the previous configuration. 
Instead, it enters into an idle state.

**Solution**
When a configuration update comes, cloudbeat will "restart". The restart process will try to connect to AWS and if fails it will return an error.
Previously [this](https://github.com/elastic/cloudbeat/pull/457/files#diff-88c2b01747f09005808187f6172586ba6793d22f97f2725c3e24f298069225b7R177) error was just logged.
With this change, the error will be reported to the error channel and [handled](https://github.com/elastic/cloudbeat/pull/457/files#diff-88c2b01747f09005808187f6172586ba6793d22f97f2725c3e24f298069225b7R177) to exit.

Closes https://github.com/elastic/cloudbeat/issues/444


olegsu:
**What is happening when an invalid configuration is streamed from Kibana?**
When Cloudbeat starts, it starts the `Launcher`, a process that is managing the underlying `Beater`.
The `Launcher` is also watching for new configurations that might be streamed so that it can stop the current Beater and start a new one.
When a new configuration is saved on Kibana, it is streamed to Cloudbeat, which might return an error when the configuration is invalid (wrong credentials).
This error is not handled, just logged. So Cloudbeat might be in an invalid state. 

**Handle the reconfiguration error**
When the launcher gets a new configuration and the new Beater returns an error, the launcher will return an error to his caller. So that the whole process can be restated.

When `launcher.Run(b *beat.Beat)` [returns](https://github.com/elastic/cloudbeat/blob/main/launcher/launcher.go#L83), it tries to close the connection with the fleet management service. 
During this period the Elastic Agent tries to get the Cloudbeat status, but a process fails since there is no running [Beater](https://github.com/elastic/cloudbeat/blob/main/beater/cloudbeat.go#L50). So Cloudbeat will be marked with "Failed" status with the message "Missed two check-ins".

Elastic Agent should restart the Cloudbeat at this point.
This is not always happening for some unknown reason. In such a case, the Cloudbeat process stays idle with a "Failed" status(I killed the process manually after 10 min).


References
1. Fleet Management [Stop](https://github.com/elastic/beats/blob/6b156e6766379a510ec4739e4e3d59446fe8d41a/x-pack/libbeat/management/manager.go#L140)  
4. Elastic Agent Client [Stop](https://github.com/elastic/elastic-agent-client/blob/main/pkg/client/client.go#L137)

launcher/launcher_test.go
```diff
@@ -281,28 +308,45 @@ func (s *LauncherTestSuite) TestWaitForUpdates() {
...
+			false,
+		},
+		{
+			"invalid updates",
+			incomingConfigs{
+				{40 * time.Millisecond, configErr},
+			},
+			nil,
+			errorReloadBeaterCreator(),
+			true,
```
amirbenun, line 321: I feel this test case needs to be in a different test function, as those are all success flows and this test covers a failure point.
Eventually, the purpose is to make the test more readable.

amirbenun (commented):
I am a bit confused by what you wrote.
It seems to me that a config update results on doing nothing but log, I am not sure how do we get to a state where we have a cloudbeat process at zombie state.

launcher/launcher_test.go
```diff
@@ -291,18 +302,20 @@ func (s *LauncherTestSuite) TestWaitForUpdates() {
...
+			go func() {
 				time.Sleep(100 * time.Millisecond)
-			}(tcase.configs)
+				mocks.cancel()
+				close(mocks.reloader.ch)
+			}()
 
 			err = sut.run()
-			s.Error(err)
+			s.NoError(err)
```
amirbenun, line 318: Why did you have to change it?

amirbenun (approved):

launcher/launcher_test.go
olegsu: I notice that it always returns an error after the channel is been closed (https://github.com/elastic/cloudbeat/pull/457/files#diff-29b1f6ad1656f858176a4986694417256c06f97b53ccaaad97c5add2a6b01ea1L294), so I have added [this](https://github.com/elastic/cloudbeat/pull/457/files#diff-29b1f6ad1656f858176a4986694417256c06f97b53ccaaad97c5add2a6b01ea1R311-R315) block to close it gracefully.

PR #463, opened 2022-10-24
fix: connect all kind containers to elastic-packge network

olegsu:
**What does this PR do?**
https://github.com/elastic/cloudbeat/commit/eece5057b1ee1988f72c33b967b06d75980db22f updated all the justfile commands to run on a multi-node cluster.
When running local env `just setup-env` connecting only one container (the control-plane) from kind to the elastic-package network, this causes cloudbeat pod to fail to send events as it can resolve the hostname of es.

**Note**
Working on multi-node cluster does not mean that `just build-deploy-cloudbeat` will deploy it in all the nodes.
To do that we need to add the worker node toleration to https://github.com/elastic/cloudbeat/blob/main/deploy/kustomize/base/daemonset.yml#L88 (which is not recommended, for local dev flow imho)


oren-zohar:
@olegsu I'm a bit surprised we have `NoSchedule` toleration for the master node, do we need it?

olegsu:
> @olegsu I'm a bit surprised we have `NoSchedule` toleration for the master node, do we need it?

I am not sure why they exist in the first place, they also exist in the other manifests we have (tests).
I think they can be removed, in such case we need to add `nodeSelector` so that the Cloudbeat will be deployed only in one node (for debugging).

oren-zohar:
@DaveSys911 maybe you can help us solve the mystery here?

justfile
```diff
@@ -77,8 +77,12 @@ elastic-stack-down:
   elastic-package stack down
 
 elastic-stack-connect-kind kind='kind-multi':
-  ID=$( docker ps --filter name={{kind}}-control-plane --format "{{{{.ID}}" ) && \
-  docker network connect elastic-package-stack_default $ID
+  #!/usr/bin/env sh
+
+  containers=$(docker ps | grep "kind-multi" | awk '{ print $1 }')
```
oren-zohar, line 82: what if it's not `kind-multi`?

oren-zohar (changes requested):

justfile
olegsu: you are right, missed it

justfile
```diff
@@ -77,8 +77,12 @@ elastic-stack-down:
...
 
 elastic-stack-connect-kind kind='kind-multi':
-  ID=$( docker ps --filter name={{kind}}-control-plane --format "{{{{.ID}}" ) && \
-  docker network connect elastic-package-stack_default $ID
+  #!/usr/bin/env sh
+
+  containers=$(docker ps | grep {{kind}} | awk '{ print $1 }')
+  for container in $containers; do
+    docker network connect elastic-package-stack_default $container
+  done
```
oren-zohar, line 85: I forgot to mention, we want to avoid adding scripts directly into our jusfile, but instead, calling them from our scripts dir (`./scripts/`) see more info in [here](https://github.com/elastic/security-team/issues/4291)

oren-zohar (changes requested):

oren-zohar (approved):
LGTM 🚀

PR #467, opened 2022-10-25
Fixing the kubernetes version of `kind`

kfirpeled:
The motivation behind fixing the kubernetes version is to solve future issues with the unsupported version.

Currently, kind 0.0.16 deploys k8s version 1.25 by default, which is currently unsupported. Therefore we can't deploy `cloudbeat` properly


kfirpeled:
Already fixed by @DaveSys911

PR #474, opened 2022-10-27
Added elastic-package to hermit

kfirpeled:
Fixed elastic-package version being used


DaveSys911 (approved):
Tested it. LGTM
image

Issue #461, opened 2022-10-24
Cloudbeat multiple panics after days of running


jeniawhite:
**Describe the bug**
After a while, cloudbeat stops running as a process inside the elastic agent.
This is due to panics and the cloudbeat application is being switched to a degraded and later on failed state by the elastic agent.
This means that from that point in time there won't be any cloudbeat running on the agent.

There are 113 log hits of this panic:
```
{
  "@timestamp": [
    "2022-10-07T22:07:01.497Z"
  ],
  "agent.console.name": [
    "cloudbeat"
  ],
  "agent.console.type": [
    "stderr"
  ],
  "agent.ephemeral_id": [
    "147744f6-c228-4918-a1ba-0f3f270e6b45"
  ],
  "agent.id": [
    "243bc9e8-defd-437c-b507-01104f719571"
  ],
  "agent.name": [
    "ip-172-31-41-176.eu-west-1.compute.internal"
  ],
  "agent.type": [
    "filebeat"
  ],
  "agent.version": [
    "8.5.0"
  ],
  "cloud.account.id": [
    "704479110758"
  ],
  "cloud.availability_zone": [
    "eu-west-1c"
  ],
  "cloud.image.id": [
    "ami-08702177b0dcfc054"
  ],
  "cloud.instance.id": [
    "i-0bb80b599dccc509b"
  ],
  "cloud.machine.type": [
    "m5.large"
  ],
  "cloud.provider": [
    "aws"
  ],
  "cloud.region": [
    "eu-west-1"
  ],
  "cloud.service.name": [
    "EC2"
  ],
  "data_stream.dataset": [
    "elastic_agent"
  ],
  "data_stream.namespace": [
    "default"
  ],
  "data_stream.type": [
    "logs"
  ],
  "ecs.version": [
    "8.0.0"
  ],
  "elastic_agent.id": [
    "243bc9e8-defd-437c-b507-01104f719571"
  ],
  "elastic_agent.snapshot": [
    true
  ],
  "elastic_agent.version": [
    "8.5.0"
  ],
  "event.agent_id_status": [
    "verified"
  ],
  "event.dataset": [
    "elastic_agent"
  ],
  "event.ingested": [
    "2022-10-09T08:16:21.000Z"
  ],
  "host.architecture": [
    "x86_64"
  ],
  "host.containerized": [
    false
  ],
  "host.hostname": [
    "ip-172-31-41-176.eu-west-1.compute.internal"
  ],
  "host.id": [
[...]

PR #483, opened 2022-11-02
Map concurrent access panic for events metadata

jeniawhite:
Transformer allocated the `eventMetadata` once and re-used it for all of the events that it sends to ES.
The underlying `elastic-agent-libs/mapstr` implementation is non-concurrent.
This causes a panic of concurrent read/writes in race scenarios when sending events.

Fixes 
- #461

### DIFF
--- a/transformer/events_creator.go
+++ b/transformer/events_creator.go
@@ -39,9 +39,9 @@ const (
 )
 
 type Transformer struct {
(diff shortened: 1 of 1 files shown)


oren-zohar (approved):
LGTM

amirbenun (approved):

PR #490, opened 2022-11-07
Fix launcher stop

amirbenun:


launcher/launcher.go
```diff
@@ -185,6 +196,9 @@ func (l *launcher) waitForUpdates() error {
 // configUpdate applies incoming reconfiguration from the Fleet server to the beater config,
 // and recreate the beater with the new values.
 func (l *launcher) configUpdate(update *config.C) error {
+	l.cfgmtx.Lock()
```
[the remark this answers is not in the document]
amirbenun: agree it's still WIP, I will also mark the PR as such

amirbenun:
Tests passes with `-race` flag on
image

launcher/launcher.go
```diff
@@ -101,28 +108,42 @@ func (l *launcher) Run(b *beat.Beat) error {
...
 func (l *launcher) Stop() {
-	l.log.Info("Beater launcher is about to shut down gracefully")
+	l.log.Info("Launcher is about to shut down gracefully")
+
+	// Stop listening for updates
+	l.reloader.Stop()
+
+	// Stop the beater without interrupting to an update
+	l.cfgmtx.Lock()
+	l.runmtx.Lock()
```
eyalkraft, line 136: I wonder if `defer l.mtx.Unlock()` is relevant also here?

Will the next run happen with a new launcher for sure?

launcher/launcher.go
```diff
@@ -153,32 +174,44 @@ func (l *launcher) stopBeater() {
 
 	// By waiting to the wait group, it make sure that the old beater has really stopped
 	l.wg.Wait()
+	l.log.Info("Launcher shutted the Beater down gracefully")
 }
 
-func (l *launcher) waitForUpdates() error {
+// waitForFinish is the function that keeps Launcher up and running
```
eyalkraft, line 180: Thanks for adding documentation!

launcher/launcher.go
```diff
@@ -153,32 +174,44 @@ func (l *launcher) stopBeater() {
...
+			return
 		}
+
+		l.cfgmtx.Lock()
+		go func() {
+			defer l.cfgmtx.Unlock()
+			if err := l.configUpdate(update); err != nil {
+				l.beaterErr <- fmt.Errorf("failed to update Beater config: %w", err)
+			}
+		}()
```
eyalkraft, line 214: not sure why that requires another routine?
In the case of multiple configurations following one another I think Ir'll be fine to handle them one by one/

eyalkraft: By not opening a new routine I think we can even get rid of `cfgmtx`.
wdyt?

eyalkraft (commented):
Looks good,
++ for the docs and tests!

not sure about a few points

launcher/launcher.go
amirbenun: I think you are right, I will give it a try! 😄

launcher/launcher.go
```diff
@@ -101,48 +98,84 @@ func (l *launcher) Run(b *beat.Beat) error {
...
+		// beater error produces nil, err
+		cfg, err := l.waitForUpdates()
+
+		// If it's not a beater error, should stop the beater
+		if err == nil {
+			l.stopBeater()
+		}
+
+		// If it's a config update let's merge the new config and continue with the next iteration
+		if cfg != nil {
```
eyalkraft, line 134: a little bit dangerous, because the case of a future developer returning `val, err` from `waitForUpdates`

launcher/launcher.go
```diff
@@ -101,48 +98,84 @@ func (l *launcher) Run(b *beat.Beat) error {
...
+			return fmt.Errorf("launcher could not run Beater: %v", err)
+		}
+
+		// Wait for something to happen:
+		// config update produces val, nil
+		// signal produces nil, nil
+		// beater error produces nil, err
+		cfg, err := l.waitForUpdates()
+
+		// If it's not a beater error, should stop the beater
```
eyalkraft, line 128: Are we sure we don't want to stop the beater in case of an error?
Are there any consequences?

eyalkraft (approved):
cool cool cool

launcher/launcher.go
amirbenun: In case it's a beater error the beater is no longer running.

launcher/launcher.go
amirbenun: We have many unit tests protecting that area, and every change here should be carefully checked.

PR #498, opened 2022-11-08
Allure add env info

gurevichdmitry:
This PR adds ability to publish environment information in allure report

image


gurevichdmitry:
@oren-zohar, @kfirpeled Publishing results job will fail due to changes in report raw data - new file environment.properties is added.
This job collects reports from all parallel test jobs and generates common report. Since the file name is the same for all reports in current implementation when unzipping data, a prompt notification appears and prevents to extract reports data.
I've updated unzip command and added overwrite option.
This change was tested on my fork and found as working solution.
Later we will check how to fix overwrite solution and merge all same files

tests/conftest.py
```diff
@@ -71,3 +71,32 @@ def pytest_addoption(parser):
         default='..',
         help='range to run tests on',
     )
+
+
+def get_fixtures():
+    return cloudbeat_agent, k8s
+
+
+def pytest_sessionfinish(session, exitstatus):
```
oren-zohar, line 80: Parameter `exitstatus` is not used

oren-zohar (changes requested):

oren-zohar (approved):

tests/conftest.py
gurevichdmitry: fixed

PR #507, opened 2022-11-09
add configuration required for EKS clusters

gurevichdmitry:
Configuration files that are required to be changed on EKS cluster nodes (config-1, config-2)


oren-zohar:
can we have them as yamls?

kfirpeled (approved):
Consider for ongoing maintenance to add a readme describing what is expected to be found on each configuration, and for which test. It would make a lot of sense trying to troubleshoot issues with the tests/ test findings

oren-zohar (changes requested):
can we have them as yamls?

gurevichdmitry:
> can we have them as yamls?

@oren-zohar the files provided in their original format. In order to avoid format / conversion issues my suggestion is to hold the files as is. Also when we will start working on infrastructure deployment it will be more easy to copy and replace files instead of implementing read / convert / write operations.

oren-zohar (approved):

PR #511, opened 2022-11-15
Fix v2 bundle path

amirbenun:
**What does this PR do?**
Since V2, cloudbeat working directory has changed into a folder that does **not** contain the binary by setting `-E path.data=/usr/share/elastic-agent/state/data/run/cloudbeat/cis_k8s-default`.
As a result, our bundle no longer exists in the new working directory.
In this PR, I changed the bundle's path to be the same as the executable.

**Related Issues**

**Checklist**


go.sum
```diff
@@ -1099,16 +1099,18 @@ github.com/mattn/go-colorable v0.1.4/go.mod h1:U0ppj6V5qS13XJ6of8GYAs25YV2eR4EVc
 github.com/mattn/go-colorable v0.1.6/go.mod h1:u6P/XSegPjTcexA+o6vUJrdnUu04hMope9wVRipJSqc=
 github.com/mattn/go-colorable v0.1.8/go.mod h1:u6P/XSegPjTcexA+o6vUJrdnUu04hMope9wVRipJSqc=
 github.com/mattn/go-colorable v0.1.9/go.mod h1:u6P/XSegPjTcexA+o6vUJrdnUu04hMope9wVRipJSqc=
-github.com/mattn/go-colorable v0.1.12 h1:jF+Du6AlPIjs2BiUiQlKOX0rt3SujHxPnksPKZbaA40=
 github.com/mattn/go-colorable v0.1.12/go.mod h1:u5H1YNBxpqRaxsYJYSkiCWKzEfiAb1Gb520KVy5xxl4=
+github.com/mattn/go-colorable v0.1.13 h1:fFA4WZxdEF4tXPZVKMLwD8oUnCTTo08duU7wxecdEvA=
+github.com/mattn/go-colorable v0.1.13/go.mod h1:7S9/ev0klgBDR4GtXTXX8a3vIGJpMovkB8vQcUbaXHg=
```
oren-zohar, line 1104: go mod tidy?

oren-zohar (changes requested):
do you want to create a script for the bundle copy?

scripts/common.sh
```diff
@@ -36,4 +36,28 @@ get_agent_sha() {
...
+    done
+
+    echo "Copied all the assets to $POD"
+  done
+}
+
+restart_agents() {
+  echo "Agent restart is not supported yet"
+  # for P in $(get_agents); do
+    # exec_pod $POD "elastic-agent restart" # https://github.com/elastic/cloudbeat/pull/458#issuecomment-1308837098
```
oren-zohar, line 61: Does this function do nothing?

scripts/common.sh
```diff
@@ -36,4 +36,28 @@ get_agent_sha() {
 
 _kubectl_node_info() {
  kubectl get node -o go-template="{{(index .items 0 ).status.nodeInfo.$1}}" 
+}
+
+copy_to_agents() {
```
oren-zohar, line 41: can you please add some documentation here?

config/config.go
```diff
@@ -90,6 +93,20 @@ func New(cfg *config.C) (Config, error) {
...
 
+func defaultConfig() (Stream, error) {
+	ret := Stream{
+		Period: 4 * time.Hour,
+	}
+
+	ex, err := os.Executable()
+	if err != nil {
+		return Stream{}, err
+	}
```
oren-zohar, line 104: can you please add documentation for this part?

oren-zohar (changes requested):
Looks good, small comments

scripts/common.sh
amirbenun: I think the right behavior will be restarting the agents after we copy the new artifacts. Instead of deleting this comment, I wanted to preserve it, do you have any idea on how?

oren-zohar (approved):

PR #513, opened 2022-11-15
Bump 8.7

CohenIdo:


oren-zohar (approved):

Makefile
```diff
@@ -1,7 +1,7 @@
 ##############################################################################
 # Variables used for various build targets.
 ##############################################################################
-CI_ELASTIC_AGENT_DOCKER_TAG?=${CLOUDBEAT_VERSION}-SNAPSHOT
+CI_ELASTIC_AGENT_DOCKER_TAG?=8.6.0-SNAPSHOT
```
amirbenun, line 4: Seems like it's not necessary, `CLOUDBEAT_VERSION` should already have the same value.

amirbenun (approved):

Makefile
oren-zohar: Since he bumped to 8.7, it will try to pull an unreleased image

PR #523, opened 2022-11-17
Remove streams struct

amirbenun:
**What does this PR do?**
As part of elastic-agent V2 changes, the beat no longer receives configuration as an array of streams (where only the first one is populated). Instead, it accepts the config itself right away.
This PR removes the stream abstraction from our code for all different running methods.

**Related Issues**

**Checklist**


oren-zohar (approved):
achla 🚀

PR #544, opened 2022-11-22
fix ci for stand-alone agent

gurevichdmitry:
**What does this PR do?**
This PR fixes ci for stand-alone agent

**Related Issues**

**Checklist**


.github/workflows/cloudbeat-ci.yml
```diff
@@ -105,109 +105,159 @@ jobs:
...
       matrix:
         include:
-          - test-target: post_merge_agent
+          - test-target: pre_merge_agent
             range: ''
-            values_files: tests/deploy/values/ci-sa-agent.yml
+            values_file: tests/deploy/values/ci-sa-agent.yml
           - test-target: pre_merge
             range: ''
+            values_file: tests/deploy/values/ci.yml
```
oren-zohar, line 113: is there a need to send it? it's set to `tests/deploy/values/ci.yml` by default

oren-zohar (changes requested):

.github/workflows/cloudbeat-ci.yml
gurevichdmitry: CI flow uses `just deploy-tests-helm ${{ matrix.test-target }} ${{ matrix.values_file }} ${{ matrix.range }}` command.
`just` does not support named args. In case `matrix.values_file` will not be provided and matrix.range will be, the `just` cannot recognize that matrix.range is the 3rd arg and shift left will be performed. This is why all values shall be provided in the matrix.

oren-zohar: so why swap the parm order in the first place?

gurevichdmitry: Some test suites are using range='' (empty state). When passing the range='' to `just` command the CLI interprets it as empty and params shift happens. So swap is required otherwise incorrect values will be passed.

oren-zohar (approved):

PR #517, opened 2022-11-16
Cloudbeat README refactor

oren-zohar:
**What does this PR do?**
Splits our main README to separate docs per type

**Checklist**
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary README/documentation (if appropriate)

PR #547, opened 2022-11-23
feat: cloud environment terraform

olegsu:
**What does this PR do?**
Add terraform plan based on https://github.com/oren-zohar/cloudbeat/tree/elastic-agent-dockerfile.
This will create a cloud-based environment and setup rules and dashboards.

This is just the first step as we need more robust automation around this process that will include:
1. eks/vanilla cluster setup
2. agent deployment
3. kspm installation
4. slack configuration

**Related Issues**
* https://github.com/elastic/security-team/issues/5034


uri-weisman:
@olegsu great initiative!
I wonder if this automation should reside in the security team repo with the rest of the manuals.

olegsu:
> @olegsu great initiative! I wonder if this automation should reside in the security team repo with the rest of the manuals.

Thank @uri-weisman 
I also thought that it might be. In the end, I put it here as it seems to be more code than manual.
In the next steps, this might be a script that will also use custom images for the different components in the stack.
Don't mind moving it though.

oren-zohar:
@uri-weisman @olegsu I think we should start moving some of our documentation to our repo (and link it from the security team repo) I've already started linking between the two in this [PR](https://github.com/elastic/cloudbeat/pull/517)

kfirpeled:
consider adding terraform to hermit
`hermit install terraform`

deploy/cloud/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+**Prerequisite**
+* [terraform](https://www.terraform.io/)
+
+**How To**
+1. Create an [API token](https://cloud.elastic.co/deployment-features/keys) from your cloud console account.
+    1.1 use the token `export EC_API_KEY={TOKEN}`
+2. run `cd deploy/cloud`
+3. run `terraform init`
+4. run `terraform apply --auto-approve` to create the environment from the latest version (the latest version is vary in cloud/regions combinations). To set a specific version use `-var="stack_version=8.5.1"`
```
kfirpeled, line 14: consider adding example of using `LATEST` and `SNAPSHOT` deployments

deploy/cloud/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+To delete the environment:
```
kfirpeled, line 16: consider giving examples how to connect to the deployed environment

like: using the cloud login / `terraform output -json` to reveal the passwords

deploy/cloud/README.md
kfirpeled, line 14: consider giving an example how to name the environment

kfirpeled (approved):
LGTM
added some comments to enhance the instruction

deploy/cloud/variables.tf
```diff
@@ -0,0 +1,54 @@
+## Deployment configuration
+
+variable "ess_region" {
+  default     = "gcp-us-central1"
```
oren-zohar, line 4: Usually, `gcp-us-west2` and `gcp-europe-west1` is the regions where all of the BCs are being published, maybe we should use them instead.

oren-zohar (approved):

deploy/cloud/variables.tf
olegsu: It depends on the "default" flow. 
If we want to `terraform apply` do we want to create a snapshot version to the latest ga?

deploy/cloud/README.md
olegsu: Removed the sensitive keyword.

deploy/cloud/variables.tf
oren-zohar: they also contain `SNAPSHOT`
image

PR #564, opened 2022-12-04
Resolve panic on race between data and beater

jeniawhite:
**What does this PR do?**
Due to the reconfiguration loop, I saw that `cloudbeat` panics sporadically.
While investigating the code, I noticed a race condition between the initialization of the wait group pointer to the closure of the `data` layer.
There is also a race for the publisher client that I've found, which will be covered later on.

**Checklist**


amirbenun (approved):

oren-zohar (changes requested):
can we add test that fails with the previous struct? (closing before calling fetch)

resources/fetchersManager/data.go
```diff
@@ -34,7 +34,7 @@ type Data struct {
 	timeout  time.Duration
 	interval time.Duration
 	fetchers FetchersRegistry
-	wg       *sync.WaitGroup
+	wg       sync.WaitGroup
```
oren-zohar, line 37: can you please also add documentation for this wait group purpose?

oren-zohar (changes requested):

oren-zohar (approved):

PR #573, opened 2022-12-07
Fixing Packaging workflow using chatGPT 

oren-zohar:
**What does this PR do?**
The encoding change did not cause the issue, but using `with` to fix the linters incorrectly was the root cause. I used ChatGPT in order to confirm it, very impressive:
image

image


eyalkraft:
🤯🤯🤯

PR #581, opened 2022-12-15
Extend the IAM provider to fetch user data

uri-weisman:
**What does this PR do?**
Adds new capability to the IAM fetcher - fetch and aggregate data regarding IAM users.

**Must be delivered with:** 
- https://github.com/elastic/csp-security-policies/pull/154

**Checklist**


resources/providers/awslib/iam/iam.go
```diff
@@ -51,9 +52,19 @@ type PasswordPolicy struct {
...
 func NewIAMProvider(log *logp.Logger, cfg aws.Config) *Provider {
-	svc := iam.NewFromConfig(cfg)
+	svc := iamsdk.NewFromConfig(cfg)
 	return &Provider{
 		log:    log,
 		client: svc,
 	}
 }
+
+type IAMClient interface {
```
uri-weisman, line 62: Required for mocking the aws-sdk.

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 )
 
 require (
+	github.com/aws/aws-sdk-go v1.44.162
```
oren-zohar, line 35: seems to me like we don't use it, and use `aws-sdk-go-v2` instead, couldn't find any reference for it

resources/fetchers/iam_fetcher.go
```diff
@@ -46,25 +46,36 @@ type IAMResource struct {
...
 func (f IAMFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
 	f.log.Debug("Starting IAMFetcher.Fetch")
+	iamResources := make([]awslib.AwsResource, 0)
 
 	pwdPolicy, err := f.iamProvider.GetPasswordPolicy(ctx)
 	if err != nil {
-		return err
+		f.log.Errorf("Unable to fetch PasswordPolicy, error: %v", err)
+	} else {
+		iamResources = append(iamResources, pwdPolicy)
```
oren-zohar, line 55: how will logging the error instead of returning it will affect the flow? is this by design?

resources/fetchers/iam_fetcher.go
```diff
@@ -46,25 +46,36 @@ type IAMResource struct {
 
 func (f IAMFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
```
oren-zohar, line 47: would you mind adding documentation of what is being fetched (pwd policy and IAM user)?

resources/providers/awslib/iam/iam.go
oren-zohar, line 62: The name starts with the package name, you can change to `Client`

oren-zohar (changes requested):
Looks good, small comments

go.mod
uri-weisman: True, removed with `go mod tidy`.
Might be nice to add to the CI.

resources/fetchers/iam_fetcher.go
uri-weisman: In case we were able to fetch a certain resource but fail to get the other, IMO, a better experience will be to generate some findings rather than nothing.

We have similar behavior in the `file system fetcher`.

resources/fetchers/iam_fetcher.go
uri-weisman: Sure thing

oren-zohar (approved):

Issue #239, opened 2022-06-23
Support health status interface


eyalkraft:
**Motivation**

Once libbeat is migrated to v2 cloudbeat should support the new health status interface to report detailed health information to the user. Including requiring the user to upgrade.
This information will surfaced in the UI. which is being tracked in .

**Definition of done**

- [ ] If there's an error, cloudbeat should report it an a way viewable from Kibana
- [ ] If there's a mismatch between the integration and the cloudbeat version (which could mean missing rules/fetchers on cloudbeats side) cloudbeat should indicate the user he needs to upgrade his agent.
- [ ] make sure we understand what is the behavior once the status is not healthy - does the agent restarts cloudbeat?

**Out of scope**

What is not included in this task  

**Related tasks/epics**

- https://github.com/elastic/security-team/issues/3494
- https://github.com/elastic/package-spec/issues/165#issuecomment-1156497651
- [Design doc](https://docs.google.com/document/d/1I0PeZKIkUTaWiY2F_Dum_gSi0cpbblcftl-tz81iYak/edit)
- blocked on https://github.com/elastic/elastic-agent/issues/100

yashtewari:
[BeatV2Manager](https://github.com/elastic/beats/blob/main/x-pack/libbeat/management/managerV2.go) with the new status/health update interfaces is not yet as mature as we'd expected. It's not being used by the "core" beats yet, as described by this issue
- https://github.com/elastic/ingest-dev/issues/1367

So it would be risky to touch it for the v8.6.0 release. I'll add more details about this in another comment.

For now, I've setup the required information/logic for version compatibility, and also started using V1 of [`Manager.UpdateStatus`](https://github.com/elastic/beats/blob/4f106cb16e58272203b411a7eed0678989919c98/libbeat/management/management.go#L66) so that it may be easier to update to V2 going forward.

- https://github.com/elastic/integrations/pull/4695
- https://github.com/elastic/cloudbeat/pull/538

The current behavior is the same as any other error in cloudbeat: the error is logged and cloudbeat is restarted by the Agent. But this is only tentative and we should make a decision on it.

yashtewari:
I was able to send a `Degraded` signal using the V1 interface and make the Agent status on Kibana change to "Unhealthy".

![Screenshot 2022-11-23 at 5 04 23 PM](https://user-images.githubusercontent.com/9020112/203579722-919b2dba-2fc3-4413-8055-41fb1a3ff723.png)

We have two options now:
[...]

PR #583, opened 2022-12-15
feat: validate benchmark type

olegsu:
**What does this PR do?**
This will use the new `config.v1` structure to validate that the cloudbeat knows how to run the selected benchmark.
In case the benchmark is not supported it may fail fast or ignore and update the status.
1. When cloudbeat just started and the given configuration is based on a benchmark that is not supported, fail fast as there is nothing to do.
2. When a reconfiguration comes with an unsupported benchmark ignore it and update the status

This PR only uses `config.v1.benchmark` to not deal with the new structure that integrations [pr](https://github.com/elastic/integrations/pull/4752) adding ( cc @kfirpeled)

**Related Issues**
Related https://github.com/elastic/cloudbeat/issues/239
Related https://github.com/elastic/integrations/pull/4752

**Checklist**


scripts/common.sh
```diff
@@ -57,8 +57,7 @@ copy_to_agents() {
 }
 
 restart_agents() {
-  echo "Agent restart is not supported yet"
-  # for P in $(get_agents); do
-    # exec_pod $POD "elastic-agent restart" # https://github.com/elastic/cloudbeat/pull/458#issuecomment-1308837098
-  # done
+  for P in $(get_agents); do
+    exec_pod $POD "elastic-agent restart" # https://github.com/elastic/cloudbeat/pull/458#issuecomment-1308837098
```
amirbenun, line 61: You can remove the comment.
Also, I just noticed we are missing `POD=$(echo $P | cut -d '/' -f 2)`.
Can you verify that it works?

config/config.go
```diff
@@ -38,25 +40,21 @@ const DefaultNamespace = "default"
...
 type Config struct {
-	Type       string                  `config:"type"`
 	AWSConfig  aws.ConfigAWS           `config:",inline"`
 	RuntimeCfg *RuntimeConfig          `config:"runtime_cfg"`
 	Fetchers   []*config.C             `config:"fetchers"`
 	KubeConfig string                  `config:"kube_config"`
 	Period     time.Duration           `config:"period"`
 	Processors processors.PluginConfig `config:"processors"`
 	BundlePath string                  `config:"bundle_path"`
+	Benchmark  string                  `config:"config.v1.benchmark"`
```
amirbenun, line 57: This change requires a matching change to the integration, isn't it?

launcher/launcher.go
```diff
@@ -239,6 +242,9 @@ func (l *launcher) reconfigureWait(timeout time.Duration) (*config.C, error) {
 				err := l.validator.Validate(update)
 				if err != nil {
 					l.log.Errorf("Config update validation failed: %v", err)
+					if errors.Is(err, cloudbeat_config.ErrBenchmarkNotSupported) {
```
amirbenun, line 245: The launcher was designed in a way it is agnostic to cloudbeat or any `Beater` implementation.
In that case, we can extend the `Validator` to support a special kind of error that will send that degraded message.
The only difference will be that the `Validator` accepts `config.C` instead of our struct.

amirbenun (changes requested):
Nice!

scripts/common.sh
olegsu: You are right, will verify

config/config.go
olegsu: We already have it https://github.com/elastic/integrations/pull/4752/files#diff-721297d41d05ea41cf84e1ee80fdda5ac230452caef26678a37e752e43fbb466R11

launcher/launcher.go
olegsu: In that case, we can update the validator behavior to return a boolean whether the configuration is valid or not.
So the validator will also receive the `management.Manager` to update the status.

I am not sure it's better as we still have a coupling between the launcher and the validator.

errors/errors.go
```diff
@@ -0,0 +1,17 @@
+package errors
+
+// Cloudbeat error is an error that is desgined to have some information that
+// can help to end user to operate cloudbeat
+// for example when a cloudbeat configuration is invalid, this error will include
+// more information about what is missing/expected and maybe links to external sources as well
+type CloudbeatError struct {
```
amirbenun, line 7: I liked it, since I still don't want the launcher to be aware of cloudbeat I would just rename it.
Maybe `BeaterDegradeError` / `BeaterUnhealthyError` or some other name that indicates what is about to happen once this error is returned.
Also, it should reside near the launcher IMO.

olegsu: This is good naming, will change the name.

> Also, it should reside near the launcher IMO.
This might be an issue with imports

launcher/launcher.go
```diff
@@ -239,6 +242,10 @@ func (l *launcher) reconfigureWait(timeout time.Duration) (*config.C, error) {
 				err := l.validator.Validate(update)
 				if err != nil {
 					l.log.Errorf("Config update validation failed: %v", err)
+					heatlhErr := &cb_errors.BeaterUnhealthyError{}
+					if errors.As(err, heatlhErr) {
```
amirbenun, line 246: Let's test this new functionality, it can be a part of `TestLauncherValidator`

amirbenun: Also, note that you are checking the error that returned from the `Validator` but `BeaterUnhealthyError` error is actually returned from `config.New` and wrapped in the validator and probably won't get detected here.

errors/errors.go
```diff
@@ -0,0 +1,37 @@
...
+
+// BeaterUnhealthyError error is an error that is desgined to have an information that
+// can help to end user to operate cloudbeat health issues.
+// For example, when a cloudbeat configuration is invalid, the error will include
+// more information about what is missing/expected and might have links to external sources as well
+type BeaterUnhealthyError struct {
+	msg string
+}
+
+func New(msg string) BeaterUnhealthyError {
```
amirbenun, line 31: Nit, rename the function to indicate the error type, maybe `errors.NewBeaterUnhealthy(...)`?

amirbenun (commented):
Last touch

launcher/launcher.go
olegsu: > Let's test this new functionality, it can be a part of `TestLauncherValidator`

Sure



> Also, note that you are checking the error that returned from the `Validator` but `BeaterUnhealthyError` error is actually returned from `config.New` and wrapped in the validator and probably won't get detected here.

The [test](https://github.com/elastic/cloudbeat/pull/583/files#diff-5a256ba4448e51ae85f5cdae75f5093675bc7ea190d00bcf687fffa782822ca8R31) for the errors pkg checking exactly this. First create the `BeaterUnhealthyError` and than wrap it, later test that using `errors.As` it will get the original error.

From golang https://pkg.go.dev/errors#As
> As finds the first error in err's chain that matches target, and if one is found, sets target to that error value and returns true. Otherwise, it returns false.

amirbenun (approved):

gurevichdmitry (approved):
@olegsu LGTM. scripts/common.sh approved.

PR #585, opened 2022-12-18
Fixing EKS cluster name

ofiriro3:
**What does this PR do?**
This PR fixes the panic that identifying the EKS cluster name caused.

**Related Issues**
- Related: https://github.com/elastic/security-team/issues/5654


oren-zohar:
should we backport this to 8.6?

justfile
```diff
@@ -43,6 +46,10 @@ build-binary:
   GOOS=linux go mod vendor
   GOOS=linux go build -v
 
+build-binary-eks:
+  GOOS=linux go mod vendor
+  GOOS=linux GOARCH=amd64 go build -v
```
oren-zohar, line 51: what makes it an EKS binary? I don't think we need to add another specific command like this one, especially when you can add a parameter to the original command for the `OS` and `ARCH`

justfile
```diff
@@ -30,6 +30,9 @@ build-deploy-cloudbeat-debug: build-cloudbeat-debug load-cloudbeat-image deploy-
 build-replace-cloudbeat: build-binary
   ./scripts/remote_replace_cloudbeat.sh
 
+build-replace-cloudbeat-eks: build-binary-eks
+  ./scripts/remote_replace_cloudbeat.sh
```
oren-zohar, line 34: can be removed, see my next comment

oren-zohar (changes requested):
Unit tests fail. Also, I'm not sure how this fix the issue, can you please explain?

justfile
ofiriro3: How can you add the params to the JUST `go build` command only?

I just thought it would be easier if we would have an easy command that knows how to replace EKS because the current command doesn't work.
I wouldn't expect anyone that hasn't worked on EKS to know that this is necessary for it to run.

If you still find it redundant after what I wrote I'll remove it.

ofiriro3:
> should we backport this to 8.6?

Nope, it was introduced only on 8.7.

> Unit tests fail. Also, I'm not sure how this fix the issue, can you please explain?

I'll fix the unit test.
See that one of the `AwsConfigProvider` was not initialized with a `MetadataProvider` provider, and it caused a nil pointer exception when the `AwsConfigProvider` ran

justfile
oren-zohar: you can add params to any just command, for example:

```justfile
my-command param_1 param_2:
  command p1={{param_1}} p2={{param_2}}
```

not only that, parameters prefixed with a `$` will be exported as environment variables. See more [here](https://github.com/casey/just#getting-and-setting-environment-variables)

ofiriro3: So I tried two alternatives:

1. Concatenating parameters -  
```justfile
build-binary GOARCH='':
  GOOS=linux go mod vendor
  GOOS=linux {{GOARCH}} go build -v
```

But I couldn't find a way to run `build-replace-cloudbeat` with a parameter.
For example, this code doesn't work:
```justfile
build-replace-cloudbeat GOARCH='' : build-binary {{GOARCH}} 
  ./scripts/remote_replace_cloudbeat.sh
```

2. I can add an environment variable to the entire JUST file, and then I can use it in the script.
But I am wondering how people will know to change it.

I thought about maybe doing something like
# For an EKS cluster please set the GOARCH as 'GOARCH=amd64'
GOARCH := ''

WDYT? Do you have a better suggestion?

oren-zohar: > not only that, parameters prefixed with a $ will be exported as environment variables. See more [here](https://github.com/casey/just#getting-and-setting-environment-variables)

oren-zohar (commented):
is it possible to add a test to validate the fix btw?

justfile
ofiriro3: reverted

ofiriro3:
> is it possible to add a test to validate the fix btw?

The functionality that was missing here was a parameter for a provider struct.
The only way that I see that we can test it properly is to add a real EKS integration since it uses an AWS library.

If we do find it very important, I'll need to change the `config_provider.go` [link](https://github.com/elastic/cloudbeat/blob/ac699883d4ec9fe9d546a53727631215651fc6cb/resources/providers/awslib/config_provider.go) so instead of using `aws.InitializeAWSConfig(cfg)` we will inject an interface has this function. And I guess that it will enforce us to do more changes in tests/structs/provider to work with it so they will inject it as well

oren-zohar (approved):

PR #594, opened 2022-12-21
Fix OPA package build

uri-weisman:
**What does this PR do?**

Change the way we build OPA's bundle.
Before the change, we used the existing csp-security-policies go mod dependency (didn't require fetching the repo) - it was a dirty hack and we don't hold a `go.mod` in the policies repo anymore.

The new solution clones the csp-security-policies, find the latest tag which is not pre-release, and build the OPA bundle.

**Related Issues**
- https://github.com/elastic/csp-security-policies/pull/151

**Checklist**


eyalkraft:
How can we easily know with what version (commit/tag) of the policies repo is a specific cloudbeat commit built?
How can we control it?

(yet another reason to merge the repos :P)

uri-weisman:
> How can we easily know with what version (commit/tag) of the policies repo is a specific cloudbeat commit built? How can we control it?
> 
> (yet another reason to merge the repos :P)

Definitely another reason to merge.
I'm logging the commit + tag that is used for the bundle build.
Currently, we don't have any control over it, I can provide a `tag` param to the `BuildOpaBundle` function in case we want to override the release.

oren-zohar (changes requested):
Fetching the latest release is a breaking change. We need to discuss the right strategy before doing so (adding automation for it + branching strategy in csp-rego-policies repo) so it's still too early for this change. Thanks for the initiative, either way, we will probably use this PR as ref in the future. Let's keep the previous behavior which was fetching a specific release version, we can set this version in `version/policy.go`

magefile.go
```diff
@@ -340,11 +339,43 @@ func PythonEnv() error {
...
+	}
+
+	// Find the commit associated with the relevant policy version tag
+	policyVersion := version.PolicyVersion().Version
+	ref, err := repo.Tag(policyVersion)
+	if err != nil {
+		return err
+	}
+
+	log.Printf("Latest production release commit hash:", ref.Hash())
```
oren-zohar, line 367: missing a placeholder

magefile.go
```diff
@@ -340,11 +339,43 @@ func PythonEnv() error {
...
 	if err != nil {
 		return err
 	}
 
-	return sh.Run("bin/opa", "build", "-b", cspPoliciesPkgDir+"/bundle", "-e", cspPoliciesPkgDir+"/bundle/compliance")
+	if err = wt.Checkout(&git.CheckoutOptions{Hash: ref.Hash()}); err != nil {
+		return err
+	}
+
+	sh.Run("bin/opa", "build", "-b", cspPoliciesPkgDir+"/bundle", "-e", cspPoliciesPkgDir+"/bundle/compliance")
```
oren-zohar, line 378: error handling could be useful here

oren-zohar (changes requested):
Thanks :pray: small comments

amirbenun:
This PR actually goes against our decision on [Define new release process for csp-security-policies](https://github.com/elastic/security-team/issues/2979).
Once a PR is merged to csp-security-policies repository it will be much easier to upgrade the dependency in cloudbeat using a `go get` command (as it used to be before that PR).
Can someone share more information about the reasoning behind that change?

oren-zohar:
> This PR actually goes against our decision on https://github.com/elastic/security-team/issues/2979.

How does it go against it?

> Once a PR is merged to csp-security-policies repository it will be much easier to upgrade the dependency in cloudbeat using a go get command (as it used to be before that PR).

https://github.com/elastic/csp-security-policies is no longer a go package (since we removed all go files and go.mod)

amirbenun:
I wasn't realized we removed the go files, makes more sense then.
And it goes against our decision because the dependency PRs to cloudbeat will be easier to automate with a `go get` command instead of changing a specific line of a specific file.

uri-weisman:
> I wasn't realized we removed the go files, makes more sense then. And it goes against our decision because the dependency PRs to cloudbeat will be easier to automate with a `go get` command instead of changing a specific line of a specific file.

As @oren-zohar mentioned, the CSP-policies can't be a go package anymore so this PR was a quick fix.
The following [code](https://github.com/elastic/cloudbeat/pull/594/commits/22a648e1571800289da376a57a64a87568886428#diff-520109d035a196d29d0a86e9c4e6c98e98abf056141b84df68c0d48167b87b25R381-L409) can help us with automating this process in the future, it basically retrieves the commit hash of the latest prod release.

PR #597, opened 2022-12-21
Remove runtime config logic

jeniawhite:
**What does this PR do?**
Removes the runtime config logic

**Checklist**
- [ ] TODO: Need to bump the policy version prior to merging this PR


beater/validator.go
```diff
@@ -30,14 +30,12 @@ import (
 type validator struct{}
 
 func (v *validator) Validate(cfg *agentconfig.C) error {
-	c, err := config.New(cfg)
+	// TODO: Should we check something?
+	// Benchmark is being checked inside config.New
+	_, err := config.New(cfg)
```
oren-zohar, line 35: I feel like the validation inside config should be here

```go
if c.Benchmark != "" {
if !isSupportedBenchmark(c.Benchmark) {
	return c, ErrBenchmarkNotSupported
	}
}
```

oren-zohar (changes requested):
looking at the new config.v1 format, it seems like we moved the credentials to there also, yet we still read our credentials from the flat version. Not sure if we should also address this change here, to fully migrate to the new format wdyt?

```yaml
config:
  v1:
    posture: {{posture}}
    deployment: {{deployment}}
    benchmark: cis_eks
    aws:
      credentials:
        {{#if access_key_id}}
        access_key_id: {{access_key_id}}
        {{/if}}
        {{#if secret_access_key}}
        secret_access_key: {{secret_access_key}}
        {{/if}}
        {{#if session_token}}
        session_token: {{session_token}}
        {{/if}}
        {{#if shared_credential_file}}
        shared_credential_file: {{shared_credential_file}}
        {{/if}}
        {{#if credential_profile_name}}
        credential_profile_name: {{credential_profile_name}}
        {{/if}}
        {{#if role_arn}}
        role_arn: {{role_arn}}
        {{/if}}
        type: {{aws.credentials.type}}
```

Also, did you test this E2E? @jeniawhite

beater/validator.go
jeniawhite: I've considered moving the validation to the validator but that would require us to move logic from the `config` package to the `beater` package.
Which requires recoding parts of the `cloudbeat` like `newCloudbeat` signatures and etc...
Reconsidering the existing solution I did not see any value in the `validator`, because we use the same validation for non-fleet managed configs and reconfigurations.
This is why I've consolidated the validity checks in the config constructor and removed the validator.

jeniawhite:
@oren-zohar 
Yeah, I've run the E2E test once again after removing the validator and consolidating logic inside the config and I can see the same amount of findings as the 8.7 SNAPSHOT without my changes.
Screen Shot 2022-12-25 at 18 48 00

amirbenun (commented):
IMO, we should keep decoupling cloudbeat and the launcher.
For that PR, it means that the validator should remain as an abstraction layer for beater specific use cases. Even if the validator only calls another validation function for the config.

jeniawhite:
@amirbenun Rolled back to the commit that leaves the `validator`.
I Ran an E2E test again to verify that I get everything:
Screen Shot 2022-12-27 at 16 54 40

evaluator/opa.go
```diff
@@ -117,8 +120,8 @@ func (o *OpaEvaluator) Eval(ctx context.Context, resourceInfo fetching.ResourceI
 	}
 
 	result, err := o.decision(ctx, OpaInput{
-		Result:         fetcherResult,
-		ActivatedRules: o.activatedRules,
+		Result:    fetcherResult,
+		Benchmark: o.benchmark,
```
amirbenun, line 124: I am missing some context here, did we change all the rules in a way they will run only when their benchmark is on?

version/policy.go
```diff
@@ -17,7 +17,7 @@
 
 package version
 
-const policyVersion = "v1.2.3"
+const policyVersion = "v1.2.4"
```
amirbenun, line 20: I guess that's your matching change for the rules?

amirbenun (approved):
LGTM

oren-zohar (approved):

PR #598, opened 2022-12-25
IAM fetcher - retrieve root account data

uri-weisman:
**What does this PR do?**
1. Adds root account credential information.
2. Fetch IAM credential report for enriching the IAM user struct.

**Related Issues**

**Checklist**


resources/providers/awslib/iam/user.go
```diff
@@ -165,55 +162,118 @@ func (p Provider) getMFADevices(ctx context.Context, user types.User) ([]AuthDev
...
+		},
+	}
+}
 
-			var lastUsed time.Time
-			if err == nil {
-				if output.AccessKeyLastUsed != nil && output.AccessKeyLastUsed.LastUsedDate != nil {
-					lastUsed = *output.AccessKeyLastUsed.LastUsedDate
+func (p Provider) getCredentialReport(ctx context.Context) (map[string]*CredentialReport, error) {
+	var ae smithy.APIError
```
olegsu, line 190: Seems a short name that might be harder to understand later.
Consider moving the declaration to https://github.com/elastic/cloudbeat/pull/598/files#diff-32fa8082bf35af27930c2716c52e4c641ffbdfb6ffddfa220baada07bd959fb8R194 as it closer to the usage, or change it to something more explicit (`apiErr`?)

resources/providers/awslib/iam/user.go
```diff
@@ -165,55 +162,118 @@ func (p Provider) getMFADevices(ctx context.Context, user types.User) ([]AuthDev
...
+	report, err := p.client.GetCredentialReport(ctx, &iamsdk.GetCredentialReportInput{})
+	if err != nil {
+		var awsFailErr *types.ServiceFailureException
+		if errors.As(err, &awsFailErr) {
+			return nil, errors.Wrap(err, "could not gather aws iam credential report")
+		}
+
+		// if we have an error, and it is not a server err we generate a report
+		if errors.As(err, &ae) {
+			if ae.ErrorCode() == "ReportNotPresent" || ae.ErrorCode() == "ReportExpired" {
```
olegsu, line 200: `ReportNotPresent` and `ReportExpired` are not exported as consts?

resources/providers/awslib/iam/user.go
```diff
@@ -165,55 +162,118 @@ func (p Provider) getMFADevices(ctx context.Context, user types.User) ([]AuthDev
...
+				// generate a new report
+				_, err := p.client.GenerateCredentialReport(ctx, &iamsdk.GenerateCredentialReportInput{})
```
olegsu, line 202: What is the usage of this? seems that you ignore the report output

resources/providers/awslib/iam/user.go
```diff
@@ -18,36 +18,27 @@
...
+)
 
-type AccessKey struct {
-	AccessKeyId  string
-	Active       bool
-	CreationDate time.Time
-	LastAccess   time.Time
-	HasUsed      bool
-}
+var countRetries = 0
```
olegsu, line 41: When the counter will be zero again?
Can it be part of function closure instead of on the global stack?

olegsu (changes requested):

resources/providers/awslib/iam/user.go
uri-weisman: Unfortunately, the ErrCodeNoSuchEntityException error code does not exist in the v2 SDK.

resources/providers/awslib/iam/user.go
uri-weisman: `GenerateCredentialReport` - generate an IAM credential report.
We don't care about the output as it contains the status of the generated report, we get the same logic by calling `report, err = p.client.GetCredentialReport(ctx, &iamsdk.GetCredentialReportInput{})`
and check for the error type.

resources/providers/awslib/iam/user.go
uri-weisman: Sure, good catch!

resources/providers/awslib/iam/root_account.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	iamsdk "github.com/aws/aws-sdk-go-v2/service/iam"
+	"github.com/aws/aws-sdk-go-v2/service/iam/types"
+	"strings"
+	"time"
+)
+
+func (p Provider) createRootAccountUser(rootAccount *CredentialReport) *types.User {
```
olegsu, line 29: The name of the function confuses me as it not really creating any user in the root account.
I can suggest something like `p.getUser(root)`

olegsu (approved):
LGTM, just a small suggestion

resources/providers/awslib/iam/user_test.go
```diff
@@ -72,55 +73,61 @@ func Test_GetUsers(t *testing.T) {
...
+					HasUsed:      true,
+					MfaActive:    false,
 					IsVirtualMFA: false,
-					Arn:          "arn:aws:iam::123456789012:user/test-user-2",
+					Arn:          "arn:aws:iam::1234567890:root",
 				},
 			},
-			wantErr: false,
+			wantErr:       false,
+			expectedUsers: 3,
```
gurevichdmitry, line 104: Test case name: Test 2: Should return _two_ users
expectedUsers: 3

resources/providers/awslib/iam/test_data.go
```diff
@@ -19,20 +19,13 @@ package iam
...
-	keyMetadata = []types.AccessKeyMetadata{
-		{
-			AccessKeyId: aws.String("test_access_key"),
-			CreateDate:  nil,
-			Status:      types.StatusTypeActive,
-			UserName:    aws.String("username"),
-		},
-	}
-
 	virtualMfaDevices = []types.MFADevice{
```
gurevichdmitry, line 29: Since User properties (name and arn) were changed please check might be also relevant to MFADevice definition.

PR #620, opened 2022-12-29
Cloudformation script for elastic-agent on EC2

amirbenun:
**What does this PR do?**
Add CloudFormation script for deploying elastic-agent on an EC2 instance.
The script creates a role for that EC2 with the relevant permissions to run vulnerability management and attaches the newly created role to the EC2.
The EC2 instance will be accessed by ssh from everywhere given the ssh key.

**Related Issues**
- Resolves: https://github.com/elastic/security-team/issues/5659

**Checklist**

### DIFF
--- a/deploy/cloudformation/elastic-agent-ec2.yml
+++ b/deploy/cloudformation/elastic-agent-ec2.yml
@@ -1,0 +1,200 @@
+---
+AWSTemplateFormatVersion: '2010-09-09'
+Description: Creates an EC2 instance with permissions to run elastic-agent
+
+Parameters:
+  LatestAmiId:
+    Type: 'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>'
+    Default: '/aws/service/canonical/ubuntu/server/jammy/stable/current/amd64/hvm/ebs-gp2/ami-id'
+  KeyName:
+    Type: AWS::EC2::KeyPair::KeyName
+    Description: SSH Keypair to login to the instance
+    Default: amir-build
+  InstanceType:
+    Type: String
+    Default: t2.micro
+    AllowedValues:
+    - t2.micro
+    - t2.small
+    - t2.medium
+    Description: The type of EC2 instance to create
+  EnrollmentToken:
+    Type: String
+    Description: The enrollment token of elasitc-agent
+  FleetUrl:
+    Type: String
+    Description: The fleet URL of elastic-agent
+  ElasticAgentVersion:
+    Type: String
+    Description: The version of elastic-agent to install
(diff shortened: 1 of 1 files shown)


deploy/cloudformation/elastic-agent-ec2.yml
```diff
@@ -0,0 +1,200 @@
...
+      Tags:
+        - Key: Name
+          Value: Ealstic Agent
+
+  # Internet Attachment for VPC
+  ElasticAgentAttachGateway:
+    Type: AWS::EC2::VPCGatewayAttachment
+    Properties:
+      VpcId: !Ref ElasticAgentVpc
+      InternetGatewayId: !Ref ElasticAgentInternetGateway
```
eyalkraft, line 82: > The EC2 instance will be accessed by ssh from everywhere given the ssh key.


I don't think we should have access from anywhere by default.
This is important for debugging but I think we can go for a more restrictive approach by default, I don't see a reason for our customers to ssh into the machine.

Can you add a checkbox in the cloudformation UI that will determine the relevant parts in the template are active or not?

This way the checkbox can be not checked by default, and we/customers can check it if we/they want to have debuggability capabilities.

If making it optional isn't easily possible, I don't think we should have that as part of our template at all.

eyalkraft (changes requested):
Nice!

deploy/cloudformation/elastic-agent-ec2.yml
amirbenun: I opened another task to discuss the security issues of the new deployment method.
[Securing elastic-agent EC2 CloudFormation template](https://github.com/elastic/cloudbeat/issues/627).

oren-zohar (changes requested):
Can you please add documentation? what it deployed, and how can we test it?

eyalkraft (approved):

deploy/cloudformation/elastic-agent-ec2.yml
```diff
@@ -0,0 +1,133 @@
...
+    Type: String
+    Default: t2.micro
+    AllowedValues:
+    - t2.micro
+    - t2.small
+    - t2.medium
+    Description: The type of EC2 instance to create
+  EnrollmentToken:
+    Type: String
+    Description: The enrollment token of elasitc-agent
```
oren-zohar, line 22:
```suggestion
    Description: The enrollment token of elastic-agent
```

oren-zohar (approved):

oren-zohar (changes requested):
please fix the **Manifest Tests** before merge

oren-zohar (approved):

PR #628, opened 2023-01-02
Move launcher errors under launcher package

amirbenun:
**What does this PR do?**
Moving launcher errors to be defined on the same package as the launcher.
The motivation behind maintaining the two on the same package is that the launcher already has the errors as a dependency. The launcher can report a different status to the fleet based on the returned error. Which means, the errors better be defined on the same package.

**Related Issues**

**Checklist**

### DIFF
--- a/launcher/launcher.go
+++ b/launcher/launcher.go
@@ -241,7 +241,7 @@ func (l *launcher) reconfigureWait(timeout time.Duration) (*config.C, error) {
 				err := l.validator.Validate(update)
 				if err != nil {
 					l.log.Errorf("Config update validation failed: %v", err)
-					heatlhErr := &BeaterUnhealthyError{}
+					healthErr := &BeaterUnhealthyError{}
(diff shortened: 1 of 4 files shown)


launcher/launcher.go
```diff
@@ -242,7 +241,7 @@ func (l *launcher) reconfigureWait(timeout time.Duration) (*config.C, error) {
 				err := l.validator.Validate(update)
 				if err != nil {
 					l.log.Errorf("Config update validation failed: %v", err)
-					heatlhErr := &cb_errors.BeaterUnhealthyError{}
+					heatlhErr := &BeaterUnhealthyError{}
```
uri-weisman, line 244: It's before your change but can you plz fix the typo?
```suggestion
					healthErr := &BeaterUnhealthyError{}
```

uri-weisman (approved):

olegsu (approved):

Issue #543, opened 2022-11-22
Implement an AWS Network fetcher


uri-weisman:
**Motivation**
Create a fetcher to collect VPC data to be evaluated by OPA.

**Definition of done**
What needs to be completed at the end of this task
- [ ] use defsec 3rd party to fetch the data from the AWS account. 
- [ ] All the required VPC data is collected.
- [ ] Fetcher's implementation is provider agnostic.

**Related tasks/epics**
- https://github.com/elastic/csp-security-policies/issues/149 probably

olegsu:
Rules to implement
- [x] 5.1 Ensure no Network ACLs allow ingress from 0.0.0.0/0 to remote server administration ports 
- [ ] 5.2 Ensure no security groups allow ingress from 0.0.0.0/0 to remote server administration ports
- [ ] 5.3 Ensure no security groups allow ingress from ::/0 to remote server administration ports
- [ ] 5.4 Ensure the default security group of every VPC restricts all traffic

PR #637, opened 2023-01-03
feat: add security groups fetcher

olegsu:
**What does this PR do?**
Add CIS AWS 5.2

**Related Issues**
- Related  https://github.com/elastic/csp-security-policies/pull/169
- Related https://github.com/elastic/cloudbeat/issues/543

**Checklist**


resources/fetchers/network_fetcher.go
```diff
@@ -39,24 +39,28 @@ type ACLFetcherConfig struct {
...
-func (f ACLFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
-	f.log.Debug("Starting ACLFetcher.Fetch")
+// Fetch collects network resource such as network acl and security groups
+func (f NetworkFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
+	f.log.Debug("Starting NetworkFetcher.Fetch")
 
-	resources, err := f.aclProvider.DescribeNeworkAcl(ctx)
+	nacl, err := f.aclProvider.DescribeNeworkAcl(ctx)
 	if err != nil {
 		return err
```
uri-weisman, line 53: I'm not sure about this behavior, in case we fail to retrieve a certain resource do we want to return an error and stop fetching the rest of the network resources? 

In the `IAM fetcher`,  I continue with the fetching cycle even if some error occurred, anyway whatever we decide -the logic should be the same.

resources/fetchers/network_fetcher.go
```diff
@@ -39,24 +39,28 @@ type ACLFetcherConfig struct {
...
 	}
 
-	for _, resource := range resources {
+	securityGroups, err := f.aclProvider.DescribeSecurityGroups(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, resource := range append(nacl, securityGroups...) {
```
uri-weisman, line 61: elegant 😎

uri-weisman (changes requested):
Nice!

resources/fetchers/network_fetcher.go
olegsu: Sounds right, will change that

uri-weisman (approved):

PR #643, opened 2023-01-03
Addition of non cert option

jeniawhite:
**What does this PR do?**
Adding an option to run `cloudbeat` without certificate

**Checklist**


README.md
```diff
@@ -75,12 +80,26 @@ Build and deploy cloudbeat on your local kind cluster:
...
 > # just build-deploy-cloudbeat <Target OS> <Target Arch>
 > just build-deploy-cloudbeat linux amd64
 > ```
+>
+> Or without certificate
+>
+> ```zsh
+> # just build-deploy-cloudbeat-nocert <Target OS> <Target Arch>
+> just build-deploy-cloudbeat-nocert linux amd64
+> ```
```
oren-zohar, line 102: can you remove the part about `<Target OS>` and `GOOS` from the docs? it's no longer relevant

oren-zohar (changes requested):
Generally speaking, I think we should avoid adding more commands if we can instead add a parameter to the relevant commands. Additinionly, we also want to avoid adding more deployment config files, is it possible to edit the original kustomization file to toggle the cert part? @olegsu?

olegsu:
> is it possible to edit the original kustomization file to toggle the cert part? @olegsu?
Our options are
1. Use `kustomize` cli to update the kustomization file. For example `kustomize edit set image monopole/hello:1=monopole/hello:latest` will update the image. In this case, we need to check that we don't commit the changes.
2. Use kustomize [components](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/components.md). This will require commenting out the component that we want to opt-out.
3. Use what @jeniawhite proposed. In that case, I think it would be better to also update [this](https://github.com/elastic/cloudbeat/blob/main/deploy/kustomize/overlays/cloudbeat-vanilla/kustomization.yml#L7) base to use the files from this pr.

olegsu (approved):

dev-docs/Development.md
```diff
@@ -93,15 +115,17 @@ For more information, see our [testing docs](/tests/README.md)
 Cloudbeat uses [`mockery`](https://github.com/vektra/mockery) as its mocking test framework.
 `Mockery` provides an easy way to generate mocks for golang interfaces.
 
-Some tests use the new [expecter]((https://github.com/vektra/mockery#expecter-interfaces)) interface the library provides.
+Some tests use the new [expecter](<(https://github.com/vektra/mockery#expecter-interfaces)>) interface the library provides.
```
oren-zohar, line 118: I think this change breaks the link

JordanSh (approved):
Worked like a charm 👍

oren-zohar (approved):

PR #649, opened 2023-01-05
Modify configuration to be aligned with new structure

uri-weisman:
**What does this PR do?**
1. Update policy version to latest - v1.2.9
2. Modify config serialization to fit new fleet configuration.

**Related Issues**
- https://github.com/elastic/integrations/pull/4931

**Checklist**


config/config.go
```diff
@@ -47,13 +47,17 @@ type Fetcher struct {
...
+	Fetchers        []*config.C             `config:"fetchers"`
+	KubeConfig      string                  `config:"kube_config"`
+	Period          time.Duration           `config:"period"`
+	Processors      processors.PluginConfig `config:"processors"`
+	BundlePath      string                  `config:"bundle_path"`
+}
+
+type BenchmarkConfig struct {
+	ID        string        `config:"benchmark"`
+	AWSConfig aws.ConfigAWS `config:"aws.credentials"`
```
amirbenun, line 60: I don't see why `AWSConfig` should be a part of `BenchmarkConfig`.
For example, vulnerability management is not a benchmark but we will want to reuse the AWS authentication logic there.

uri-weisman: I see.
We can keep this structure and rename `BenchmarkConfig` to something more general to include vulnerability management.

deploy/aws/cloudbeat-aws.yml
```diff
@@ -2,9 +2,11 @@ cloudbeat:
   config:
     v1:
       benchmark: cis_aws
+      aws:
+        credentials:
+          access_key_id: ${AWS_ACCESS_KEY_ID:no_access_key_was_set}
```
jeniawhite, line 7: Why do we have these values instead of nil or empty values that will `omitempty` the keys in unmarshal flow?
`no_access_key_was_set`
`no_secret_was_set`

I think this creates a more complex flow where we populate fields with mock values just to fail later on.
I would expect the behavior to be different where we do not populate the fields and just fail on validation that the fields were non-existent or continue the flow without keys in case of anonymous access (which this change blocks since it mocks something invalid).

Obviously, this would require changes to the `config.go` and potentially changes to the `validator` in case we do not want to rely on `config.go`, but as of today all of our validations happen in the `config.go` layer.

config/config.go
amirbenun: Besides the integration already sending the AWS credentials as part of the benchmark config, is there a real reason to enforce that coupling? In the future we will have also `GCPConfig` and `AzureConfig` are they all going to be part of an object that also contains the benchmark details?
I think we should have a `CloudConfig` object, that for now holds only aws credentials, while benchmark config should be a separate object.

deploy/aws/cloudbeat-aws.yml
uri-weisman: No real reason just wanted to be aligned with the EKS config file.
I'll revert it to an empty string.

deploy/kustomize/overlays/cloudbeat-eks/cloudbeat.yml
```diff
@@ -2,9 +2,11 @@ cloudbeat:
   config:
     v1:
       benchmark: cis_eks
+      aws:
+        credentials:
+          access_key_id: ${AWS_ACCESS_KEY_ID:no_access_key_was_set}
+          secret_access_key: ${AWS_SECRET_ACCESS_KEY:no_secret_was_set}
```
gurevichdmitry, line 8: @uri-weisman, the file `tests/deploy/cloudbeat-eks-pytest.yaml` should be also updated with the code above.

gurevichdmitry (commented):
@uri-weisman, `deploy` dir files approved. Please also update `cloudbeat-eks-pytest.yaml` in tests dir.

deploy/kustomize/overlays/cloudbeat-eks/cloudbeat.yml
uri-weisman: @gurevichdmitry - thanks updated.

gurevichdmitry (approved):
@uri-weisman, deploy files approved.

amirbenun (approved):

PR #654, opened 2023-01-09
Add Multi region support for CSPM

uri-weisman:
**What does this PR do?**
For regional AWS services, add support for multi-region data collection (only from enabled regions).

**Checklist**


resources/fetchers/network_fetcher.go
```diff
@@ -48,7 +48,7 @@ type NetworkResource struct {
 func (f NetworkFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
 	f.log.Debug("Starting NetworkFetcher.Fetch")
 
-	nacl, err := f.provider.DescribeNeworkAcl(ctx)
+	nacl, err := f.provider.DescribeNetworkAcl(ctx)
```
uri-weisman, line 51: fixed typo

config/config.go
```diff
@@ -110,10 +105,6 @@ func Datastream(namespace string, indexPrefix string) string {
 	return indexPrefix + "-" + namespace
 }
 
-type AwsConfigProvider interface {
```
uri-weisman, line 113: Removed - seems to be a duplication

resources/providers/awslib/aws.go
```diff
@@ -18,20 +18,85 @@
...
+	Clients map[string]T
+}
+
+// CreateMultiRegionClients is a utility function that is used to create a map of client instances of a given type T for multiple regions.
+func CreateMultiRegionClients[T any](client AWSCommonUtil, cfg awssdk.Config, factory func(cfg awssdk.Config) T, log *logp.Logger) *MultiRegionWrapper[T] {
+	var clientsMap = make(map[string]T, 0)
+	for _, region := range getRegions(client, log) {
+		cfg.Region = region
+		client := factory(cfg)
+		clientsMap[region] = client
```
amirbenun, line 64: As we talked about on Thursday I think this map should hold our providers instead of the raw AWS clients. It will give us more flexibility in the future. (There's nothing wrong with the function implementation just the usage)

resources/providers/awslib/aws_test.go
```diff
@@ -0,0 +1,95 @@
...
+			args: args{
+				client: func() AWSCommonUtil {
+					m := &MockAWSCommonUtil{}
+					m.On("DescribeRegions", mock.Anything, mock.Anything).Return(successfulOutput, nil)
+					return m
+				},
+				cfg: awssdk.Config{},
+				log: logp.NewLogger("aws-test"),
+			},
+			want: map[string]*ec2sdk.Client{DefaultRegion: ec2Client, "eu-west-1": ec2Client},
```
amirbenun, line 79: The end goal is to validate that the client was generated with the right parameters, in that case I would change a few things:
1. Set `want` map to have a type string to the config object `want map[string]awssdk.Config`
2. Your factory should just return the config
3. Then you can assert that you got the right config object with the correct region

resources/providers/awslib/aws.go
```diff
@@ -18,20 +18,85 @@
...
+}
+
+// GetRegions will initialize the singleton instance and perform the API request to retrieve the regions list only once, even if the function is called multiple times.
+// Subsequent calls to the function will return the stored regions list without making another API request.
+// In case of a failure the function returns the default region.
+func getRegions(client AWSCommonUtil, log *logp.Logger) []string {
+	log.Debug("GetRegions starting...")
+
+	var initErr error
+	once.Do(func() {
```
amirbenun, line 81: Race condition.
In the case of two parallel calls to `getRegions`, the first one will get into the `once.Do` and before it finishes, the other will skip it and return nil.

resources/providers/awslib/config_provider.go
```diff
@@ -28,21 +28,25 @@ type ConfigProvider struct {
 	MetadataProvider MetadataProvider
 }
 type ConfigProviderAPI interface {
-	InitializeAWSConfig(ctx context.Context, cfg aws.ConfigAWS, log *logp.Logger) (awssdk.Config, error)
+	InitializeAWSConfig(ctx context.Context, cfg aws.ConfigAWS, log *logp.Logger, useDefaultRegion bool) (awssdk.Config, error)
 }
 
-func (p ConfigProvider) InitializeAWSConfig(ctx context.Context, cfg aws.ConfigAWS, log *logp.Logger) (awssdk.Config, error) {
+func (p ConfigProvider) InitializeAWSConfig(ctx context.Context, cfg aws.ConfigAWS, log *logp.Logger, useDefaultRegion bool) (awssdk.Config, error) {
```
amirbenun, line 34: What is the use case of `useDefaultRegion`? It feels like it's suppose to be a part of the same mechanism of `CreateMultiRegionClients`.
A good practice we can use here is always using the multi-region mapping and the mapping can have a single region listed.

resources/providers/awslib/ec2/provider.go
```diff
@@ -38,44 +38,50 @@ type Client interface {
...
-		output, err := p.client.DescribeNetworkAcls(ctx, &input)
-		if err != nil {
-			return nil, err
-		}
-		allAcls = append(allAcls, output.NetworkAcls...)
-		if output.NextToken == nil {
-			break
+func (p *Provider) DescribeNetworkAcl(ctx context.Context) ([]awslib.AwsResource, error) {
+	var allAcls []types.NetworkAcl
+	for _, client := range p.Clients {
```
amirbenun, line 43: This function involves too much logic and which causes confusion.
The old implementation wasn't "region aware" which is good.
The cross-region logic ideally should not be a part of that function, instead, we should implement it in a different layer above that function.

amirbenun (changes requested):
We should extend the usage of the `MultiRegionWrapper`

resources/providers/awslib/config_provider.go
```diff
@@ -21,28 +21,26 @@ import (
...
 	awsConfig, err := aws.InitializeAWSConfig(cfg)
 	if err != nil {
 		return awssdk.Config{}, err
 	}
-
 	metadata, err := p.MetadataProvider.GetMetadata(ctx, awsConfig)
 	if err != nil {
-		log.Errorf("MetadataProvider.GetMetadata Error: %v, setting AWSConfig region to default - %s", err, awsConfig.Region)
-		return awsConfig, nil
+		return awssdk.Config{}, err
```
uri-weisman, line 41: reverted to pre cspm changes.

amirbenun: Don't we want to keep the log though?

amirbenun: Also, I think we should return a pointer to `awssdk.Config` and that way we will be able to return `nil` here.

resources/providers/awslib/multi_region.go
```diff
@@ -0,0 +1,135 @@
...
+type DescribeCloudRegions interface {
+	DescribeRegions(ctx context.Context, params *ec2.DescribeRegionsInput, optFns ...func(*ec2.Options)) (*ec2.DescribeRegionsOutput, error)
+}
+
+type MultiRegionWrapper[T any] struct {
+	clients map[string]T
+}
+
+// NewMultiRegionClients is a utility function that is used to create a map of client instances of a given type T for multiple regions.
+func (w *MultiRegionWrapper[T]) NewMultiRegionClients(client DescribeCloudRegions, cfg awssdk.Config, factory func(cfg awssdk.Config) T, log *logp.Logger) *MultiRegionWrapper[T] {
```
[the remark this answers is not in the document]
uri-weisman: Agreed, that would be best.
Having said that, for the consumer factory (`S3Factory`/`NetworkFactory`) to be testable I have to pass the `CrossRegionUtil` through the factory struct, I have two options to do it.

1. Pass an initialization func to be called later (same as was done with the `IdentityProvider`).
2. Instantiate a struct that implements the `CrossRegionUtil` interface.

The first option is impossible as a generic function can not be passed as an argument, therefore I choose the second.

resources/providers/awslib/multi_region.go
```diff
@@ -0,0 +1,135 @@
...
+	NewMultiRegionClients(client DescribeCloudRegions, cfg awssdk.Config, factory func(cfg awssdk.Config) T, log *logp.Logger) *MultiRegionWrapper[T]
+	Fetch(fetcher func(T) ([]AwsResource, error)) ([]AwsResource, error)
+	GetMultiRegionsClientMap() map[string]T
+}
+
+type DescribeCloudRegions interface {
+	DescribeRegions(ctx context.Context, params *ec2.DescribeRegionsInput, optFns ...func(*ec2.Options)) (*ec2.DescribeRegionsOutput, error)
+}
+
+type MultiRegionWrapper[T any] struct {
```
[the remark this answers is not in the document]
uri-weisman: That's true, related to my previous comment.

resources/providers/awslib/multi_region.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			initErr = fmt.Errorf("failed DescribeRegions: %w", err)
+			once = &sync.Once{} // reset singleton upon error
+			return
+		}
+
+		for _, region := range output.Regions {
+			awsRegions.enabledRegions = append(awsRegions.enabledRegions, *region.RegionName)
+		}
+	})
```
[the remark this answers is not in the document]
uri-weisman: A race condition can happen here as well:
```
	if initErr != nil {
		log.Errorf("Error: %v, init client only for the default region: %s", initErr, DefaultRegion)
		awsRegions.enabledRegions = append(awsRegions.enabledRegions, DefaultRegion)
	}
```
Even though we're initializing a set with the same value so it will have no effect, wdyt?

resources/providers/awslib/config_provider.go
uri-weisman: Regarding keeping the log, there is no real point in logging and returning an error.
This error is propagated all the way to `cloudbeat.go` and will cause the beat to stop.

resources/providers/awslib/multi_region.go
```diff
@@ -0,0 +1,135 @@
...
+
+	if initErr != nil {
+		log.Errorf("Error: %v, init client only for the default region: %s", initErr, DefaultRegion)
+		awsRegions.enabledRegions = append(awsRegions.enabledRegions, DefaultRegion)
```
amirbenun, line 130:
```suggestion
		return []string{DefaultRegion}
```

resources/providers/awslib/multi_region.go
```diff
@@ -0,0 +1,131 @@
...
+var (
+	awsRegions *cachedRegions
+	once       = &sync.Once{}
+)
+
+type cachedRegions struct {
+	enabledRegions []string
+}
+
+type CrossRegionUtil[T any] interface {
```
amirbenun, line 39: Nit: I would rename to `CrossRegionFetcher`

resources/providers/awslib/ec2/cross.go
```diff
@@ -0,0 +1,54 @@
...
+	})
+}
+
+func (c *crossRegionElasticCompute) DescribeSecurityGroups(ctx context.Context) ([]awslib.AwsResource, error) {
+	return c.crossRegionUtil.Fetch(func(ec ElasticCompute) ([]awslib.AwsResource, error) {
+		return ec.DescribeSecurityGroups(ctx)
+	})
+}
+
+func NewCrossEC2Provider(log *logp.Logger, awsAccountID string, cfg aws.Config, util awslib.CrossRegionFactory[ElasticCompute]) ElasticCompute {
```
amirbenun, line 45: Nit: rename `util` -> `factory`

resources/providers/awslib/ec2/cross.go
```diff
@@ -0,0 +1,54 @@
...
+	"context"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/elastic/cloudbeat/resources/providers/awslib"
+	"github.com/elastic/elastic-agent-libs/logp"
+)
+
+type crossRegionElasticCompute struct {
+	crossRegionUtil awslib.CrossRegionUtil[ElasticCompute]
```
amirbenun, line 30: Nit: rename `crossRegionUtil` -> `crossRegionFetcher`

resources/providers/awslib/s3/provider.go
```diff
@@ -22,64 +22,76 @@ import (
...
 	"github.com/aws/smithy-go"
 	"github.com/elastic/cloudbeat/resources/fetching"
 	"github.com/elastic/cloudbeat/resources/providers/awslib"
 	"github.com/elastic/elastic-agent-libs/logp"
-	"gotest.tools/gotestsum/log"
 )
 
-func NewProvider(cfg aws.Config, log *logp.Logger) *Provider {
-	client := s3Client.NewFromConfig(cfg)
+func NewProvider(cfg aws.Config, log *logp.Logger, util awslib.CrossRegionFactory[Client]) *Provider {
```
amirbenun, line 34: Nit: rename `util` -> `factory`

resources/fetchers/network_factory.go
```diff
@@ -34,14 +34,14 @@ import (
...
-		IdentityProvider:  awslib.GetIdentityClient,
+		CrossRegionUtil:  &awslib.MultiRegionClientFactory[ec2.ElasticCompute]{},
+		IdentityProvider: awslib.GetIdentityClient,
 	})
 }
 
 type EC2NetworkFactory struct {
-	AwsConfigProvider config.AwsConfigProvider
-	IdentityProvider  func(cfg awssdk.Config) awslib.IdentityProviderGetter
+	CrossRegionUtil  awslib.CrossRegionFactory[ec2.ElasticCompute]
```
amirbenun, line 43: Nit: rename `CrossRegionUtil` -> `CrossRegionFactory`

resources/fetchers/s3_factory.go
```diff
@@ -32,12 +31,12 @@ import (
...
 func init() {
 	fetchersManager.Factories.RegisterFactory(fetching.S3Type, &S3Factory{
-		AwsConfigProvider: awslib.ConfigProvider{MetadataProvider: awslib.Ec2MetadataProvider{}},
+		CrossRegionUtil: &awslib.MultiRegionClientFactory[s3.Client]{},
 	})
 }
 
 type S3Factory struct {
-	AwsConfigProvider config.AwsConfigProvider
+	CrossRegionUtil awslib.CrossRegionFactory[s3.Client]
```
amirbenun, line 39: Nit: rename `CrossRegionUtil` -> `CrossRegionFactory`

resources/providers/awslib/ec2/cross.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/elastic/cloudbeat/resources/providers/awslib"
+	"github.com/elastic/elastic-agent-libs/logp"
+)
+
+type crossRegionElasticCompute struct {
```
amirbenun, line 29: Nit: rename `crossRegionElasticCompute` -> `crossRegionProvider`

resources/providers/awslib/ec2/cross.go
amirbenun, line 45: Nit: rename `NewCrossEC2Provider` -> `NewCrossRegionProvider` (As the package named `ec2`)

resources/providers/awslib/cross_region_util_mock.go
```diff
@@ -0,0 +1,81 @@
...
+// under the License.
+
+// Code generated by mockery v2.15.0. DO NOT EDIT.
+
+package awslib
+
+import mock "github.com/stretchr/testify/mock"
+
+// MockCrossRegionUtil is an autogenerated mock type for the CrossRegionUtil type
+type MockCrossRegionUtil[T interface{}] struct {
```
amirbenun, line 25: Nit: rename struct and file

amirbenun (approved):
Nice! I would rename the `util` to a more concrete name (factory/fetcher depends on the usage)

PR #680, opened 2023-01-27
Mockery

olegsu:
### Summary of your changes
This will replace and add all the missing mocks using `mockery`. 
It will use the standard mockery name pattern to keep predictability.
Also, it will crash the CI process if any mock is missing or the interface was updated but not the mock.

**Important**
No changes in codebase

### Screenshot/Data

image

### Checklist


amirbenun (approved):
Nice!

justfile
```diff
@@ -112,6 +112,10 @@ expose-ports:
   CLOUDBEAT_POD=$( kubectl get pods -o=name -n kube-system | grep -m 1 "cloudbeat" ) && \
   kubectl port-forward $CLOUDBEAT_POD -n kube-system 40000:40000 8080:8080
 
+generate-mocks:
```
oren-zohar, line 115: Nice! can you please document this? would be nice to mention this [here](https://github.com/elastic/cloudbeat/blob/main/dev-docs/Development.md#mocking)

scripts/make/validate-mocks.sh
```diff
@@ -0,0 +1,25 @@
...
+if [ $diff_count -eq 0 ] && [ $untracked_count -eq 0 ]; then
+    exit 0
+fi
+
+
+echo "There are missing mocks for interfaces, please run \"$mock\" to update all mocks"
+echo "Mock files:"
+eval $diff
+eval $untracked
+exit 1
```
oren-zohar, line 25: I wonder if running the just command and then adding the `run: git diff --exit-code` step in the workflow will achieve the same. I want to avoid adding bash scripts to our workflows - for now, I don't see any security concern in that specific workflow since we don't use secrets, it's possible it will happen in the future and this will be an easy exploit. let's talk about this offline

oren-zohar (changes requested):

oren-zohar (approved):

PR #705, opened 2023-02-06
Aws cis 2 2 1

olegsu:
### Summary of your changes
add aws cis 2.2.1

### Related Issues
Related https://github.com/elastic/csp-security-policies/issues/145

### Checklist

### DIFF
--- a/deploy/aws/cloudbeat-aws.yml
+++ b/deploy/aws/cloudbeat-aws.yml
@@ -10,11 +10,11 @@ cloudbeat:
   # Defines how often an event is sent to the output
   period: 30s
   fetchers:
-    - name: aws-iam
-    - name: aws-s3
+    # - name: aws-iam
+    # - name: aws-s3
     - name: aws-ec2-network
-    - name: aws-trail
-    - name: aws-monitoring
+    # - name: aws-trail
+    # - name: aws-monitoring
   evaluator:
     decision_logs: false
 # =================================== Kibana ===================================
--- a/deploy/aws/cloudbeat-aws.yml
+++ b/deploy/aws/cloudbeat-aws.yml
@@ -10,11 +10,11 @@ cloudbeat:
   # Defines how often an event is sent to the output
   period: 30s
   fetchers:
-    # - name: aws-iam
-    # - name: aws-s3
+    - name: aws-iam
+    - name: aws-s3
     - name: aws-ec2-network
-    # - name: aws-trail
-    # - name: aws-monitoring
+    - name: aws-trail
+    - name: aws-monitoring
   evaluator:
     decision_logs: false
 # =================================== Kibana ===================================
--- a/resources/providers/awslib/ec2/ebs_encryption.go
+++ b/resources/providers/awslib/ec2/ebs_encryption.go
@@ -18,27 +18,24 @@
 package ec2
 
 import (
-	"context"
+	"fmt"
 
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/service/ec2"
-	"github.com/elastic/cloudbeat/resources/providers/awslib"
-	"github.com/elastic/elastic-agent-libs/logp"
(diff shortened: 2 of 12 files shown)


deploy/aws/cloudbeat-aws.yml
```diff
@@ -10,11 +10,11 @@ cloudbeat:
   # Defines how often an event is sent to the output
   period: 30s
   fetchers:
-    - name: aws-iam
-    - name: aws-s3
+    # - name: aws-iam
```
uri-weisman, line 13: plz revert

resources/providers/awslib/ec2/ebs_encryption.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/elastic/cloudbeat/resources/fetching"
+)
+
+type EBSEncryption struct {
+	Enabled bool
+	region  string
+}
+
+func (e EBSEncryption) GetResourceArn() string {
+	return fmt.Sprintf("ebs-encryption-by-default-%s", e.region)
```
uri-weisman, line 32: Multi AWS accounts will lead to "collision".
Let's add the account id.

```suggestion
	return fmt.Sprintf("default-ebs-encryption-%s-%s", e.account, e.region)
```

resources/providers/awslib/ec2_metadata_provider.go
```diff
@@ -42,5 +42,5 @@ func (provider Ec2MetadataProvider) GetMetadata(ctx context.Context, cfg aws.Con
 		return ec2imds.GetInstanceIdentityDocumentOutput{}.InstanceIdentityDocument, err
 	}
 
-	return identityDocument.InstanceIdentityDocument, err
+	return identityDocument.InstanceIdentityDocument, nil
```
uri-weisman, line 45: Is it related to this PR?

resources/providers/awslib/multi_region.go
```diff
@@ -92,15 +93,50 @@ func MultiRegionFetch[T any, K any](ctx context.Context, set map[string]T, fetch
...
 	wg.Wait()
 	return crossRegionResources, err
 }
 
+// shouldDrop checks the target type and return true if
+// type is pointer -> the pointer is nil
+// type is struct -> the type zero value
+// type is slice -> len(slice) == 0
+// and false otherwise
+func shouldDrop(t interface{}) bool {
```
uri-weisman, line 117: Cool, why we needed this now?

uri-weisman (changes requested):

resources/providers/awslib/multi_region.go
olegsu: The fetcher can return a slice, struct, or pointer.
When a pointer is nil it will add to the final slice and at the end will have a slice of nils.
To avoid this I thought about checking this.
When we use generics we cant just compare to nil, or check zero value.
So we need to use reflection.

resources/providers/awslib/ec2/ebs_encryption.go
```diff
@@ -0,0 +1,41 @@
...
+package ec2
+
+import (
+	"fmt"
+
+	"github.com/elastic/cloudbeat/resources/fetching"
+)
+
+type EBSEncryption struct {
+	Enabled bool
```
uri-weisman, line 27:
```suggestion
	Enabled bool `json: enabled`
```

resources/providers/awslib/ec2_metadata_provider.go
olegsu: No.
It does not make sense to return the `err` that is nil, instead, return nil.
Note that this error was checked above.

resources/providers/awslib/ec2/ebs_encryption.go
olegsu: not sure why

uri-weisman: Not a must but it will be cleaner to address this field in rego like:
`input.resource.enabled`
rather than
`input.resource.Enabled`

PR #713, opened 2023-02-07
refactor: align providers to be multi regional

olegsu:
### Summary of your changes
refactor securityhub, ec2 and rds providers to be multi-regional like the other aws providers we have

### Checklist

### DIFF
--- a/resources/fetchers/monitoring_factory.go
+++ b/resources/fetchers/monitoring_factory.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 
 	awssdk "github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/service/ec2"
 	"github.com/elastic/beats/v7/x-pack/libbeat/common/aws"
 	"github.com/elastic/cloudbeat/resources/fetchersManager"
 	"github.com/elastic/cloudbeat/resources/providers/aws_cis/monitoring"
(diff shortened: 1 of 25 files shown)


uri-weisman (changes requested):
I'm not sure about shifting the multi-region logic to the providers, it makes sense to me to keep the providers region-agnostic and apply it inside the consumer (fetchers).
If we decide to go for it, a compromise can be to create two public functions inside the provider:

1. A function that holds the logic and can be consumed with a single client.
2. Another one that uses the multi-region utility to iterate all regions.

Wdyt?

olegsu:
> I'm not sure about shifting the multi-region logic to the providers, it makes sense to me to keep the providers region-agnostic and apply it inside the consumer (fetchers). If we decide to go for it, a compromise can be to create two public functions inside the provider:
> 


Thank you, @uri-weisman 
When I looked on our aws providers we have I see that half of them are already multi-regional
* [cloudtrail](https://github.com/elastic/cloudbeat/blob/main/resources/providers/awslib/cloudtrail/provider.go)
* [cloudwatch](https://github.com/elastic/cloudbeat/blob/main/resources/providers/awslib/cloudwatch/provider.go)
* [cloudwatchlogs](https://github.com/elastic/cloudbeat/blob/main/resources/providers/awslib/cloudwatch/logs/provider.go)
* [sns](https://github.com/elastic/cloudbeat/blob/main/resources/providers/awslib/sns/provider.go)

And region specific are
* securityhub
* rds
* ec2
* s3 ( which i didn't include in this pr, and i should)

And we have a non-regional provider and other providers that are not used in aws cis
* iam
* elb
* ecr

So I think consolidation is required here. Either all the providers are region aware or the fetcher is region aware.

> 1. A function that holds the logic and can be consumed with a single client.
> 2. Another one that uses the multi-region utility to iterate all regions.

I agree, but I think that when it will be required we can refactor to support it.

uri-weisman (approved):

PR #721, opened 2023-02-09
CIS AWS Add a config provider

uri-weisman:
### Summary of your changes
Add a config provider to address rule 3.5 of the CIS AWS benchmark

### Related Issues
- https://github.com/elastic/csp-security-policies/issues/188 

### Checklist


resources/fetchers/logging_fetcher.go
```diff
@@ -72,3 +89,18 @@ func (r LoggingResource) GetMetadata() (fetching.ResourceMetadata, error) {
 	}, nil
 }
 func (r LoggingResource) GetElasticCommonData() any { return nil }
+
+func (c ConfigResource) GetMetadata() (fetching.ResourceMetadata, error) {
+	return fetching.ResourceMetadata{
+		ID:      c.GetResourceArn(),
+		Type:    fetching.CloudConfig,
```
olegsu, line 96: Why not add it to the logging type? it is in the same cis section so it make sense some how

resources/providers/awslib/configservice/configservice.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (c Config) GetResourceArn() string {
+	return fmt.Sprintf("config-service-%s-%s", c.region, c.accountId)
+}
+
+func (c Config) GetResourceName() string {
+	if c.Recorders == nil || len(c.Recorders) == 0 || c.Recorders[0].Name == nil {
+		return ""
+	}
+	return *c.Recorders[0].Name
```
olegsu, line 80: Does this idiomatic?

resources/utils/testhelper/helper.go
```diff
@@ -26,3 +26,12 @@ func CollectResources[T any](ch chan T) []T {
 
 	return results
 }
+
+func CreateMockClients[T any](client T, regions []string) map[string]T {
```
olegsu, line 30: 🔥

olegsu (changes requested):

resources/fetchers/logging_fetcher.go
uri-weisman: IMO, The type field should represent a group of resources and be cloud/benchmark agnostic as much as possible.
`Logging` refers to a benchmark section and should be used only if we had to create a hybrid type such as the `Monitoring` type.

resources/providers/awslib/configservice/configservice.go
uri-weisman: Anything specific u would change?

uri-weisman: lets go with configservice-<REGION>

olegsu (approved):

Issue #622, opened 2023-01-02
Terraform refactoring


olegsu:
**Motivation**
At the moment we have one root module with 4 sub-modules that use it
As our Terraform module grows, we need a better structure so that the user is more clear.

We should have logical modules that are doing everything that it needs
1. Elastic-Stack should include the dashboards built-in
2. EKS should be creating the package policy
3. Ec2 should be creating the package policy

DOD
- [ ] Spin up all the modules in one command
- [ ] Be able to spin up on Elastic-stack module
- [ ] All the resources have some common property ( tag, label, name prefix, etc ), we can use [random_pet](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/pet) to generate human-readable name
- [ ] Don't assume the existence of IAM policies but instead create them
- [ ] Lock down the versions of all imported modules so things won't break - specifically the apm tf module we're using
- [x] Set up ECR image pull for the vanilla cluster (See [comment](https://github.com/elastic/security-team/issues/5689#issuecomment-1376018205) for details)

**Related tasks/epics**
Tech debt ticket for https://github.com/elastic/security-team/issues/5656

eyalkraft:
> - [ ] All the resources have some common property ( tag, label, name prefix, etc ), we can use [random_pet](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/pet) to generate human-readable name

I think currently only the resources related to the vanilla deployment lack the common prefix. random_pet looks cool.

Also adding these tasks from my last list that weren't competed yet:

> 2. Currently the script assumes the existence of the IAM policies
    ```
    Developers_eks = "arn:aws:iam::704479110758:policy/Developers_eks"
    EKS_ReadAccess = "arn:aws:iam::704479110758:policy/EKS_ReadAccess"
    ```
    A more proper way would be to have the tf script provision the required policy itself.

> 6. Lock down the versions of all imported modules so things won't break

ofiriro3:
I think it would be nice also to add an owner tag, so we can easily track our EKS clusters/Ec2 machines

uri-weisman:
I've encountered several quotas limits while creating a new environment.
Would be nice if we'll use existing resources if possible (NAT Gateway, VPC).

ofiriro3:
Another gap that I think we should address is that the EC2 deployment is strongly dependent on an EKS deployment I.E you cannot currently deploy and EC2 machine without deploying an EKS cluster

olegsu:
[...]

PR #746, opened 2023-02-14
Weekly environment - Vanilla POC

ofiriro3:
### Summary of your changes
This is the first item in the development of the Daily-environment [epic](https://github.com/elastic/security-team/issues/4475).

In this task, we created a job that deploys a vanilla KSPM integration on an Elastic cloud environment.
The job currently uses a pre-exited EC2 machine and it deploys an agent there (no new EC2 machines will be created).

The expected behavior is:
1. A new ELK environment is created with each execution.
2. A new Vanilla integration is installed on this environment.
3. The same EC2 machine whose credentials exist in the GitHub secret will deploy the agent.
4. A slack message will be sent to `ofir-personal-channel` with the relevant credentials for this environment.

**How to check the daily job**
Currently, the flow is deployed using a dispatch workflow, meaning - you will need to apply it using GitHub actions.
You can apply the flow from my fork, using [actions](https://github.com/ofiriro3/cloudbeat/actions/workflows/weekly-enviroment.yml).
Please set the branch to `vanilla_poc` before running the dispatch workflow.
It should look like this:
image

The flow will deploy the integration to https://cloud.elastic.co/home, in the next PRs we will change it to a staging environment and replace the latest image with the SNAPSHOT version.

### Related Issues
Related https://github.com/elastic/security-team/issues/5903

### Checklist


.github/workflows/weekly-enviroment.yml
```diff
@@ -0,0 +1,141 @@
...
+      run:
+        working-directory: ${{ env.WORKING_DIR }}
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v2
+        with:
+          terraform_version: 1.3.5
```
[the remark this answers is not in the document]
oren-zohar: https://github.com/elastic/cloudbeat/blob/41cd8848d17f30f90380e20ba3e06649cd70214a/.github/workflows/cloudbeat-ci.yml#L27

.github/workflows/weekly-enviroment.yml
```diff
@@ -0,0 +1,141 @@
...
+        run: terraform fmt -check
+        continue-on-error: true
+
+      - name: Terraform Init
+        id: init
+        run: terraform init -no-color
+
+      - name: Terraform Validate
+        id: validate
+        run: terraform validate -no-color
```
oren-zohar, line 59: Nice! but I think the linting should be in a separate step, as it is not part of the deployment logic, can run in parallel, and on failure, easier to understand what part failed and when.

.github/workflows/weekly-enviroment.yml
```diff
@@ -0,0 +1,141 @@
...
+
+      - name: Deploy Elastic Cloud
+        id: apply
+        run: terraform apply --auto-approve
+
+      - name: Set terraform output as env variable
+        id: kibana_url
```
oren-zohar, line 66: not sure why you add the `id` optional field in all of the steps, if you're not using it

.github/workflows/weekly-enviroment.yml
```diff
@@ -0,0 +1,141 @@
...
+        id: install_vanilla_integration
+        working-directory: ${{ env.SCRIPTS_DIR }}
+        run: |
+          ./install-kspm-vanilla-integration.sh ${{ env.KIBANA_URL }} ${{ env.ELASTICSEARCH_PASSWORD }}
+
+      - name: Deploy agent on EC2
+        id: deploy_agent_on_ec2
+        working-directory: ${{ env.SCRIPTS_DIR }}
+        run: |
+          echo -e "${{ secrets.WEEKLY_ENVIRONMENT_EC2_SECRET }}" > weekly-key.pem
```
oren-zohar:
```suggestion
          echo -e "${{ secrets.WEEKLY_ENVIRONMENT_EC2_PRIVATE_KEY }}" > weekly-key.pem
```

oren-zohar (changes requested):
Only went over the workflow

olegsu:
Do we plan to consolidate it to the other terraform module at some point? what are the requirementes?
If not, why?

.github/workflows/weekly-enviroment.yml
```diff
@@ -0,0 +1,141 @@
...
+          - WARN
+          - ERROR
+
+env:
+  WORKING_DIR: deploy/weekly-environment
+  SCRIPTS_DIR: deploy/weekly-environment/scripts/benchmarks/kspm_vanilla
+  TF_VAR_ec_api_key: ${{ secrets.WEEKLY_ENVIRONMENT_EC_API_KEY }}
+  TF_VAR_environment: ${{ github.event.inputs.logLevel }}
+  TF_LOG: ${{ github.event.inputs.logLevel }}
+  TF_VAR: 8.6.1
```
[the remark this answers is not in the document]
ofiriro3: Yes, we would like the weekly environment to load a specific version every time it runs.

We would like it to run the latest SNAPSHOT version available. But we would like to change it manually every time since we would like to update the integration and the relevant rules.

.github/workflows/weekly-enviroment.yml
```diff
@@ -0,0 +1,141 @@
...
+
+jobs:
+  terraform:
+    name: Deploy KSPM/CSPM cloud environment
```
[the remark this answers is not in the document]
ofiriro3: I updated it to KSPM 👍

.github/workflows/weekly-enviroment.yml
ofiriro3: I removed it 👍

.github/workflows/weekly-enviroment.yml
ofiriro3: Yeah, it is an optional field.

There are approaches that consider adding them anyway as best practice. 

Anyway, I removed them 👍

.github/workflows/weekly-enviroment.yml
ofiriro3: This one was already updated after out call 👍

ofiriro3:
> Do we plan to consolidate it to the other terraform module at some point? what are the requirements?

Currently, the weekly environment job uses terraform to create a new Elastic cloud environment and does not require any other terraform module from our solutions. Other actions are done by ssh to the relevant EC2 machine and applying the specific operations there.

For the first deliverable solution, we decided that the job won't create a new EC2 machine and its infrastructure, and instead will use an existing machine. This assumption was made to enable us to move fast with a reliable solution and not be dependent on the rest of the terraform modules. So in this case, I thought that doing the ssh will be done much easier in a bash script.

As I see it, in the future we will might decide to consolidate it with other terraform modules.
For example, we will might decide to create an EC2 machine if such does not exist as this ticket suggests - https://github.com/elastic/security-team/issues/5788, but before doing so, we will have to reconsider some of the dependencies we currently have:
1. EC2 creation + agent deployment - we would like to deploy an agent on every job execution, but not to deploy a new EC2 every time.
2. EKS creation + agent deployment - same as 1.
3. API module - I guess that this entire module can be refactored.

.github/workflows/weekly-enviroment.yml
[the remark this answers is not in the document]
oren-zohar: I'm not sure I understand the logic behind it. Why not use the branch version as a default value?

.github/workflows/weekly-enviroment.yml
oren-zohar: I still think it's valuable to have it - just in a separate step, maybe in one of the other workflows where we run linters?

olegsu:
Thank you for a detailed explanation, @ofiriro3.
> Currently, the weekly environment job uses terraform to create a new Elastic cloud environment and does not require any other terraform modules from our solutions. Other actions are done by ssh to the relevant EC2 machine and applying the specific operations there.

Imo, this is not ideal to have multiple scripts doing practically the same thing.  I think a consolidation should be added to https://github.com/elastic/cloudbeat/issues/622m, not to DOD but to something that we want to track and achieve when possible. Otherwise, it would be hard to maintain two scripts.

.github/workflows/weekly-enviroment.yml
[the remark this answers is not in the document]
ofiriro3: Tested it, good advice! I removed the Terraform setup step and added a `Init Hermit` step

ofiriro3:
> https://github.com/elastic/cloudbeat/issues/622, not to DOD but to something that we want to track and achieve when possible. Otherwise, it would be hard to maintain two scripts.

Done. I commented on the ticket

.github/workflows/weekly-enviroment.yml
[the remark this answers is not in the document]
ofiriro3: So just to verify that I understood you, you meant for example if the Cloudbeat version is `8.8.0`, so deploy `8.8.0-SNAPSHOT`?

So that is a great idea, but for the matter of this PR, we wanted to start with a POC that will use a hard-coded version.
In the future we would like even to support multiple versions:
- https://github.com/elastic/security-team/issues/6078

But for this POC, we assumed that it uses a hard-coded version 8.6.1 as the current Terraform default value suggests.

.github/workflows/unit-test.yml
```diff
@@ -153,3 +154,16 @@ jobs:
...
+    name: terraform-lint
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+
+      - name: Init Hermit
+        run: ./bin/hermit env -r >> $GITHUB_ENV
+
+      - name: Terraform fmt
```
olegsu, line 168: Lets have it atm only on the weekly env dir

.github/workflows/weekly-enviroment.yml
```diff
@@ -0,0 +1,126 @@
...
+          - TRACE
+          - DEBUG
+          - INFO
+          - WARN
+          - ERROR
+
+env:
+  WORKING_DIR: deploy/weekly-environment
+  SCRIPTS_DIR: deploy/weekly-environment/scripts/benchmarks/kspm_vanilla
+  TF_VAR_ec_api_key: ${{ secrets.WEEKLY_ENVIRONMENT_EC_API_KEY }}
```
olegsu, line 26: Who is the user behind this key?

olegsu (commented):
Looks good!
What happens if the machine does not exist?
Does the slack message still goes to `ofir-personal-channel` ?

ofiriro3:
> Looks good!
What happens if the machine does not exist?
Does the slack message still goes to ofir-personal-channel ?

If the machine does not exist, no message is sent to the slack channel, but the job fails.

.github/workflows/unit-test.yml
ofiriro3: I don't mind changing it, but why we wouldn't want it to run on the entire solution? 

The CI passes, and it will keep all the terraform in format.

.github/workflows/weekly-enviroment.yml
ofiriro3: Currently its my user, we have two tickets related:

1. Short term solution - https://github.com/elastic/security-team/issues/6017, add the entire team to have permission on my account.

2. Long term solution - Create a team account - https://github.com/elastic/security-team/issues/5921

olegsu (approved):

oren-zohar (approved):

PR #748, opened 2023-02-15
Fix concurrency

olegsu:
### Summary of your changes

<details><summary>refactor: replace time.After with time.NewTimer (289cb25)</summary> 

`time.After` is using `time.NewTime` https://pkg.go.dev/time#example-Time.After.
by using directly `time.NewTime` we can call the `Data.fetchIteration`
in one place.
The zero interval will make sure to trigger it immediatly and after that
we reset it to the input interval
 </details>
<details><summary>fix: concurrency in `fetchersManager.Data` (33c0023)</summary> 

The `Data` used to use `sync.WaitGroup` in two usecases:
1. Waiting for all the fetchers to complete the work
2. When `Data.Stop` is called to make sure that it also waits for all
   the fetchers to complete

Those two usecase used to share the same instance of `sync.WaitGroup`
from different go routines.

By removing the `wg.Wait` from `Data.Stop` the race was removed.

To make sure aligment, the changes are:
1. Remove `Data.Stop` to allow it to be called twice. So `Data.Run` will
   return `Stop` funciton is a wrapper around `Data.stopData` and can
ensure that it will be executed only once (the pattern is similar to
ctx, cancel := context.WithTimeout)
2. Adding `Data.stopNotice` propery as channel to make sure that once
   the stop funciton is called, no more fetchers will be executed.
3. Running `Data.fetchIterantion` as separate go routine will make sure
   that it will not block. So that the `Data.fetchAndSleep` can still
wait receive data from the channels
4. Stop function will get now grace period to wait before the fetchers
   are stopped.
5. Adding cancelable context to `Data.fetchAndSleep` and canceling it when
the stop channel is closed will get the same result.
 </details>
<details><summary>feat: test cloudbeat also with -race flag (9f1733f)</summary> 

1. Add new step to UnitTest workflow with race detection flag in the `go
test` command.
2. Increase the overall timeout of the CI as the new step adding
   requires more time.
 </details>
<details><summary>fix: concurrency in `awslib.MultiRegionFetch` (3e666e6)</summary> 

The `awslib.MultiRegionFetch` function used to share the `error` from
multiple go routines.
Every go routine that called the `fetcher` function was able to update
this variable.
 </details>

### Checklist


resources/fetchersManager/data.go
```diff
@@ -70,8 +72,12 @@ func (d *Data) fetchAndSleep(ctx context.Context) {
...
 			return
-		case <-time.After(d.interval):
-			d.fetchIteration(ctx)
+		case <-timer.C:
+			// upadte the interval
+			timer.Reset(d.interval)
+			// this is blocking so the stop will not be called until all the fetchers are finished
+			// in case there is a blocking fetcher it will halt (til the d.timeout)
+			go d.fetchIteration(ctx)
 		}
```
uri-weisman, line 81: I wonder if a graceful shutdown in which all the healthy fetchers completed their cycle is not better than stopping them immediately.
We can let the fetcher's timeout handle blocked fetchers, wdyt?

resources/providers/awslib/multi_region.go
```diff
@@ -100,11 +100,16 @@ func MultiRegionFetch[T any, K any](ctx context.Context, set map[string]T, fetch
 				return
 			}
 			crossRegionResources = append(crossRegionResources, results)
-		}(client, region)
+		}(client, region, errChan)
 	}
 
 	wg.Wait()
-	return crossRegionResources, err
+	select {
```
uri-weisman, line 107: Nice!

uri-weisman (changes requested):
Great Effort 🎖️

resources/fetchersManager/data.go
olegsu: You are right, graceful shutdown will break this way.

resources/providers/awslib/multi_region.go
```diff
@@ -100,11 +100,16 @@ func MultiRegionFetch[T any, K any](ctx context.Context, set map[string]T, fetch
...
+	case err := <-errChan:
+		return crossRegionResources, err
```
amirbenun, line 109: Another option here is to return a multi-error (From Go 1.20)
https://lukas.zapletalovi.com/posts/2022/wrapping-multiple-errors/

uri-weisman: Cool! will be useful once we'll run 1.20

resources/fetchersManager/data.go
```diff
@@ -30,48 +30,72 @@ import (
...
+func (d *Data) Run(ctx context.Context) Stop {
 	go d.fetchAndSleep(ctx)
-	return nil
+	called := false
+	return func(ctx context.Context, grace time.Duration) {
+		if called {
+			return
+		}
+		called = true
+		d.stopData(ctx, grace)
```
uri-weisman, line 66: Can we use here `sync.once` instead?

resources/fetchersManager/data.go
```diff
@@ -30,48 +30,72 @@ import (
...
+	// set immidiate exec for first time run
+	timer := time.NewTimer(0)
+	defer func() {
+		cancel()
+		timer.Stop()
+	}()
+	run := true
 	for {
 		select {
+		case grace := <-d.stopNotice:
```
uri-weisman, line 81: Is it possible to pass the main context to the stop function and cancel it after the grace period has ended, thereby achieving the same result?

uri-weisman (changes requested):

resources/fetchersManager/data.go
olegsu: I am not sure I understand. The `stopNotice` channel comes to make sure that no new fetchers will be executed once `stop` is called. How passing the main context will achieve this?

uri-weisman: We can store the `cancelFunc` generated in `fetchAndSleep` and call it once `stop` is called.
```
	ctx, d.cancelFunc = context.WithCancel(ctx)

and then in `stop`

       d.cancelFunc()

```

amirbenun (approved):
LGTM

resources/fetchersManager/data.go
uri-weisman: Decided on making it an atomic boolean.

olegsu: Agreed to use `atomic.Bool` instead

uri-weisman (approved):

PR #758, opened 2023-02-20
Update daily environment to use 8.7 SNAPSHOT

ofiriro3:
**Prerequisite**
- Before merging this change we need to replace the secret.API_KEY with a relevant stagging API_KEY.

### Summary of your changes
We would like to update the daily environment to use the 8.7 SNAPSHOT while using the stagging cloud environment.

**In-depth**
As developers, we would like that the daily environment to represent the current content in the `main` branch.
That way, we would be able to detect bugs earlier in the development process, bugs that may appear only after a few fetching cycles. 

** Currently 8.8 SNAPSHOT is broken, so this PR will deploy 8.7 SNAPSHOT.
I created another ticket to replace the hard-coded version with the latest SNAPSHOT.
- https://github.com/elastic/security-team/issues/6077

 **How to test**
Currently, the flow is deployed using a dispatch workflow, meaning - you will need to apply it using GitHub actions.
You can apply the flow from my fork, using [actions](https://github.com/ofiriro3/cloudbeat/actions/workflows/weekly-enviroment.yml).
Please set the branch to `weekly-enviroment-vanilla-work-with-SNAPSHOT` before running the dispatch workflow.
image

A new environment should be created with the 8.7 SNAPSHOT version. 
You will receive the credentials to a dedicated slack channel named `ofir-personal-channel`.

**Expected behavior**

- [ ] **A new 8.7 SNAPSHOT** will be created with the vanilla benchmark installed. 
- [ ] The cloud environment should be deployed on `staging.found.no`
- [ ] A slack message with the relevant credentials should be sent to `ofir-personal-channel`.
- [ ] You should get findings for the relevant benchmark.

### Related Issues
- https://github.com/elastic/security-team/issues/6016
- https://github.com/elastic/security-team/issues/6024

### Depends On 
- https://github.com/elastic/cloudbeat/pull/746


deploy/weekly-environment/scripts/benchmarks/kspm_vanilla/data/package_policy_vanilla.json
```diff
@@ -15,50 +15,24 @@
...
                 }
             ]
         }
     ],
     "package": {
         "name": "cloud_security_posture",
-        "title": "Kubernetes Security Posture Management (KSPM)",
-        "version": "1.1.1"
+        "title": "Security Posture Management (CSPM/KSPM)",
+        "version": "1.2.10"
```
oren-zohar, line 26: this version is not stable, no?

deploy/weekly-environment/variables.tf
```diff
@@ -7,7 +7,7 @@ variable "ec_api_key" {
 
 variable "ess_region" {
   default     = "gcp-us-central1"
-  description = "Optional ESS region where the deployment will be created. Defaults to gcp-us-west2"
+  description = "Optional ESS region where the deployment will be created. Defaults to gcp-us-central1"
```
oren-zohar, line 10: why did you change the region?

oren-zohar (changes requested):

oren-zohar (changes requested):
> As developers, we would like that the daily environment to represent the current content in the main branch.

Main is 8.8, and if this is the idea, why not dynamically use the version from the branch?

deploy/weekly-environment/variables.tf
ofiriro3: I didn't change the region, I just change the description so it will be aligned with the default value.

deploy/weekly-environment/scripts/benchmarks/kspm_vanilla/data/package_policy_vanilla.json
ofiriro3: As far as I know the broken integration is `1.2.11`, but the integration will auto-upgrade to `1.2.11` after the deployment.
As far as I saw I got findings as usual, although the agent wasn't healthy (on 1.2.11).

Do you think it would be better to deploy @jeniawhite version? `1.2.12-next`

ofiriro3 (commented):
Commented 🙏

deploy/weekly-environment/variables.tf
oren-zohar: So I think it should be aligned with the rest of the repo to `gcp-us-west2` as this is the region they publish the BCs to AFAIK

ofiriro3: But we use staging here, and staging contains only one region for GCP - `us-centeral1`

ofiriro3:
> Main is 8.8, and if this is the idea, why not dynamically use the version from the branch?

Yes that is a valid option.
Currently 8.8-SNAPSHOT is broken, so we put the 8.7-SNAPSHOT instead.

I'll update the description here.

This is a reference to an item I already created before:
- https://github.com/elastic/security-team/issues/6077

oren-zohar (approved):

PR #762, opened 2023-02-21
add re-run option

gurevichdmitry:
Re-run failed tests


gurevichdmitry:
pytest-rerunfailures package was tested. This package allow configuration of re-run independent tests. It may be used on our functional test suites. Integration test suites use dependency, when there are test cases which running is based on the result of the previous tests execution. pytest-rerunfailures package is not relevant for integration test suites. So the solution is partial and also not checked well on functional tests.

PR #772, opened 2023-02-23
Refactor Deployment Docs

oren-zohar:
### Summary of your changes
- Refactoring the deployment methods docs and splitting the ELK and Cloudbeat deployment options to make it clearer
- Pinning hermit GO version to match the repo version

PR #773, opened 2023-02-26
Reduce time in Cloudbeat CI

ofiriro3:
### Summary of your changes
After multiple attemps to run the `unit-test.yml` in the CI, it seems that it the racing validation step causes a timeout.
Instead of increasing the step timeout, and by that increasing the general time that the CI runs, we would like to merge to steps in the CI:
- The race validation check
- Coverage and unit test step

By merging these two into one step we hope to get the same result in less CI time.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -49,12 +49,8 @@ jobs:
 
       - name: Unit-Test Cloudbeat
         run: |
-          GOOS=linux go test -v -coverpkg=./... -coverprofile=cover.out.tmp ./...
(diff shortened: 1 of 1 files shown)


oren-zohar:
it's not because of the race flag, and we need this flag, closing

olegsu:
@oren-zohar the flag was just moved to the unit test, so we don't run the unit test twice.

oren-zohar:
will be fixed [here](https://github.com/elastic/cloudbeat/pull/772) (+ using the go dependencies caching to further reduce the running time)

oren-zohar (approved):

Issue #771, opened 2023-02-23
S3 provider API calls fails due to fetcher timeout


uri-weisman:
**Describe the bug**
We encounter an issue with some S3 API calls that fail to collect data about certain buckets, resulting in a `context deadline exceeded` error message. 
This is due to a hard-coded timeout of **5 minutes** for a single fetcher in Cloudbeat, which is causing problems as we have ~270 S3 buckets in our dev account.

It results in inconsistent findings because the timeout is not always exceeded. 
In addition, our S3 provider has implemented a fallback logic that sends the resource to evaluation with an empty string when an API error occurs, even though the required data has not been collected, this is causing OPA to generate failed findings.

**Evidence:**
When we reach the timeout, API calls start to fail.
```
12:24:13.257
elastic_agent.cloudbeat
[elastic_agent.cloudbeat][info] Starting S3Fetcher.Fetch

12:29:13.282
elastic_agent.cloudbeat
[elastic_agent.cloudbeat][error] Could not get bucket versioning for bucket demo-rds-dump. Err: operation error S3: GetBucketVersioning, context deadline exceeded
```

**Preconditions**
Run cloudbeat version 8.7/8.8.

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Run the CSPM integration using your build security dev account and notice the above error in the agent logs.

**Expected behavior**
1. Consistent findings for s3 buckets.
2. No `context deadline exceeded` error message should appear in the logs.

**Suggested solution**

Short-term (8.7)
Increase the timeout for a single fetcher to 10 mins and avoid sending an empty string in case we fail to retrieve the bucket's data.

Long-term (8.8)
Optimizing the data collection process in the S3 provider by fetching data for multiple buckets in parallel.

Issue #653, opened 2023-01-09
Fetchers timeout should be configurable


ari-aviran:
**Motivation**
Cloudbeat should have configurable fetchers timeout. The old timeout before CSPM fetchers was set to 1 minute, but fetchers that fetch data from an internet API (e.g. cloud resources) might take longer than that to produce resources for evaluation, and should not timeout.

**Definition of done**
- [ ] Allow configuration of fetchers timeout via `cloudbeat.yml`
- [ ] Set default CSPM timeout for 10 minutes
- [ ] Set default KSPM timeout for 1 minute

ofiriro3:
Interesting topic 👍

One question that I am curious about is - how will we define a timeout for all fetchers.
I mean some of the fetchers require more work/time than others.

For example, the ECR fetcher can have multiple requests depending on the number of pods it needs to process.

PR #775, opened 2023-02-26
Update s3 provider fallback logic

uri-weisman:
1. Modify the s3 provider to send nil values in case we failed to query the required data.
2. CSP policies version bump.

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/771
- Related: https://github.com/elastic/csp-security-policies/pull/200

### Checklist


flavors/posture.go
```diff
@@ -74,9 +74,9 @@ func NewPosture(_ *beat.Beat, cfg *agentconfig.C) (*posture, error) {
 		return nil, err
 	}
 
-	// TODO: timeout should be configurable and not hard-coded. Setting to 5 minutes for now to account for CSPM fetchers
+	// TODO: timeout should be configurable and not hard-coded. Setting to 10 minutes for now to account for CSPM fetchers
 	// 	https://github.com/elastic/cloudbeat/issues/653
-	data, err := fetchersManager.NewData(log, c.Period, time.Minute*5, fetchersRegistry)
+	data, err := fetchersManager.NewData(log, c.Period, time.Minute*10, fetchersRegistry)
```
oren-zohar, line 79: Honestly, having it configurable is the best approach. First, who should configure it, the user? from the integration page? how will the user know what is the right timeout? what is the other option? users will see an issue and will contact us? again, not the best solution since we will also not know what is the correct timeout, and most users will not even notice that there is an issue. I think the right approach should be something robust that will not require configuration from the user at all.

oren-zohar (changes requested):

resources/providers/awslib/s3/provider.go
```diff
@@ -117,7 +122,7 @@ func (p Provider) GetBucketPolicy(ctx context.Context, bucketName *string, regio
 	var bucketPolicy BucketPolicy
 	jsonErr := json.Unmarshal([]byte(*rawPolicy.Policy), &bucketPolicy)
 	if jsonErr != nil {
-		return nil, jsonErr
+		return map[string]any{}, jsonErr
```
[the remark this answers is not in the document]
uri-weisman: `bucket_policy` is a map.
`bucket_policy.statement` is an array.

resources/providers/awslib/s3/provider_test.go
```diff
@@ -140,20 +139,20 @@ func (s *ProviderTestSuite) TestProvider_DescribeBuckets() {
 				"GetBucketPolicy":     {{{mock.Anything, mock.Anything}, {nil, errors.New("bla")}}},
 				"GetBucketVersioning": {{{mock.Anything, mock.Anything}, {nil, errors.New("bla")}}},
 			},
-			expected:    []awslib.AwsResource{BucketDescription{Name: bucketName, SSEAlgorithm: "AES256", BucketPolicy: map[string]any(nil), BucketVersioning: BucketVersioning{false, false}}},
+			expected:    []awslib.AwsResource{BucketDescription{Name: bucketName, SSEAlgorithm: aws.String("AES256"), BucketPolicy: map[string]any(nil), BucketVersioning: nil}},
```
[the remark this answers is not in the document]
uri-weisman: Yes, instead of creating a string variable.

flavors/posture.go
uri-weisman: Got it, in my opinion, it should not be configurable by the user and this issue should be handled programmatically.
I'll open a ticket to discuss the matter.

ari-aviran: There's an issue in the comment https://github.com/elastic/cloudbeat/issues/653 you can expand the discussion

oren-zohar (approved):

Issue #727, opened 2023-02-12
Handle creds and configurations for VM workers and Trivy


jeniawhite:
**Summary**
Trivy needs to pull the same credentials including region as the configuration for the EC2 fetcher

**In Depth**
The worker that implements the fetching of EC2 instances, and the creation of snapshots is taking the AWS credentials configuration from the `cloudbeat.yml` or the fleet reconfiguration changes.
`Trivy` doesn't work like that (at least from my brief investigation I did not see an option to provide it any credentials), `Trivy` just uses the default resolver of AWS SDK.
We need to make sure that both of these flows work on the same credentials and are synced instead of having two different configurations.

**Definition of Done**
- [ ] Trivy works the same creds as the VM workers 
- [ ] All VM workers share the same creds
- [ ] Including region configuration

jeniawhite:
Looking at the `Trivy` code I've noticed that there is no option to provide any credentials.
The EBS is being scanned by the scanner:
```
case TargetVM:
		if report, err = r.ScanVM(ctx, opts); err != nil {
			return xerrors.Errorf("vm scan error: %w", err)
		}
```
Looking deeper you are able to see:
```
case TypeEBS:
		target = strings.TrimPrefix(target, TypeEBS.Prefix())
		e, err := newEBS(target, storage, opt.AWSRegion)
		if err != nil {
			return nil, xerrors.Errorf("new EBS error: %w", err)
		}
		return e, nil
```
Notice that we are able to pass `region` only:
```
func newEBS(snapshotID string, vm Storage, region string) (*EBS, error) {
	ebs, err := ebsfile.New(ebsfile.Option{
		AwsRegion: region,
	})
	if err != nil {
		return nil, xerrors.Errorf("new ebsfile error: %w", err)
	}

	return &EBS{
		Storage:    vm,
		snapshotID: snapshotID,
		ebs:        ebs,
	}, nil
}
```
The configuration of the `ebs` is done by a third-party repository (https://github.com/masahiro331/go-ebs-file)
```
type Option struct {
	AwsSecretKey string
	AwsAccessKey string
	AwsRegion    string
}

func New(option Option) (*EBS, error) {
	cfg, err := config.LoadDefaultConfig(context.TODO())
	if err != nil {
		return nil, err
	}
	if option.AwsRegion != "" {
		cfg.Region = option.AwsRegion
	}
	ebsClient := ebs.NewFromConfig(cfg)

	return &EBS{ebsClient}, nil
}
```
[...]

PR #781, opened 2023-02-28
Support Trivy credentials configuration

jeniawhite:
### Summary of your changes
Adding the ability to configure the AWS config that `Trivy` uses in order to scan EBS snapshots.
Currently, this is a temporary solution and is based on a forked `Trivy` using the following commits:
- https://github.com/masahiro331/go-ebs-file/pull/2
- https://github.com/jeniawhite/trivy/commit/77c59c27eda01973cfb9fe86e16bc3b19b4fa70b

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/727

### Checklist

### DIFF
--- a/vulnerability/worker.go
+++ b/vulnerability/worker.go
@@ -31,10 +31,9 @@ import (
 )
 
 type VulnerabilityWorker struct {
-	log      *logp.Logger
-	cfg      *config.Config
-	provider *ec2.Provider
-	// cloudIdentity *awslib.Identity
+	log        *logp.Logger
+	cfg        *config.Config
+	provider   *ec2.Provider
 	fetcher    VulnerabilityFetcher
 	replicator VulnerabilityReplicator
 	verifier   VulnerabilityVerifier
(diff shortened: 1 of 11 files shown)


vulnerability/worker.go
```diff
@@ -111,7 +113,7 @@ func (f *VulnerabilityWorker) Run(ctx context.Context) {
 			f.wg.Add(1)
 			go func() {
 				defer f.wg.Done()
-				err := f.fetcher.FetchInstances(ctx)
+				err := f.fetcher.FetchInstances(ctx, f.region)
```
amirbenun, line 116: Instead of passing the region over and over again, a more clean solution will be to define the region once when instantiating the provider.
The usage of the multiRegion struct is good IMO as it's a more generic way to get data. We already have `MultiRegionClientFactory` that initializes the multi-region provider. Maybe we can add some methods to this struct that specify just certain regions instead of using all of them by default. For example:
```
WithRegions(region ...string)
WithDefaultRegion()
```
What do you think?

amirbenun (commented):
Looks good, one suggestion regarding the multi region usage.

amirbenun (approved):

PR #796, opened 2023-03-07
implement tests for aws iam management rules

gurevichdmitry:
### Summary of your changes

This PR adds functional test for AWS IAM management rules.


tests/commonlib/utils.py
```diff
@@ -282,3 +282,10 @@ def get_findings(elastic_client, config_timeout, match_type):
         time.sleep(1)
 
     return result
+
+
+def identifier_by_name(case_identifier, eval_resource) -> bool:
```
oren-zohar, line 287: Please add docs in any case

gurevichdmitry: documentation is added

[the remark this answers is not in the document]
gurevichdmitry: @oren-zohar I'm agree with you. The [task](https://github.com/elastic/security-team/issues/6115) for this issue already exists. In current context it will be good enough to move out the identifier function from the test function body in order to not duplicate the code.

oren-zohar (approved):

PR #797, opened 2023-03-07
Fix Build step for EKS-CI

oren-zohar:
### Summary of your changes

Fixing the build step in `EKS-CI` as it doesn't need the hermit for it anymore (it uses the cached go deps which cause conflict with hermit)

### Checklist


gurevichdmitry (approved):
LGTM

Issue #696, opened 2023-02-02
 Implement vulnerabilities database periodical updates


amirbenun:
**Motivation**
As we have discussed on a separate ticket (see below), the vulnerability database trivy uses needs to be updated before every cycle in order to provide a reliable product to our users.

**Definition of done**
- [ ] Update trivy database before every vulnerability scanning cycle
- [ ] Trivy database should be cached using the built-in cache layer
- [ ] When the cache doesn't exist, a failed database download should result in cloudbeat exit with error
- [ ] When the cache exists, a failed database download should change cloudbeat to an unhealthy state (and keep running)

**Out of scope**
- https://github.com/elastic/cloudbeat/issues/697

**Related tasks/epics**
- https://github.com/elastic/cloudbeat/issues/624

kfirpeled:
@amirbenun do we send the DB version as part of the vulnerability?

amirbenun:
@kfirpeled we don't, Trivy doesn't expose it.

kfirpeled:
Thanks @amirbenun 
For the upcoming QA cycle, how do you suggest to verify this ticket?

uri-weisman:
Verified - DB is updated every cycle (24h)



![Image](https://github.com/elastic/cloudbeat/assets/68195305/707c3069-bc9a-45b0-ac4a-c637e2760470)

PR #799, opened 2023-03-08
Skip vulnerabilities DB update when scanning snapshots

amirbenun:
### Summary of your changes
Skips the DB update on scan operations so that it will only get updated before every cycle when instantiating the trivy runner.
Resolves https://github.com/elastic/cloudbeat/issues/696

### Checklist


flavors/vulnerability.go
```diff
@@ -96,16 +86,17 @@ func (bt *vulnerability) Run(b *beat.Beat) error {
 
 	// TODO: Should trigger first cycle without waiting for ticker.
 	bt.log.Warn("vulnerability.Run ticker running for period ", bt.config.Period)
-	workerTicker := time.NewTicker(bt.config.Period)
+	workerTicker := time.NewTicker(1)
```
uri-weisman, line 89: Instead of resetting the ticker, what do u think about wrapping the logic we have under `workerTicker.C` in a function and just calling it before the for loop?

uri-weisman (changes requested):
Cool!
If the agent will run for a long time don't we want a follow up task in which the DB will be updated every pre-defined interval?

PR #837, opened 2023-03-31
magefile: BuildOpaBundle: Always cleanup

orestisfl:
### Summary of your changes

Always cleanup the /tmp directory when `BuildOpaBundle` is run.
This improves behaviour in the case where a build is cancelled mid-build (e.g. with Ctrl-C). Before the change, the directory would remain in /tmp and on the next build cloning it would fail, leading to a confusing error.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -23,7 +23,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"github.com/pkg/errors"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


magefile.go
```diff
@@ -338,11 +337,18 @@ func PythonEnv() error {
 	return err
 }
 
-func BuildOpaBundle() error {
+func BuildOpaBundle() (err error) {
 	owner := "elastic"
 	r := "csp-security-policies"
 	cspPoliciesPkgDir := "/tmp/" + r
 
+	defer func() {
```
olegsu, line 345: I think better to just use the `tempDir` 
https://go.dev/src/io/ioutil/tempfile.go

olegsu (changes requested):

olegsu (approved):

PR #838, opened 2023-03-31
Add .dockerignore to speed-up docker builds

orestisfl:
### Summary of your changes

Add .dockerignore to speed-up docker builds
Without this change, the whole directory has to be copied to the docker context, unnecessarily copying large folders like `vendor/`.


uri-weisman (approved):
💪

Issue #607, opened 2022-12-26
Implement the side scanning logic (snapshotting, deleting snapshots)


oren-zohar:
**Summary**
 
- Implement the side-scanning logic (snapshotting, deleting snapshots)
- Deleting snapshots 
  - What are the risks?
  - Is there an alternative for deleting? TTL?
- Add support to the new configuration
- Should it be a fetcher? a new component?

**Out of scope**
- Snapshotting in serverless

tehilashn:
I think it's worth checking what our competitors do with snapshots and whether they delete them. 
In addition, we should explore archiving snapshots and not deleting them

tehilashn:
Thanks @uri-weisman !
@jeniawhite - now that this is in I recommend our E2E will be done with the latest version since I think it's a significant change.

kfirpeled:
QA: deploy on a dedicated region for the QA cycle, and post which region is being used for the tests to make sure no one will use the same region

CohenIdo:
✅  BC1 QA Cycle: I deployed vul integration using cloud formation, once the cycle ended I checked the SNAPSHOT status in aws and there were deleted: 



![Image](https://user-images.githubusercontent.com/90558359/236240148-aba92c2e-f88f-4488-8066-61b861539a1c.png)

PR #845, opened 2023-04-03
Add snapshot cleanup to vulnerability management

uri-weisman:
### Summary of your changes
1. Delete generated snapshots at the end of the cycle.
2. Add a name tag to the snapshots we create.

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/607

### Checklist


resources/providers/awslib/ec2/provider.go
```diff
@@ -231,3 +232,18 @@ func (p *Provider) DescribeSnapshots(ctx context.Context, snapshot EBSSnapshot)
 	}
 	return result, nil
 }
+
+func (p *Provider) DeleteSnapshots(ctx context.Context, ids []string) error {
+	_, err := awslib.MultiRegionFetch(ctx, p.clients, func(ctx context.Context, region string, c Client) ([]Ec2Instance, error) {
+		for _, id := range ids {
+			_, err := c.DeleteSnapshot(ctx, &ec2.DeleteSnapshotInput{SnapshotId: aws.String(id)})
+			if err != nil {
+				p.log.Errorf("Error deleting snapshot %s: %v", id, err)
```
orestisfl, line 241: Did you mean to `return` the `err` here?

vulnerability/replicator.go
```diff
@@ -40,7 +40,7 @@ func NewVulnerabilityReplicator(log *logp.Logger, provider *ec2.Provider) Vulner
 	}
 }
 
-func (f VulnerabilityReplicator) SnapshotInstance(ctx context.Context, insCh chan ec2.Ec2Instance) {
+func (f VulnerabilityReplicator) SnapshotInstance(ctx context.Context, insCh chan ec2.Ec2Instance, ids *[]string) {
```
orestisfl, line 43: Personal preference but I think `return`ing the `ids` is slightly cleaner.

resources/providers/awslib/ec2/provider.go
```diff
@@ -231,3 +232,18 @@ func (p *Provider) DescribeSnapshots(ctx context.Context, snapshot EBSSnapshot)
 	}
 	return result, nil
 }
+
+func (p *Provider) DeleteSnapshots(ctx context.Context, ids []string) error {
+	_, err := awslib.MultiRegionFetch(ctx, p.clients, func(ctx context.Context, region string, c Client) ([]Ec2Instance, error) {
```
amirbenun, line 237: Note that a snapshot resides in a single region, and so, deleting a snapshot should be done only in the region where it exists. In that case, you can accept an argument of type `EBSSnapshot` as it has the region and also the snapshot id. (see example in DescribeSnapshots)
Since we don't know if all the snapshots are of the same region, the provider can accept a single `EBSSnapshot`, and you could rename the method to `DeleteSnapshot` (singular).

vulnerability/worker.go
```diff
@@ -116,6 +118,13 @@ func (f *VulnerabilityWorker) Run(ctx context.Context) {
 			f.log.Info("VulnerabilityWorker.work waiting on workers")
 			f.wg.Wait()
 			f.log.Info("VulnerabilityWorker.work finished waiting on workers")
+
+			err := f.cleaner.CleanSnapshots(ctx, snapshotIds)
```
amirbenun, line 122: I would try to align with other methods that get called here.
It is worth spending some time to make `CleanSnapshots` run in a goroutine and accept a channel of snapshots to delete.

amirbenun (commented):
A few notes

vulnerability/worker.go
uri-weisman: Seems like it makes the cleanup flow complicated.
Decided for the current iteration to clean snapshots at the end of the cycle without aligning with the rest of the steps.

resources/providers/awslib/ec2/provider.go
```diff
@@ -196,7 +197,15 @@ func (p *Provider) CreateSnapshots(ctx context.Context, ins Ec2Instance) ([]EBSS
 		InstanceSpecification: &types.InstanceSpecification{
 			InstanceId: ins.InstanceId,
 		},
-		Description: aws.String("Cloudbeat Vulnerability Snapshot."),
+		Description: aws.String("Uri Vulnerability test."),
```
amirbenun, line 200: remove `Uri`

vulnerability/cleaner.go
```diff
@@ -0,0 +1,51 @@
...
+
+func NewVulnerabilityCleaner(log *logp.Logger, provider snapshotCleaner) Cleaner {
+	log.Debug("VulnerabilityCleaner: New")
+	return Cleaner{
+		log:      log,
+		provider: provider,
+	}
+}
+
+func (f Cleaner) CleanSnapshots(ctx context.Context, snapshots []ec2.EBSSnapshot) {
```
amirbenun, line 43: Add unit tests. (Just make sure that the provider mock gets called over and over again)

vulnerability/worker.go
```diff
@@ -96,11 +98,11 @@ func (f *VulnerabilityWorker) Run(ctx context.Context) {
 				f.log.Info("VulnerabilityWorker.work FetchInstances finished")
 			}()
 			f.wg.Add(1)
-			go func() {
+			go func(snapshots *[]ec2.EBSSnapshot) {
```
amirbenun, line 101: Not a fan of this solution but just in order to move forward.

resources/providers/awslib/ec2/provider.go
```diff
@@ -231,3 +240,16 @@ func (p *Provider) DescribeSnapshots(ctx context.Context, snapshot EBSSnapshot)
...
 }
+
+func (p *Provider) DeleteSnapshot(ctx context.Context, snapshot EBSSnapshot) error {
+	client, err := awslib.GetClient(aws.String(snapshot.Region), p.clients)
+	if err != nil {
+		return err
+	}
+	_, err = client.DeleteSnapshot(ctx, &ec2.DeleteSnapshotInput{SnapshotId: aws.String(snapshot.SnapshotId)})
+	if err != nil {
+		return fmt.Errorf("error deleting snapshot %s: %v", snapshot.SnapshotId, err)
```
orestisfl, line 251:
```suggestion
		return fmt.Errorf("error deleting snapshot %s: %w", snapshot.SnapshotId, err)
```

`%v` does not wrap errors: https://go.dev/play/p/JdJZ6ycbtF1

vulnerability/worker.go
```diff
@@ -77,8 +79,8 @@ func NewVulnerabilityWorker(log *logp.Logger, c *config.Config) (*VulnerabilityW
 }
 
 func (f *VulnerabilityWorker) Run(ctx context.Context) {
-	// TODO: Handle deletion of snapshots
 	f.log.Info("Starting VulnerabilityWorker.work")
+	snapshots := make([]ec2.EBSSnapshot, 0)
```
orestisfl, line 83: To avoid use of pointers completely, this can be made into a channel. If the main thread pops from the channel before `wg.Wait()` the channel doesn't even have to be buffered I think. I feel it would be more idiomatic that way.

uri-weisman: Agree, seems to be more elegant.
However, the ch has to be buffered in order for the pushing action to not be blocked.

vulnerability/worker.go
uri-weisman: Modified

resources/providers/awslib/ec2/provider.go
olegsu: I would suggest also updating the linter [settings](https://github.com/elastic/cloudbeat/blob/main/.golangci.yaml)

amirbenun (approved):

Issue #853, opened 2023-04-10
Vulnerability management single cycle trigger


jeniawhite:
**Describe the bug**
I've run the vulnerability management flavor and saw that it only worked for a single cycle, it looks like we are not triggering multiple cycles.
Trying to code in a Golang playground it looks like it should work. 

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Run vulnerability management

**Expected behavior**
We should trigger multiple cycles

PR #856, opened 2023-04-10
Avoid deadlock on Trivy database

jeniawhite:
### Summary of your changes
During refactoring, we've moved the code that initializes the scanner instead of being global to being created every cycle from scratch.
This caused Trivy to deadlock the database.
In order to avoid this, I've added a closure of the database after every run.

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/853

### Checklist


vulnerability/worker.go
```diff
@@ -79,6 +79,8 @@ func NewVulnerabilityWorker(log *logp.Logger, c *config.Config) (*VulnerabilityW
 func (f *VulnerabilityWorker) Run(ctx context.Context) {
 	// TODO: Handle deletion of snapshots
 	f.log.Info("Starting VulnerabilityWorker.work")
+	// TODO: Maybe use the context from the worker? But what if the worker is canceled?
```
olegsu, line 82: What would happen if the worker is canceled and this continues to run?
imh it should be context from worker

PR #857, opened 2023-04-11
Send all aws configs as a single resource

orouz:
### Summary of your changes

in [CIS AWS 3.5](https://github.com/elastic/csp-security-policies/pull/207) we need to check **all** the aws config objects match a certain schema. this PR changes the relevant fetcher to send all configs as a single resource for evaluation. 

### Screenshot/Data
Screenshot 2023-04-20 at 9 24 47
tested on 
|    component | version |
|-----------|------------------------------------------------|
| kibana        | 8.8.0                                          |
| es        | 8.8.0                                          |
| CSPM      | 1.2.13                                         |
| cloudbeat | 8.8.0                                          |
| policy    | v1.3.9                                         |

### Related Issues

- https://github.com/elastic/cloudbeat/pull/721
- closes https://github.com/elastic/csp-security-policies/issues/188

### Checklist


resources/fetchers/logging_fetcher.go
```diff
@@ -68,9 +68,11 @@ func (f LoggingFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetad
 		f.log.Errorf("failed to describe config recorders: %v", err)
 	}
 
-	f.resourceCh <- fetching.ResourceInfo{
-		Resource:      ConfigResource{configs: configs, identity: f.cloudIdentity},
-		CycleMetadata: cMetadata,
+	if len(configs) > 0 {
```
orouz, line 71: i'm not sure, but maybe if the are no configs at all, the rule should fail?

olegsu: how can it happen? the config is either enabled or disabled, no?

orouz: `configs` can be empty when `MultiRegionFetch` fails to fetch and returns an empty `[]awslib.AwsResource` and an error. we log the error but shouldn't push to the resource channel. in second thought tho this shouldn't fail the rule as it's not operating on any resources.

olegsu (approved):

resources/fetchers/logging_fetcher.go
orouz: replaced the `count` check with returning early when we get an error

uri-weisman (approved):

uri-weisman:
test & ship

Issue #745, opened 2023-02-14
Add region to CSPM findings


ari-aviran:
We want to have the region of a resource in CSPM findings. We should probably use the [`cloud.region`](https://www.elastic.co/guide/en/ecs/current/ecs-cloud.html#field-cloud-region) ECS field.

**Definition of done**
- [ ] A region field is present in every CSPM finding of a resource that is region-bound
- [x] Define a _global_ value and include it on non region specific resources
- [x] Document the field (and the global value) in the technical index
- [x] Add the field to the mapping in the CSP integration

eyalkraft:
Let's also

- [ ] Define a _global_ value and include it on non region specific resources + document it in the technical index

Omolola-Akinleye:
Added a filter query ->  `cloud.region: *` in Discover 
![Image](https://user-images.githubusercontent.com/17135495/235554025-a3d2e5dc-32b6-40ae-af36-721c1cc030d8.png)

Checked mappings in  stack management - logs-cloud_security_posture.findings_latest-default
![Image](https://user-images.githubusercontent.com/17135495/235554438-043686c9-e403-4c87-8b4b-98354eaa4bcb.png)

PR #848, opened 2023-04-09
Vulnerability index structure

jeniawhite:
### Summary of your changes
Implementation of vulnerability index schema inside `cloudbeat`.

### Screenshot/Data
Example of a single finding that it generates.
```json
{
  "_index": ".ds-logs-cloud_security_posture.vulnerabilities-default-2023.04.09-000002",
  "_id": "ZvmnbIcBJR1IOaing7FG",
  "_version": 1,
  "_score": 0,
  "_source": {
    "agent": {
      "name": "ip-172-31-26-22",
      "id": "045369a6-2670-43c1-ba69-8f2dcec8b83b",
      "type": "cloudbeat",
      "ephemeral_id": "bcd96458-d221-4124-9c29-d2a6852244b9",
      "version": "8.8.0"
    },
    "resource": {
      "ID": "0985018013c800117",
      "Name": "github_runner01"
    },
    "elastic_agent": {
      "id": "045369a6-2670-43c1-ba69-8f2dcec8b83b",
      "version": "8.8.0",
      "snapshot": true
    },
    "vulnerability": {
      "severity": "HIGH",
      "package": {
        "fixed_version": "4.3.2",
        "name": "System.Private.Uri",
        "version": "4.3.0"
      },
      "description": "A denial of service vulnerability exists when .NET Framework or .NET Core improperly handle web requests, aka '.Net Framework and .Net Core Denial of Service Vulnerability'. This CVE ID is unique from CVE-2019-0820, CVE-2019-0981.",
      "classification": "CVSS",
      "title": "dotnet: infinite loop in Uri.TryCreate leading to ASP.Net Core Denial of Service",
      "data_source": {
        "ID": "ghsa",
        "URL": "https://github.com/advisories?query=type%3Areviewed+ecosystem%3Anuget",
        "Name": "GitHub Security Advisory Nuget"
      },
      "reference": "https://avd.aquasec.com/nvd/cve-2019-0980",
      "cwe": [
        "CWE-19"
      ],
      "score": {
        "impact": 7.5,
        "version": "3.0",
        "base": 7.5
      },
      "scanner": {
        "vendor": "Trivy",
        "version": "v0.35.0"
      },
      "id": "CVE-2019-0980",
      "enumeration": "CVE",
      "published_date": "2019-05-16T19:29:00Z",
      "cvss": {
        "redhat": {
          "V3Vector": "CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H",
[...]

PR #864, opened 2023-04-16
Add cloud data to VM events

uri-weisman:
### Summary of your changes
I made more extensive changes than required for vulnerability management because we have an additional CSPM task of adding the region to the findings and I don't want code duplications.

1. Add cloud data to the vm events.
2. Modify the events creator to act as another pipeline step.

### Screenshot/Data
Added fields
Screenshot 2023-04-16 at 14 25 03

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/745

### Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -79,8 +79,8 @@ func New(cfg *config.C) (*Config, error) {
 
 func defaultConfig() (*Config, error) {
 	ret := &Config{
-		Period:    4 * time.Hour,
-		Benchmark: CIS_K8S,
+		Period: 4 * time.Hour,
+		//Benchmark: CIS_K8S,
 	}
 
 	bundle, err := getBundlePath()
--- a/config/config.go
+++ b/config/config.go
@@ -80,7 +80,6 @@ func New(cfg *config.C) (*Config, error) {
 func defaultConfig() (*Config, error) {
 	ret := &Config{
 		Period: 4 * time.Hour,
-		//Benchmark: CIS_K8S,
 	}
 
 	bundle, err := getBundlePath()
--- a/vulnerability/events_creator.go
+++ b/vulnerability/events_creator.go
@@ -18,6 +18,11 @@
 package vulnerability
 
 import (
+	"context"
+	"github.com/elastic/cloudbeat/dataprovider"
+	"github.com/elastic/cloudbeat/resources/fetching"
+	"github.com/elastic/cloudbeat/resources/providers/awslib/ec2"
+	"github.com/elastic/elastic-agent-libs/logp"
 	"strings"
 	"time"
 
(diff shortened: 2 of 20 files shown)


config/config.go
```diff
@@ -79,8 +79,7 @@ func New(cfg *config.C) (*Config, error) {
 
 func defaultConfig() (*Config, error) {
 	ret := &Config{
-		Period:    4 * time.Hour,
-		Benchmark: CIS_K8S,
```
uri-weisman, line 83: why is it needed? all integrations besides vulnerability management are sending benchmark type.
It causes us to enrich vm config with the wrong benchmark (`cis_k8s`).

vulnerability/events_creator.go
```diff
@@ -45,7 +49,7 @@ type Vulnerability struct {
 	Description    string     `json:"description,omitempty"`
 	Severity       string     `json:"severity,omitempty"`
 	Classification string     `json:"classification,omitempty"`
-	PublishedDate  time.Time  `json:"published_date,omitempty"`
+	PublishedDate  *time.Time `json:"published_date,omitempty"`
```
uri-weisman, line 52: Bug fix - not all vulnerabilities contain `PublishedDate`, causing null pointer dereference.

vulnerability/events_creator.go
```diff
@@ -148,14 +188,26 @@ func createVulnerabilityEvent(reportResult trivy_types.Result, vul trivy_types.D
...
+	if err != nil {
+		e.log.Errorf("failed to enrich event: %v", err)
+	}
+
+	return event
 }
 
 func getIdentifierType(id string) string {
-	return strings.Split(id, "-")[0]
+	s := strings.Split(id, "-")
```
uri-weisman, line 205: better be safe than sorry

config/config.go
olegsu: Does it cause a change in the schema of the events from other integrations as well?

uri-weisman: All other integrations send the `benchmark` value so they are not affected by the default config.

olegsu (approved):

vulnerability/events_creator.go
jeniawhite: Maybe we shouldn't work with pointers directly in our structures and have something similar to this?
https://github.com/elastic/cloudbeat/pull/865

My initial solution was exactly like your fix here is the commit:
https://github.com/elastic/cloudbeat/pull/848/commits/e23845639bb60977afd37ee2968e0152908bde40

uri-weisman: @jeniawhite - that's a good solution as well.
IMHO, it will be a better approach to omit the field instead of sending the wrong data in case of a missing value.

Issue #832, opened 2023-03-30
E2E testing


oren-zohar:
**Summary**
Before continuing with the rest of the priority zero tasks, we need to verify our application works end-to-end to discover any critical issues

**Definition of done**
- [x] our solution works OOTB with the latest image, with the integration, using the cloud formation script deployment
- [x] our solution works OOTB with the latest image, with the integration, locally
- [x] scanning images completely (not encrypted)
- [x] review logs for any unexpected errors

> **Note**
> The verification should include multiple cycles and images

jeniawhite:
Woking on the E2E, I've opened the following tickets:
- https://github.com/elastic/kibana/issues/154233
- https://github.com/elastic/cloudbeat/issues/852
- https://github.com/elastic/cloudbeat/issues/853
- https://github.com/elastic/cloudbeat/issues/854

tehilashn:
@jeniawhite - thank you for the thorough work you've been doing in this e2e testing!

As part of this task, I recommend we also verify that we did not create any regression for KSPM and CSPM. 
Can you we please test the following? 

- [ ] Test # 1: install CSPM and KSPM using the 8.8 agent and cloudbeat (you can add this to your existing environment) 
- [ ] Test # 2: 
(1) Set up 8.7 ELK with CSPM and KSPM Integrations sending data
(2) Upgrade ELK to 8.8, ensure 8.7 version agent keeps working
(3) Deploy 8.8 version agent, and ensure it works as expected

olegsu:
More issues that were found by @jeniawhite 

Kibana
- [ ] Missing fields and empty rows in the finding
- [ ] Transform health check posture type is incorrect, we should verify it in a different way
- [ ] The transform seems to cut off a lot of findings


Cloudbeat
- [ ] Add the missing permission for the `commonDataProvider` (related to https://github.com/elastic/cloudbeat/pull/864, https://github.com/elastic/security-team/issues/6339)
- [ ] Document the dev flow to support ssh connection to the instance and the snapshot link to download the cloudbeat



**Empty rows**
![Image](https://user-images.githubusercontent.com/23185631/232719813-5b08e78f-227c-4798-8f9e-bd009ffa50f3.png)

jeniawhite:
Attempted to do the upgrade flow for 8.7.1 -> 8.8 BC.
Doing that I've deployed KSPM (on kind) and CSPM (on EC2 as binary).
[...]

PR #865, opened 2023-04-16
Vulnerability pointer published date non existent

jeniawhite:
### Summary of your changes
Handle cases when we cannot dereference the pointer due to field non-existence and panic

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/832

### Checklist

### DIFF
--- a/vulnerability/events_creator.go
+++ b/vulnerability/events_creator.go
@@ -148,7 +148,7 @@ func createVulnerabilityEvent(reportResult trivy_types.Result, vul trivy_types.D
 				Description:    vul.Description,
 				Severity:       vul.Severity,
 				Classification: vulScoreSystemClass,
-				PublishedDate:  *vul.PublishedDate,
+				PublishedDate:  getPublishedDate(vul),
 			},
 		},
 	}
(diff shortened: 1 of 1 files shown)


uri-weisman:
Great @jeniawhite!
I've fixed it as part of https://github.com/elastic/cloudbeat/pull/864

vulnerability/events_creator.go
```diff
@@ -193,3 +193,10 @@ func getDataSource(vul trivy_types.DetectedVulnerability) DataSource {
 		URL:  vul.DataSource.URL,
 	}
 }
+
+func getPublishedDate(vul trivy_types.DetectedVulnerability) time.Time {
+	if vul.PublishedDate == nil {
```
olegsu, line 198: The default vaule of `time.Time` is `0001-01-01 00:00:00 +0000 UTC`. I would consider `time.Now()`

uri-weisman: IMO, we can omit this field in case it does not have a value.

jeniawhite:
Merged this in this PR
- https://github.com/elastic/cloudbeat/pull/864

PR #866, opened 2023-04-17
Cloudformation script to support snapshot deletion

uri-weisman:
### Summary of your changes
Add permission of snapshot deletion to the role Cloudformation attaches to the newly created ec2.

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/607

### Checklist

### DIFF
--- a/deploy/cloudformation/elastic-agent-ec2.yml
+++ b/deploy/cloudformation/elastic-agent-ec2.yml
@@ -60,6 +60,7 @@ Resources:
             - ec2:DescribeSnapshots
             - ec2:CreateSnapshots
             - ec2:CreateTags
+            - ec2:DeleteSnapshot
             Resource: "*"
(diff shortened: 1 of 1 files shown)


olegsu (approved):

deploy/cloudformation/elastic-agent-ec2.yml
```diff
@@ -60,6 +60,7 @@ Resources:
             - ec2:DescribeSnapshots
             - ec2:CreateSnapshots
             - ec2:CreateTags
+            - ec2:DeleteSnapshot
             Resource: "*"
```
olegsu, line 64: Not related to this pr, but shouldn't we have only ec2 resources here?

uri-weisman: you are right. it will be safer

PR #868, opened 2023-04-18
Add score fields to vm findings

uri-weisman:
### Checklist


uri-weisman:
It decided to wait with implementation until we get new product requirements.

PR #886, opened 2023-04-20
Add AccessAnalyzer resource to implement CIS AWS 1.20

orestisfl:
### Summary of your changes

Uses the [accessanalyzer](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/accessanalyzer) packages from the AWS SDK to fetch all access analyzers per region. One resource is created per region since for CIS AWS 1.20 we need to check if there is at least one analyzer per region.

### Related Issues

https://github.com/elastic/csp-security-policies/issues/222

### Checklist


flavors/common.go
```diff
@@ -57,7 +57,7 @@ func getAWSDataProvider(ctx context.Context, log *logp.Logger, cfg config.Config
 	}
 
 	identityClient := awslib.GetIdentityClient(awsConfig)
-	iamProvider := iam.NewIAMProvider(log, awsConfig)
+	iamProvider := iam.NewIAMProvider(log, awsConfig, nil)
```
orestisfl, line 60: Not super happy with allowing `nil` here but I don't have a better way right now.

`getAWSDataProvider` doesn't need to use the access accessanalyzer API so having to add the multi-region boilerplate is wrong. The one real alternative is having a whole new fetcher just for access analyzers which seems like a worse solution.

olegsu: I agree, there are better ways in golang to construct instances and we need to handle that soon

olegsu:
Please update the csp version prior merge

PR #889, opened 2023-04-23
Allow ssh connections for CloudFormation template

amirbenun:
### Summary of your changes
I decide to rewrite in golang since we have a plan on having `ccli` tool to deploy cloudbeat everywhere.
I figured if this command will be written in go, it will be easier to merge into the `ccli`.
Besides that, the app accept all parameters from `.env` file (in gitignore) to avoid pushing secrets (enrollment token) into our repo.

Usage:
From `deploy/cloudformation` the run command is `go run go*`.
This command accepts raw environment variables or it will try to read a `.env` file in the form:
```
STACK_NAME="my-env"
FLEET_URL="https://"
ENROLLMENT_TOKEN="SOME-TOKEN"
DEV=true
KEY_NAME="amir-build" // Valid key in my AWS region
```
Notice that `KEY_NAME` is only required if `DEV=true`

### Related Issues
- Resolves: https://github.com/elastic/security-team/issues/6320

### Checklist


deploy/cloudformation/godev.go
```diff
@@ -0,0 +1,85 @@
...
+
+func editArtifactURL(content string) string {
+	prodURL := "https://artifacts.elastic.co/downloads/beats/elastic-agent/"
+
+	// TODO: Dynamically get the latest snapshot URL
+	devURL := "https://snapshots.elastic.co/8.8.0-3f572553/downloads/beats/elastic-agent/"
+	return strings.ReplaceAll(content, prodURL, devURL)
+}
+
+func allowSSHFromAnywhere(content string) string {
```
olegsu, line 24: there is no way to do that in a more idiomatic way?
I would expect to have something like (instead of replacing strings)
```go
stack := Cloudformation{}
stack.Resource["ElasticAgentSecurityGroup"].Properties.GroupDescription = "Allow SSH from anywhere"
stack.Resource["ElasticAgentSecurityGroup"].Properties.SecurityGroupIngress = ...
```

olegsu (commented):
Would suggest to wrap it with just command and add some docs with exmaple

olegsu (approved):

PR #890, opened 2023-04-23
Fix timeout for post-merge aws functional tests.

gurevichdmitry:
### Summary of your changes

- Cloudbeat period was increased to 600 seconds.
- CI action (cloudbeat deployment) waiting timeout was increased to 600s.
- Fixed just commands to support `kind-mono` and `kind-multi` deployments


.github/workflows/k8s-ci.yml
```diff
@@ -328,9 +328,19 @@ jobs:
...
 
+      - name: Set Kind Configuration
+        if: matrix.test-target == 'aws'
+        run: |
+          echo "KIND_CONFIG=kind-mono" >> $GITHUB_ENV
+
+      - name: Set Kind Configuration
+        if: matrix.test-target != 'aws'
+        run: |
+          echo "KIND_CONFIG=kind-multi" >> $GITHUB_ENV
```
olegsu, line 339: Can we do it in one step?

olegsu (changes requested):

olegsu:
Do we have a plan for how to overcome this? waiting for 10 min seems like something we want to avoid.

.github/workflows/k8s-ci.yml
gurevichdmitry: @olegsu github actions do not support "if else", I can implement inline script for this functionality.

gurevichdmitry:
> Do we have a plan for how to overcome this? waiting for 10 min seems like something we want to avoid.

@olegsu, Not for this case. Fetching resources from AWS depends on some factors - network, cloud services availability, resources count, etc. In current context, the fetching max time is limited to 10 minutes. By design, timeouts are provided at all stages, and customization depends on what we are trying to reach. Currently the requirement is to execute aws functional tests on post-merge and we need to retrieve stable results (not flaky). AWS functional tests run takes around 17 minutes and executed in parallel with other test jobs. I think this is acceptable time and improvement is not required in this case.

olegsu (approved):

Issue #902, opened 2023-04-27
Flaky test: pipeline.TestStep


orestisfl:
**Describe the bug**

Found in https://github.com/elastic/cloudbeat/actions/runs/4817244021/jobs/8577715908?pr=900

**To Reproduce**

```shell
$ while go test -race -count=1 -v ./pipeline; do go clean -testcache ; done
=== RUN   TestStep
=== RUN   TestStep/Should_receive_value_from_output_channel
=== RUN   TestStep/Pipeline_function_returns_error_-_no_value_received
--- PASS: TestStep (0.00s)
    --- PASS: TestStep/Should_receive_value_from_output_channel (0.00s)
    --- PASS: TestStep/Pipeline_function_returns_error_-_no_value_received (0.00s)
PASS
ok  	github.com/elastic/cloudbeat/pipeline	0.019s
=== RUN   TestStep
=== RUN   TestStep/Should_receive_value_from_output_channel
=== RUN   TestStep/Pipeline_function_returns_error_-_no_value_received
--- PASS: TestStep (0.00s)
    --- PASS: TestStep/Should_receive_value_from_output_channel (0.00s)
    --- PASS: TestStep/Pipeline_function_returns_error_-_no_value_received (0.00s)
PASS
ok  	github.com/elastic/cloudbeat/pipeline	0.019s
=== RUN   TestStep
=== RUN   TestStep/Should_receive_value_from_output_channel
=== RUN   TestStep/Pipeline_function_returns_error_-_no_value_received
--- PASS: TestStep (0.00s)
    --- PASS: TestStep/Should_receive_value_from_output_channel (0.00s)
    --- PASS: TestStep/Pipeline_function_returns_error_-_no_value_received (0.00s)
PASS
ok  	github.com/elastic/cloudbeat/pipeline	0.019s
...
=== RUN   TestStep
=== RUN   TestStep/Should_receive_value_from_output_channel
=== RUN   TestStep/Pipeline_function_returns_error_-_no_value_received
--- PASS: TestStep (0.00s)
    --- PASS: TestStep/Should_receive_value_from_output_channel (0.00s)
    --- PASS: TestStep/Pipeline_function_returns_error_-_no_value_received (0.00s)
PASS
ok  	github.com/elastic/cloudbeat/pipeline	0.019s
=== RUN   TestStep
=== RUN   TestStep/Should_receive_value_from_output_channel
    pipeline_test.go:69: 
[...]

PR #909, opened 2023-04-30
Fix TestStep race condition

orestisfl:
CollectResources is problematic because it tests using len() on a channel, which only uses the length of the queue of the channel.

### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/902

### Checklist


uri-weisman (approved):
Great!

Issue #897, opened 2023-04-25
Move non-ECS fields from root level event


kfirpeled:
**Motivation**

Currently, having custom (non ECS) fields at the root level of vulnerability is a wrong doing
It can cause conflicts with other integrations

The proposed way for integrations having customized fields is to add them under the package name
In our case, `cloud_security_posture.*`

**Definition of done**
- [ ] Move/remove root level fields like: `type`, `class`, `cluster_id` and make sure cloudbeat's events follow the specified guideline
- [ ] Fields that already exist prior 8.8, for backward compatibility, add to the ingest pipeline rules to remove these fields / move them to the proper place (only if they are not being used in `kibana`)
- [ ] Create follow-up for custom fields we want to be part of ECS
- [ ] Explore if there are more fields that weren't mentioned in DOD

**Out of scope**

- 

**Related tasks/epics**

-

Omolola-Akinleye:
@kfirpeled is additional work needed in Kibana,?  I remember the [bug fix ](https://github.com/elastic/kibana/pull/155206) for calculating the unique transform key.  `vulnerability.package.name`, `vulnerability.package.version`, `vulnerability.package.fix_version` are used in Kibana cc: @opauloh

uri-weisman:
> @kfirpeled is additional work needed in Kibana,? I remember the [bug fix ](https://github.com/elastic/kibana/pull/155206) for calculating the unique transform key. `vulnerability.package.name`, `vulnerability.package.version`, `vulnerability.package.fix_version` are used in Kibana cc: @opauloh

Yeah, I noticed that those fields are omitted from the findings table.
Another option is just adding the `package` object to the root level and keeping the old fields as is for BC.

kfirpeled:
> Another option is just adding the package object to the root level and keeping the old fields as is for BC.

Anything that is not defined as part of ECS should be under `cloud_security_posture.*` and not the root level. That is the new guidance and the purpose of this issue :) 

> @kfirpeled is additional work needed in Kibana,? I remember the https://github.com/elastic/kibana/pull/155206for calculating the unique transform key. vulnerability.package.name, vulnerability.package.version, vulnerability.package.fix_version are used in Kibana

That is a good question, and currently, it is still a middle ground. Custom fields under `vulnerability.*` are a subject for further discussion. If it fits to the ECS, we should keep them there and promote that discussion. If not, we should move them to `cloud_security_posture.*`. However, that is currently a lower priority and out of the scope of this issue.

[...]

PR #912, opened 2023-04-30
Update CNVM findings schema

uri-weisman:
### Summary of your changes
1. class -> vulnerability.class
2. target -> package.path
3. type -> package.type
4. Move vulnerability.package to be directly under package as defined in the technical index and in [ECS](https://www.elastic.co/guide/en/ecs/8.7/ecs-package.html).
5. Upgrade elastic-package binary to the latest.

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/897

### Checklist

Screenshot 2023-05-01 at 13 45 13
Screenshot 2023-05-01 at 13 45 35


amirbenun (approved):
1. Do we want to backport?
2. In https://github.com/elastic/cloudbeat/issues/897 there is also `cluster_id` do we want to change it on a different PR?

oren-zohar:
@kfirpeled do we want to backport this?

uri-weisman:
> 2. In [Move non-ECS fields from root level event #897](https://github.com/elastic/cloudbeat/issues/897) there is also `cluster_id` do we want to change it on a different PR?

@amirbenun - Feels like it does not relate to the CNVM flow as we don't send this field.
I'll open a follow-up task for removing the `cluster_id` field (it has already an alternative - `orchestrator.cluster.id`), it will be a bit more complicated as it is a breaking change.

kfirpeled:
> @kfirpeled do we want to backport this?

Yes, looking at the current implementation: 
- having package both in the root level and under vulnerability
- moving `type`, `target` and `class` from the root level

We agree to backport that to 8.8

cc: @uri-weisman

Issue #913, opened 2023-05-01
RDS fetcher does not fetch all potential databases


ari-aviran:
**Describe the bug**
The RDS fetcher only fetches the 100 first databases per region. [This](https://github.com/elastic/cloudbeat/blob/4042150a06e5e5d1a6869becf5d1e81818ef7332/resources/providers/awslib/rds/provider.go#L47) API call [is paginated](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DescribeDBInstances.html) but we only query the first page.

**Preconditions**
- AWS environment with over 100 RDS DB instances in a region (I think default quota is 40 but can be raised).
- Elastic stack with CSPM integration connected to the AWS environment.

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Setup the AWS environment and CSPM integration.
2. See that only 100 databases are inspected.

**Expected behavior**
All databases should be inspeceted.

olegsu:
This cannot be verified that easily, it will require the creation of 100+ db instances, which might take a lot of time.
Since this rule is not evaluated db clusters we cannot also create serverless aurora dbs.

The PR is following the other pagination that we use. 👍

PR #918, opened 2023-05-01
Add pagination in RDS provider to describe DB instances

orouz:
### Summary of your changes

adds pagination by using the output's marker property. 

### Related Issues
- fixes https://github.com/elastic/cloudbeat/issues/913 

### Checklist

### DIFF
--- a/resources/providers/awslib/rds/provider.go
+++ b/resources/providers/awslib/rds/provider.go
@@ -19,6 +19,7 @@ package rds
 
 import (
 	"context"
+
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/rds"
 	"github.com/aws/aws-sdk-go-v2/service/rds/types"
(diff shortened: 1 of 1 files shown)


resources/providers/awslib/rds/provider.go
```diff
@@ -44,16 +45,28 @@ func NewProvider(log *logp.Logger, cfg aws.Config, factory awslib.CrossRegionFac
...
+
+			dbInstancesInput.Marker = output.Marker
 		}
 
-		for _, dbInstance := range dbInstances.DBInstances {
+		for _, dbInstance := range dbInstances {
 			subnets, subnetsErr := p.getDBInstanceSubnets(ctx, region, dbInstance)
 			if subnetsErr != nil {
-				p.log.Errorf("Could not get DB instance subnets. DB: %s. Error: %v", *dbInstance.DBInstanceIdentifier, err)
+				p.log.Errorf("Could not get DB instance subnets. DB: %s. Error: %v", *dbInstance.DBInstanceIdentifier, subnetsErr)
```
orouz, line 69: i _think_ using `err` wasn't correct to begin with. previously `err` was set by  `c.DescribeDBInstances` and we already accounted for it by logging and returning it. 

so here logging `subnetsErr`  seems like what we want to do here.

orestisfl (approved):

resources/providers/awslib/rds/provider.go
```diff
@@ -44,16 +45,28 @@ func NewProvider(log *logp.Logger, cfg aws.Config, factory awslib.CrossRegionFac
...
+			if output.Marker == nil {
+				break
+			}
+
+			dbInstancesInput.Marker = output.Marker
 		}
 
-		for _, dbInstance := range dbInstances.DBInstances {
+		for _, dbInstance := range dbInstances {
 			subnets, subnetsErr := p.getDBInstanceSubnets(ctx, region, dbInstance)
```
orestisfl, line 67: I don't think we need another `err` variable here

```suggestion
			subnets, err := p.getDBInstanceSubnets(ctx, region, dbInstance)
```

Issue #928, opened 2023-05-02
Add more data regarding snapshot creation, scanning and encryption


uri-weisman:
**Motivation**
Add more data to our logs.
1. creation time of a snapshot.
2. Scanning time of a snapshot.
3. Is the snapshot encrypted?
4. Volume size

**Definition of done**
Logs contain the above data.

orestisfl:
Verified

![image](https://github.com/elastic/cloudbeat/assets/5778622/1412b2bc-9772-4d56-9ba4-cddbf2c13cfc)

Issue #740, opened 2023-02-13
Configure various timeouts like scanning timeout, waiting for snapshots timeout, and iteration times


jeniawhite:
Make sure that all of the timeouts across the feature make sense and are compliant with what we want to achieve.
Example of this is the snapshot verifier:

```go
func (f VulnerabilityVerifier) verify(ctx context.Context, snap types.SnapshotInfo) {
	f.log.Info("Starting VulnerabilityVerifier.verify")
	// TODO: Change this to variable
	timer := time.After(5 * time.Minute)
	for {
		select {
		case <-timer:
			f.log.Info("VulnerabilityVerifier.verify timed out waiting for snapshot ", *snap.SnapshotId)
			return
		case <-ctx.Done():
			f.log.Info("VulnerabilityVerifier.verify context canceled")
			return
		// TODO: Change this to variable
		case <-time.After(15 * time.Second):
			sp, err := f.provider.DescribeSnapshots(ctx, snap)
			if err != nil {
				f.log.Errorf("VulnerabilityVerifier.verify.DescribeSnapshots failed: %v", err)
				continue
			}
			// TOOD: Add a layer of "smart" cache to avoid checking and sending the same snapshot
			// and not to wait on all snapshots to be completed, sending them periodically
			allCompleted := true
			for _, i := range sp {
				if i.State != types.SnapshotStateCompleted {
					f.log.Info("VulnerabilityVerifier.verify.VerifySnapshot snapshot not completed yet - ", *i.SnapshotId)
					allCompleted = false
					break
				}
			}
			if allCompleted {
				for _, i := range sp {
					f.ch <- i
				}
				return
			}
		}
	}
}
```

You can see comments on timeouts like `// TODO: Change this to variable`.
Another place is the scanning of the EBS snapshots itself by `Trivy` this is also capped:

```go
		GlobalOptions: flag.GlobalOptions{
			// TODO: Make configurable
			Timeout: 1 * time.Hour,
			Quiet:   false,
			Debug:   true,
			// CacheDir: "/tmp/trivy",
		},
```
In this case comment on timeout `// TODO: Make configurable`.
We would like to evaluate the times there and make sure that they are making sense.

uri-weisman:
I have generated a report that includes data on snapshot creation time and scanning.
I ignored the data on encrypted snapshots that we do not currently support, it appears that we are experiencing an issue with the current snapshot creation timeout, which is set at **10 mins**. 
[...]

PR #930, opened 2023-05-03
[CNVM] Update snapshot creation timeout and add logs

uri-weisman:
### Summary of your changes
1. Update snapshot creation timeout to 30 mins
2. Add more informative logs to include this specific data.

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/928
- Related: https://github.com/elastic/cloudbeat/issues/740

### Checklist


vulnerability/scanner.go
```diff
@@ -126,14 +126,21 @@ func (f VulnerabilityScanner) scan(ctx context.Context, snap ec2.EBSSnapshot) {
 			Severities: []db_types.Severity{0, 1, 2, 3, 4},
 		},
 	}
-
+	now := time.Now()
 	report, err := f.runner.ScanVM(ctx, opts)
+	f.log.Debugf(
```
oren-zohar, line 131: should we change it to debug? will we have any indication that snapshots were scanned?

oren-zohar (changes requested):

vulnerability/scanner.go
uri-weisman: Didn't want to spam the logs with information that will be relevant only when we want to debug.
If you think it's critical to the user I'll change to info.

oren-zohar: @jeniawhite Since you initially set it to `info`, what are your thought about this, will it spam the logs? how frequently will we see this log? Is it necessary for the first step of debugging (before asking users to switch log levels)

uri-weisman: The start scanning log is Info level (in the beginning of the func), so either both will be info or debug.

uri-weisman: Changed to info.
I changed my opinion it will be helpful to the user.

oren-zohar (approved):

PR #941, opened 2023-05-07
Pipeline test infra

gurevichdmitry:
### Summary of your changes

This PR adds new ci sanity workflow that currently supports demand execution on main (post-merge).

This flow contains two input params:
- elk-stack-version - ELK version to be deployed (required)
- ess-region - Elastic Cloud region (required)

Flow actions:
- Deploy Elactic Cloud
- Save terraform state
- Destroy Elastic Cloud

Terraform state is saved to AWS S3 bucket. It is just a backup for the case when deployment occurred but the stack wasn't destroyed. In order to be able to destroy  the stack in conventional way we need to be able to reach out the terraform state, that's why the state is saved to bucket.

@oren-zohar, this flow requires additional secret to be added `CI_EC_API_KEY`. To be able to deploy to Elastic cloud we need to provide `TF_VAR_ec_api_key`. The sanity flow maps environment variable `CI_EC_API_KEY` to `secrets.TF_VAR_ec_api_key`


.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,102 @@
+name: sanity workflow
+
+on:
+  # Ability to execute on demand
+  workflow_dispatch:
+    inputs:
+      elk-stack-version:
+        required: true
+        description: "Elastic Cloud stack version"
+        default: "8.8.0-SNAPSHOT"
```
amirbenun, line 10: I think you should find it by running `VERSION=$(grep defaultBeatVersion version/version.go | cut -f2 -d "\"")` to avoid maintaining this variable manually.

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,102 @@
+name: sanity workflow
+
+on:
+  # Ability to execute on demand
+  workflow_dispatch:
```
amirbenun, line 5: Are we going to change it later on?

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,102 @@
...
+  Setup:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 15
+    outputs:
+      stack-version: ${{ steps.set-elk-vars.outputs.stack-version }}
+      ess-region: ${{ steps.set-elk-vars.outputs.ess-region }}
+      tf-state: ${{ steps.gen-tf-name.outputs.tf-file}}
+    steps:
+      - id: set-elk-vars
+        if: github.event_name == 'workflow_dispatch'
```
amirbenun, line 34: Why do we need that if anyway that's the only trigger?

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,102 @@
...
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_REGION: "eu-west-3"
+  TF_VAR_ec_api_key: ${{ secrets.CI_EC_API_KEY }}
+  EC_DEPLOYMENT: "module.ec_deployment"
+  WORKING_DIR: deploy/cloud
+
+jobs:
+  Setup:
```
amirbenun, line 25: Seems like we don't do anything on `Setup` besides outputting variables, in that case we can have a single job.

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,102 @@
...
+        id: apply
+        run: |
+          terraform apply --auto-approve \
+            -target ${{ env.EC_DEPLOYMENT }} \
+            -state=${{ env.TF_STATE_FILE}}
+
+      - name: Upload tf state
+        if: success()
+        run: |
+          aws s3 cp "${{ env.TF_STATE_FILE }}" "s3://tf-state-bucket-test-infra/${{ env.TF_STATE_FILE }}"
```
amirbenun, line 87: Can you elaborate why do we need this file in S3? We should be able to list those environments by a naming convention. Also the cleanup step always runs, do you still worry about an edge case that will keep some environments up?

amirbenun (commented):
Do you think we should have a single environment for KSPM, CSPM and CNVM Or each one of them should a dedicated environment and a dedicated workflow?

amirbenun (commented):
Do you think we should have a single environment for KSPM, CSPM and CNVM Or each one of them should a dedicated environment and a dedicated workflow?

.github/workflows/test-environment.yml
gurevichdmitry: @amirbenun in current context,  `elk-stack-version` is the input parameter for this workflow and the default version string will be displayed in the UI:
Screenshot 2023-05-07 at 17 26 05

Setup job parses params and setup appropriate output vars. This parameter is defined by the user that triggers this workflow.
Your suggestion is relevant for scheduled workflow, when you need to get current latest version. But this option will be added later.

.github/workflows/test-environment.yml
gurevichdmitry: We are going to add also scheduled option when the workflow will be executed automatically.

.github/workflows/test-environment.yml
gurevichdmitry: There are several use cases for this workflow.
1. Weekly sanity checks -> in this case we need to provision infrastructure -> execute tests -> destroy infrastructure.
2. Provision infrastructure for some reasons: before QA cycle, once a week, etc. in this case destroy infrastructure is not relevant because you need this environment for manual testing or investigation of something and therefore you should decide when to destroy. To do that we need to save terraform state file.

.github/workflows/test-environment.yml
gurevichdmitry: `Setup` contains only outputting variables because it is an initial stage of workflow development. Adding scheduler will add a logic for all variables and in this case a separate `Setup` job will be better than to have a single job. In addition this flow will manage also deployment of KSPM and CSPM components.  Defining logic of what should be deployed in a separate `Setup` job is preferable than a single job for all.

.github/workflows/test-environment.yml
gurevichdmitry: You are right, for now it can be removed.

gurevichdmitry:
> Do you think we should have a single environment for KSPM, CSPM and CNVM Or each one of them should a dedicated environment and a dedicated workflow?

I think we should have a single workflow for all deployments and the logic of what should be deployed will be managed through job actions configuration

deploy/cloud/modules/provision-eks-cluster/variables.tf
```diff
@@ -8,3 +8,21 @@ variable "cluster_name" {
   description = "EKS cluster name"
   type        = string
 }
+
+variable "node_group_one_desired_size" {
+  default     = 3
```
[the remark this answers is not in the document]
gurevichdmitry: I will update the default value of desired_size to 1 and change the nodes size to 3 in the main cloud EKS module. Please note that the nodes count is important in the live-track environment as it affects the number of findings and the configuration of environment alerts.

deploy/test-environments/README.md
```diff
@@ -0,0 +1,96 @@
...
+Follow the [prerequisites](/README.md#prerequisites) chapter of our main README.
+
+## Environment Variables
+
+To successfully deploy the environment, ensure that the following variables are provided as deployment parameters or exported as environment variables:
+
+```bash
+export TF_VAR_ec_api_key={TOKEN} # <-- should be replaced by Elastic Cloud TOKEN
+export TF_VAR_stack_version=8.6.0-SNAPSHOT
+export TF_VAR_ess_region=gcp-us-west2
```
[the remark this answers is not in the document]
gurevichdmitry: updated

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,127 @@
+name: sanity workflow
```
oren-zohar, line 1: Just for consistency 
```suggestion
name: Sanity tests
```

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,127 @@
...
+    inputs:
+      deployment_name:
+        type: string
+        description: |
+          Name your environment (Only a-zA-Z0-9 and `-`).
+          For example: john-8-7-2-June01'
+        required: true
+      elk-stack-version:
+        required: true
+        description: "Elastic Cloud stack version"
```
oren-zohar, line 15: maybe add a note regarding BC

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,127 @@
...
+          aws s3 cp "${{ env.EC2_CSPM_KEY }}" "${{ env.S3_BUCKET }}/cspm.pem"
+          aws s3 cp "${{ env.EC2_KSPM_KEY }}" "${{ env.S3_BUCKET }}/kspm.pem"
+
+      # Placeholder action that need to be replaced by the following:
+      # Additional deployment EC2 / K8s
+      # # Tests execution
+      # - name: Sleep action
+      #   if: success()
+      #   run: |
+      #     sleep 300
```
oren-zohar, line 122: update on this?

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,127 @@
...
+
+jobs:
+  Setup:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 15
+    outputs:
+      stack-version: ${{ steps.set-input-vars.outputs.stack-version }}
+      ess-region: ${{ steps.set-input-vars.outputs.ess-region }}
+      deployment-name: ${{ steps.set-input-vars.outputs.deployment-name }}
+      tf-folder-date: ${{ steps.gen-tf-name.outputs.tf-folder-date}}
```
oren-zohar, line 40: I don't understand this part, why do we need it?

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1-node16
+        with:
+          aws-access-key-id: ${{ env.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ env.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ env.AWS_REGION }}
+
+      - name: Terraform Init
+        run: terraform init -no-color
```
oren-zohar, line 85: why no color? :smile:

deploy/cloud/variables.tf
```diff
@@ -68,6 +68,19 @@ variable "eks_region" {
   type        = string
 }
 
+variable "node_group_one_desired_size" {
+  default     = 3
```
oren-zohar, line 72: 2?

deploy/cloud/variables.tf
```diff
@@ -68,6 +68,19 @@ variable "eks_region" {
...
+  type        = number
+  description = "Node group one default size"
+}
+
+variable "node_group_two_desired_size" {
+  default     = 3
```
oren-zohar, line 78: 2?

oren-zohar (changes requested):

.github/workflows/test-environment.yml
gurevichdmitry: Alright, I will remove the Setup job and merge the input variables into the Deploy job.

deploy/cloud/variables.tf
gurevichdmitry: updated default value.

deploy/cloud/variables.tf
gurevichdmitry: default value is updated to 2.

.github/workflows/test-environment.yml
gurevichdmitry: When running Terraform as part of a CI/CD pipeline, colorized output is typically not necessary. However, if you prefer colorization for better readability, I can update the command to include it.

.github/workflows/test-environment.yml
gurevichdmitry: updated

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,127 @@
...
+        description: "Elastic Cloud deployment region"
+        default: "gcp-us-west2"
+      cleanup-env:
+        description: "Boolean: Cleanup resources after provision"
+        default: false
+
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_REGION: "eu-west-3"
```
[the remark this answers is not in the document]
oren-zohar: Why would it be a good idea to use another region? I think it would make it easy for us to miss unutilized resources when we perform clean-ups. IMO we should try and stick to one region as where ever it is possible

.github/workflows/test-environment.yml
oren-zohar: yes please

oren-zohar (changes requested):

.github/workflows/test-environment.yml
gurevichdmitry: terraform commands were updated

.github/workflows/test-environment.yml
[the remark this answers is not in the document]
gurevichdmitry: I just want to reduce false positive behavior. Human errors will always occur. Since we are allowing the environment to remain uncleared in this job, we can replicate the current state as it exists today in eu-west-1, where we have multiple deployed EKS clusters and EC2 machines. The main purpose of this job is to ensure stability over time, and one crucial aspect is to use a separate region.

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,223 @@
...
+    inputs:
+      deployment_name:
+        type: string
+        description: |
+          Name your environment (Only a-zA-Z0-9 and `-`).
+          For example: john-8-7-2-June01'
+        required: true
+      ec-api-key:
+        type: string
+        description: "Elastic Cloud API KEY"
```
oren-zohar, line 15: Can we have hyperlinks here? Getting the API is not so obvious in the cloud UI, maybe we can attach the relevant link (https://cloud.elastic.co/deployment-features/keys)

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,223 @@
...
+          summary=$(cat <<-EOF
+          Kibana URL: [kibana]($kibana_url)
+          Environment Details: $s3_bucket_link
+          EOF
+          )
+          echo "$summary" >> $GITHUB_STEP_SUMMARY
+          echo "$summary" # Print the summary to the workflow log
+
+      - name: Cleanup Environment
+        if: github.event.inputs.cleanup-env == 'true'
```
oren-zohar, line 221: Do you think it makes sense to also do the cleanup if some of the stages failed?

deploy/test-environments/README.md
```diff
@@ -0,0 +1,96 @@
...
+This project utilizes AWS and Elastic Cloud accounts. To ensure proper deployment and usage, it is essential to obtain appropriate licenses in compliance with the licensing terms and conditions provided by the respective service providers.
+
+Follow the [prerequisites](/README.md#prerequisites) chapter of our main README.
+
+## Environment Variables
+
+To successfully deploy the environment, ensure that the following variables are provided as deployment parameters or exported as environment variables:
+
+```bash
+export TF_VAR_ec_api_key={TOKEN} # <-- should be replaced by Elastic Cloud TOKEN
```
oren-zohar, line 18: Can you add the link somewhere (https://cloud.elastic.co/deployment-features/keys)

oren-zohar (changes requested):
Looks really good! Can you please add a documentation file about the sanity workflow itself? Maybe under `/dev-docs`. And please include a section about how to use credentials

.github/workflows/test-environment.yml
gurevichdmitry: I've attempted to add a hyperlink, but it is displayed as plain text instead. As a result, I have made updates to the documentation in the README file to provide detailed instructions on how to generate an Elastic Cloud API key.

.github/workflows/test-environment.yml
gurevichdmitry: Developing the logic for that might be a bit complicated. It depends on the nature of the failure—if the failed steps are unrelated to resource provisioning, it might not be necessary to perform cleanup. In such cases, the environment itself might be in good condition, and the issue could be related to the installation of integrations or the execution of tests. Personally, I believe it's better to start with a simpler approach and improve the action as we encounter specific problems along the way.

deploy/test-environments/README.md
gurevichdmitry: readme file is updated

.github/workflows/test-environment.yml
oren-zohar: would doing a cleanup anyway could lead to any issues? if not, why not run it on every failure?

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,223 @@
...
+      ess-region:
+        required: true
+        description: "Elastic Cloud deployment region"
+        default: "gcp-us-west2"
+      docker-image-override:
+        required: false
+        description: "Provide the full Docker image path to override the default image (BC versions only)"
+      cleanup-env:
+        description: "Boolean: Cleanup resources after provision"
+        default: false
```
oren-zohar, line 30: we can use the type `boolean` here instead of specifying it's a boolean, no?

oren-zohar (changes requested):

.github/workflows/test-environment.yml
gurevichdmitry: Running a cleanup process on every failure may not be necessary or relevant in certain cases. For example, if the environment provisioning was successful but there were issues with installing agent policies or other components, performing a cleanup might not be the appropriate action.

oren-zohar (approved):
Approved. Let's sync regarding providing better accessibility to the deployed Kibana instance, I think we may did a bit of overkill from a security perspective, since we assume now it's part of the user cloud env

kfirpeled:
Hi @gurevichdmitry
I like it a lot, great job!

1 thing to keep in mind, when someone chose not to kill the environment - how do we destroy it later on? I guess we will need another job for that

.github/workflows/test-environment.yml
```diff
@@ -0,0 +1,224 @@
...
+      - name: Init Hermit
+        run: ./bin/hermit env -r >> $GITHUB_ENV
+        working-directory: ./
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+
+      - name: Install Poetry
```
amirbenun, line 69: Add it to hermit instead?

gurevichdmitry: I couldn't find a suitable Hermit package for Poetry. However, the [Poetry documentation](https://python-poetry.org/docs/#ci-recommendations) suggests using the official installer.

PR #949, opened 2023-05-08
Move shutdown grace period to Launcher

### LINKED ISSUE #783: Cloudbeat does not stop when sending SIGINT
**Describe the bug**
Cloudbeat does not stop when we send a SIGINT signal (CTRL + C).

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Run cloudbeat as a binary `./cloudbeat -c "deploy/aws/cloudbeat-aws.yml"`.
2. Try and stop the program using CTRL + C.

**Expected behavior**
The program should shutdown gracefully when a SIGINT signal is sent.
When hitting CTRL+C second time we should exit

**Additional context**
 - OS: darwin

### LINKED ISSUE #905: Cloudbeat panics after graceful shutdown trying to close a closed channel
**Describe the bug**

after completing the graceful shutdown process triggered with CTRL+C (SIGINT), cloudbeat has a panic runtime error:

```
{"log.level":"fatal","@timestamp":"2023-04-27T16:38:20.533+0300","log.logger":"cloudbeat","log.origin":{"file.name":"instance/beat.go","file.line":206},"message":"Failed due to panic.","service.name":"cloudbeat","error":{"message":"close of closed channel"},"stack":"github.com/elastic/beats/v7/libbeat/cmd/instance.Run.func1.1\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/elastic/beats/v7/libbeat/cmd/instance/beat.go:207\nruntime.gopanic\n\t/Users/orouz/Library/Caches/hermit/pkg/go-1.19.6/src/runtime/panic.go:890\nruntime.closechan\n\t/Users/orouz/Library/Caches/hermit/pkg/go-1.19.6/src/runtime/chan.go:365\ngithub.com/elastic/cloudbeat/launcher.(*launcher).Stop\n\t/Users/orouz/Documents/projects/cloudbeat/launcher/launcher.go:173\ngithub.com/elastic/beats/v7/libbeat/management.(*fallbackManager).Stop.func1\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/elastic/beats/v7/libbeat/management/management.go:187\nsync.(*Once).doSlow\n\t/Users/orouz/Library/Caches/hermit/pkg/go-1.19.6/src/sync/once.go:74\nsync.(*Once).Do\n\t/Users/orouz/Library/Caches/hermit/pkg/go-1.19.6/src/sync/once.go:65\ngithub.com/elastic/beats/v7/libbeat/management.(*fallbackManager).Stop\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/elastic/beats/v7/libbeat/management/management.go:186\ngithub.com/elastic/cloudbeat/launcher.(*launcher).Run\n\t/Users/orouz/Documents/projects/cloudbeat/launcher/launcher.go:108\ngithub.com/elastic/beats/v7/libbeat/cmd/instance.(*Beat).launch\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/elastic/beats/v7/libbeat/cmd/instance/beat.go:522\ngithub.com/elastic/beats/v7/libbeat/cmd/instance.Run.func1\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/elastic/beats/v7/libbeat/cmd/instance/beat.go:215\ngithub.com/elastic/beats/v7/libbeat/cmd/instance.Run\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/elastic/beats/v7/libbeat/cmd/instance/beat.go:216\ngithub.com/elastic/beats/v7/libbeat/cmd.genRunCmd.func1\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/elastic/beats/v7/libbeat/cmd/run.go:36\ngithub.com/spf13/cobra.(*Command).execute\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/spf13/cobra/command.go:920\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/spf13/cobra/command.go:1040\ngithub.com/spf13/cobra.(*Command).Execute\n\t/Users/orouz/Documents/projects/cloudbeat/vendor/github.com/spf13/cobra/command.go:968\nmain.main\n\t/Users/orouz/Documents/projects/cloudbeat/main.go:29\nruntime.main\n\t/Users/orouz/Library/Caches/hermit/pkg/go-1.19.6/src/runtime/proc.go:250","ecs.version":"1.6.0"}
```

**Preconditions**

cloudbeat `57143d4199af1c6e9c569362a1506521202a31e5`

**To Reproduce**

1. run cloudbeat: `./cloudbeat -c "deploy/aws/cloudbeat-aws.yml" --path.logs logs -d -e -d "service"`
2. hit CTRL+C
3. check log at `logs/<file>.ndjson`
4. last log is `Failed due to panic.`

**Expected behavior**

cloudbeat shouldn't panic

orouz:
### Summary of your changes

1 - call `Manager.Stop()` only when cloudbeat is not running standalone (otherwise our `Stop()` function gets called twice, which panics trying to `close(l.beaterErr)`)

2 - moves the grace period mechanism to the launcher. we now cancel the beater (posture) main context (bt.ctx) once a SIGINT is sent, resulting in exiting almost immediately. if for some reason the beater didn't exit, we will exit after the defined grace period (which i changed to `5s`, mostly because of tests)

### Related Issues
- fixes #905
- fixes #783

### Checklist


resources/fetchersManager/data.go
```diff
@@ -57,19 +58,21 @@ func NewData(log *logp.Logger, interval time.Duration, timeout time.Duration, fe
 
 // Run starts all configured fetchers to collect resources.
 func (d *Data) Run(ctx context.Context) Stop {
-	go d.fetchAndSleep(ctx)
+	fetchCtx, cancel := context.WithCancel(ctx)
+	go d.fetchAndSleep(fetchCtx, cancel)
```
orouz, line 62: lifted the context, making it used by fetchers and also the grace period timeout context. 

so cloudbeat will shutdown once grace period timeout context is done, which will happen:
- once the timeout has exceeded, or 
- when its parent context was canceled

whichever comes first. 

to get an estimate what will come first, we need to look at where we return from `fetchAndSleep`, as that is when the context is canceled.

we return in 3 places: 
https://github.com/elastic/cloudbeat/blob/c1d27fee3ebfc69cbf85c9a8570041e413f4c64b/resources/fetchersManager/data.go#L85-L93
 
 - stop won't get called until the grace period context is done , won't run
 - the context won't be done until we exit fetchAndSleep, won't run (can be deleted?)
 - when the next interval will run after already receiving a notice, it will return and then cancel the shared context 
 
 so the shorter the fetch period and the higher grace period, the more likely we are to exit **before** the grace period ended.

resources/fetchersManager/data.go
```diff
@@ -165,9 +168,18 @@ func (d *Data) fetchProtected(ctx context.Context, k string, metadata fetching.C
...
 	<-ctx.Done()
 	d.fetchers.Stop()
 	close(d.stop)
+
+	if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+		d.log.Debug("Context timeout exceeded")
+	}
+	if errors.Is(ctx.Err(), context.Canceled) {
+		d.log.Debug("Context canceled", time.Since(startTime).String())
+	}
```
orouz, line 184: you can test this branch by plugging the periods from the table in the description and looking at the logs to find the relevant line.

launcher/launcher.go
```diff
@@ -81,10 +81,12 @@ func (l *launcher) Run(b *beat.Beat) error {
 	if err := b.Manager.Start(); err != nil {
 		return err
 	}
-	defer b.Manager.Stop()
 
 	// Wait for Fleet-side reconfiguration only if beater is running in Agent-managed mode.
 	if b.Manager.Enabled() {
+
+		defer b.Manager.Stop()
+
```
orouz, line 89: this needs to run only when we're in managed mode otherwise it will call our `Stop()` function again which will result in a panic.

resources/fetchersManager/data.go
```diff
@@ -50,68 +57,87 @@ func NewData(log *logp.Logger, interval time.Duration, timeout time.Duration, fe
...
 		interval:   interval,
 		fetchers:   fetchers,
-		stop:       make(chan struct{}),
-		stopNotice: make(chan time.Duration),
+		stopNotice: make(chan time.Duration, 1),
 	}, nil
 }
 
+var ErrGracePeriodTooLong = errors.New("graceful shutdown too long")
+var ErrGracePeriodEnded = errors.New("graceful shutdown ended")
```
orouz, line 65: depending on why we exit, we cancel with a different `cause`. this is needed for tests to assert on why we exited. otherwise we wouldn't bother with it.

resources/fetchersManager/data_test.go
```diff
@@ -307,60 +316,43 @@ func (s *DataTestSuite) TestDataStopWithGracefulShutdown() {
...
-
-	stop := d.Run(ctx)
-	time.Sleep(2 * time.Second)
-
-	stop(ctx, time.Second)
-	time.Sleep(2 * time.Second)
-	s.LessOrEqual(f.execCounter, 4)
-}
-
-func (s *DataTestSuite) TestDataDoubleStop() {
```
orouz, line 347: removed as this test does not include any assertions

resources/fetchersManager/data.go
```diff
@@ -50,68 +57,87 @@ func NewData(log *logp.Logger, interval time.Duration, timeout time.Duration, fe
...
+
+		case grace := <-d.stopNotice:
+			fmt.Printf("shutdown grace period started. will exit after %s or next iteration \n", grace.String())
+			exitNextIteration.Store(true)
+
+		case <-interval.C:
+			if exitNextIteration.Load() {
+				fmt.Printf("exit new fetch iteration\n")
+				cancel(ErrGracePeriodTooLong)
 				return
```
orouz, line 118: this is essentially our mechanism for shutting down before the grace period ended. 
it depends on the interval period, which in real envs is in mins/hrs, so really we'll only exit after waiting the full grace period.

resources/fetchersManager/data.go
```diff
@@ -19,27 +19,34 @@ package fetchersManager
...
+	// Interval between two consecutive cycles
+	interval time.Duration
+
+	fetchers FetchersRegistry
+
+	// Maximum timeout for graceful shutdown
 	stopNotice chan time.Duration
 }
 
-type Stop func(context.Context, time.Duration)
```
orouz, line 42: `Stop` creates its own `WithCancel` context, derived from the `bt.ctx` passed to `Run()`, which is the same context we used to pass here.

launcher/launcher.go
```diff
@@ -81,10 +84,28 @@ func (l *launcher) Run(b *beat.Beat) error {
...
+			for range sigc {
+				sigCount++
+				if sigCount >= 2 {
+					l.log.Infof("Received second signal, terminating immediately")
+					os.Exit(1)
+				}
+			}
+
+		}()
+	}
```
orouz, line 102: this is intended to run only on standalone cloudbeat and enable support for exiting immediately after a double SIGINT 

note the we [won't](https://github.com/elastic/elastic-agent-libs/blob/main/service/service.go#L46-L55) be notified by a second signal

lmk if this should be kept in the PR or not

resources/fetchersManager/data.go
```diff
@@ -165,9 +166,16 @@ func (d *Data) fetchProtected(ctx context.Context, k string, metadata fetching.C
...
+		d.fetchers.Stop()
+		close(d.stop)
+	}()
+
+	select {
+	case <-time.After(grace):
+		return
+	case <-ctx.Done():
+		return
+	}
```
orouz, line 180: wait for grace period or context to be cancelled, whichever comes first. context will get canceled if a new fetch iteration started after a grace notice was received.

resources/fetchersManager/data.go
```diff
@@ -57,19 +57,20 @@ func NewData(log *logp.Logger, interval time.Duration, timeout time.Duration, fe
 
 // Run starts all configured fetchers to collect resources.
 func (d *Data) Run(ctx context.Context) Stop {
-	go d.fetchAndSleep(ctx)
+	fetchContext, cancel := context.WithCancel(ctx)
```
orouz, line 60: lifted the context to be shared with `d.stopData` so we can select a case to return from: `After(grace)` / `ctx.Done()`

flavors/posture.go
```diff
@@ -122,7 +121,7 @@ func (bt *posture) Run(b *beat.Beat) error {
 		return err
 	}
 
-	bt.dataStop = bt.data.Run(bt.ctx)
+	bt.data.Run(bt.ctx)
```
orouz, line 124: `Run` knows it needs to stop once `bt.ctx` is closed. didn't find a need to add a Stop function

flavors/posture.go
```diff
@@ -192,9 +191,7 @@ func initRegistry(log *logp.Logger, cfg *config.Config, ch chan fetching.Resourc
 
 // Stop stops posture.
 func (bt *posture) Stop() {
-	if bt.dataStop != nil {
-		bt.dataStop(bt.ctx, shutdownGracePeriod)
-	}
+	bt.data.Stop()
```
amirbenun, line 194: 👏

resources/fetchersManager/data.go
```diff
@@ -23,49 +23,40 @@ import (
...
-func (d *Data) Run(ctx context.Context) Stop {
+func (d *Data) Run(ctx context.Context) {
 	go d.fetchAndSleep(ctx)
-	called := atomic.NewBool(false)
-	return func(ctx context.Context, grace time.Duration) {
-		defer called.Store(true)
-		if called.Load() {
-			return
-		}
-		d.stopData(ctx, grace)
```
amirbenun, line 67: Can we really avoid all of that? @olegsu

launcher/launcher.go
```diff
@@ -193,11 +208,34 @@ func (l *launcher) stopBeater() {
 	l.log.Infof("Launcher shut %s down gracefully", l.name)
 }
 
-// waitForUpdates is the function that keeps Launcher runLoop busy
+// Returns an error indicating if the beater was stopped gracefully or not
+func (l *launcher) stopBeaterWithTimeout(duration time.Duration) error {
```
amirbenun, line 212: Love it

resources/fetchersManager/data.go
```diff
@@ -23,49 +23,40 @@ import (
...
-	}
+}
+
+func (d *Data) Stop() {
+	d.fetchers.Stop()
```
amirbenun, line 59: It seems like calling the `Stop` by itself, won't cause the `fetchAndSleep` to return (and it should).

resources/fetchersManager/data_test.go
```diff
@@ -136,8 +136,10 @@ func (s *DataTestSuite) TestDataRun() {
 	d, err := NewData(s.log, interval, timeout, s.registry)
 	s.NoError(err)
 
-	stop := d.Run(s.ctx)
-	defer stop(context.Background(), time.Second)
+	ctx, cancel := context.WithCancel(s.ctx)
+	defer cancel()
```
amirbenun, line 140: All these `defer cancel()` should be replaced with `defer d.Stop()` except of a specific test that is focused on the context cancellation.

amirbenun (commented):
Really nice!

launcher/launcher.go
olegsu, line 212: Why use an error here?

resources/fetchersManager/data.go
```diff
@@ -23,75 +23,62 @@ import (
...
-		if called.Load() {
-			return
-		}
-		d.stopData(ctx, grace)
-	}
+func (d *Data) Run() {
+	go d.fetchAndSleep(d.ctx)
+}
+
+func (d *Data) Stop() {
```
olegsu, line 65: What happen if you call this function twice?

olegsu (commented):
Thank you @orouz, 
Looks really good!!
Do you mind attaching log examples to the PR? so we know what to expect later

orouz:
> Do you mind attaching log examples to the PR? so we know what to expect later

this diagram shows how the termination process flows: 

```mermaid
flowchart TD
libbeat -->|stops| Launcher
Launcher -->|stops| Posture
Posture -->|stops| DataManager
Posture -->|stops| Evaluator
DataManager-->|stops| Fetchers
``` 

and this one shows how it eventually exists:
```mermaid
flowchart TD
DataManager-->|cancels context and returns| Posture
Posture -->|cancels context and returns| Launcher
Launcher -->|returns ErrorStopBeaterTimeout | libbeat
Launcher -->|returns ErrorStopBeater| libbeat
libbeat --> exit

``` 


and this is a sampled, ordered log (cleaned, with comments):

beats got a signal, termination process started
```
"Received signal \"interrupt\", stopping" 
```

our stop func is called
```
"Launcher is about to shut down gracefully" 
```

we close `beaterErr`, triggering a stop signal update from `waitForUpdates`, which will call `stopBeaterWithTimeout`
```
"Launcher is shutting Cloudbeat down gracefully" 
```

which calls `l.beater.Stop()`,  which calls `bt.evaluator.Stop()`
```
"Stopping bundle loader."  // OPA log
"Stopping decision logger." // OPA log
```
 and  also calls`bt.data.Stop()`, which calls `d.fetchers.stop()`
```
"Fetcher for key \"aws-ec2-network\" stopped" 
"Fetcher for key \"aws-trail\" stopped" 
```

and also cancels local context, then returns to posture 
```
"Fetchers manager canceled" 
```

existing goroutines listening to context are canceled
```
"Error running fetcher for key aws-iam: fetcher aws-iam was canceled" 
```

after `l.beater.Stop()` call canceled `bt.ctx`, posture returns to launcher
```
"Posture context is done" 
```

launcher finished running cloudbeat
```
"Cloudbeat run has finished" 
```

We waited for the launcher wait group to finish, otherwise this log would be "Grace period for cloudbeat ended"
```
"Launcher shut Cloudbeat down gracefully" 
"Launcher stopped successfully"
"Listener is about to stop"
```


current iteration ended
```
"Manager finished waiting and sending data after 9131 milliseconds"
"Cycle 1686489885 resource fetching has ended"
```

all of these are from active goroutines running fetchers that were canceled, and are now returning errors
```
"Error getting bucket policy for bucket aws-cloudtrail-logs-704479110758-61552557: operation error S3: GetBucketPolicy, context canceled"
"fail to list inline policies for user: ari-cis-credentials-test, error: operation error IAM: ListUserPolicies, context canceled"
"Error fetching flow logs for VPC vpc-058b21b3bf0f435b0: operation error EC2: DescribeFlowLogs, context canceled"
"fail to list attached policies for user: ari-cis-credentials-test, error: operation error IAM: ListAttachedUserPolicies, context canceled"
"Error getting bucket ACL for bucket aws-cloudtrail-logs-704479110758-61552557: operation error S3: GetBucketAcl, context canceled"
"Could not get bucket location for bucket automation-reports-qa-632253853. Not describing this bucket. Error: operation error S3: GetBucketLocation, context canceled"
"Error getting bucket logging for bucket aws-cloudtrail-logs-704479110758-61552557: operation error S3: GetBucketLogging, context canceled"
"Error fetching flow logs for VPC vpc-6cb55a15: operation error EC2: DescribeFlowLogs, context canceled"
"fail to list inline policies for user: cis-aws-qa-cycle, error: operation error IAM: ListUserPolicies, context canceled"
"Could not get bucket location for bucket automation-reports-qa-632499135. Not describing this bucket. Error: operation error S3: GetBucketLocation, context canceled"
```

libbeat is terminating cloudbeat
```
"cloudbeat stopped." // logged from instance/beat.go
"Error fetching flow logs for VPC vpc-36a1394e: operation error EC2: DescribeFlowLogs, https response error StatusCode: 0, RequestID: , canceled, context canceled"
"Exiting: stop beater" // logged from instance/beat.go
"fail to list mfa device for user: cloudqueryPOC, error: operation error IAM: ListMFADevices, context canceled"
"Could not get bucket location for bucket automation-reports-qa-638729572. Not describing this bucket. Error: operation error S3: GetBucketLocation, context canceled"
```


i think logs show that the termination works, as in:
1. once a signal is received, context cancelation propogates until we exit. probably before the grace period, but at most after it. 
2. no new iterations are scheduled 


however, any current workloads being done on a goroutine (evaluation, fetching) will still run, even though most of it will be logs due to canceled context. to mitigate this too, as in - exit all workloads once a context is canceled, we'd need to periodically check in each workload if the context is done and return from there. don't think it's in-scope for this task though as this would be a significant refactor. these error logs are annoying but harmless.

one thing i'm not sure about is how could we get a log that says the cycle ended but then get errors from running fetchers, as if the wait group didn't wait for them too. 

wdyt? does the flow/logs make sense to you?

resources/fetchersManager/data.go
orouz: as discussed, this seems to work fine when called twice (no panics), but as per the beats docs, it will only be called [once](https://github.com/elastic/beats/blob/c4398dcf5f62ca4606eaf7d1dbf76635ecf76146/libbeat/beat/beat.go#L49-L510)

launcher/launcher.go
orouz: used to differentiate between returning with/without waiting for the launcher's wait group. could use any pair values but an error seemed most explicit way in go.

olegsu:
Amazing @orouz! 
Showing the flow of the context when canceled is very useful.
As you mentioned, we dont have much to do with the logs that come from synchronous work, those log messages will come after the cancelation.

olegsu (approved):

launcher/launcher.go
```diff
@@ -193,11 +208,34 @@ func (l *launcher) stopBeater() {
...
 // It will finish for one of following reasons:
-// 1. The Stop function got called
-// 2. The beater run has returned
-// 3. A config update received
+//  1. The Stop function got called 	(nil, nil)
+//  2. The beater run has returned 		(nil, err)
+//  3. A config update received 		(val, nil)
 func (l *launcher) waitForUpdates() (*config.C, error) {
 	select {
 	case err := <-l.beaterErr:
```
uri-weisman, line 241: @orouz
wdyt about indicating that a stop function has been called explicitly?
```suggestion
	case err, ok := <-l.beaterErr:
	 if !ok {
	     return nil, fmt.errorf("close channel/stop function has been called")
	 }
```

Issue #596, opened 2022-12-21
CloudFormation for CIS AWS


oren-zohar:
**Motivation**
Having CloudFormation script will ease our integration onboarding. Similar to vulnerability 

**Defeintion of Done** 
- [x] Create cloudformation script
- [x] Update to S3 from the CI
- [ ] Update CSPM integration with new variable with the link "quick-creation"
- [x] Add new deployment method to Kibana - Open task for Kibana team

**Related:** 
- https://github.com/elastic/security-team/issues/5659
- https://github.com/elastic/sre/issues/1480
- https://github.com/elastic/security-team/issues/5637

amirbenun:
With [CloudFormation mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html) we can have a single template that suits for many use cases based on parameters.
I would consider it as an option instead of maintaining multiple elastic-agent templates that are 90% similar. 
In this specific case, CSPM and CNVM templates should be identical except of the elastic-agent role.
I would try to have a map that contain the two roles and we dynamically create the relevant role based on the installed integration.

tehilashn:
@orestisfl - why is this task blocked?

tehilashn:
Thanks Orestis. 
So if I understand correctly, Cloudbeat's part is done and we're now waiting for Kibana. 
In that case, I would close this task (and remove " Update CSPM integration with new variable with the link "quick-creation" from the DoD), since we have a task to represent this on Kibana's side.

orouz:
- [ ] Create cloudformation script
 `https://github.com/elastic/cloudbeat/tree/main/deploy/cloudformation`

- [ ]  Update to S3 from the CI
 `https://github.com/elastic/cloudbeat/blob/main/.github/workflows/publish-cloudformation.yml`
 
- [ ]  Update CSPM integration with new variable with the link "quick-creation"
https://github.com/elastic/integrations/commit/e938366b876917271194c039d1b2f95b5338433c

- [ ]  Add new deployment method to Kibana 
   - verified in https://github.com/elastic/kibana/issues/157761

PR #956, opened 2023-05-09
cloudformation: Change dev deployment script

orestisfl:
# Summary of your changes

## Change CloudFormation template to conditionally enable ssh access depending on the stack name.

For example, the stack name `vulnerability-management` will create a stack where the EC2 machine is *not* accessible by ssh but the stack `developeraccess-vulnerability-management` will allow SSH access.

This PR completely removes the goformation-based [1] dev script that instead generated a temporary CloudFormation template with the SSH access disabled. The reason is that this script seems to block a map-based approach for #596: we want to conditionally change the `ElasticAgentRole.Properties.Policies` field depending on whether we are deploying the CSPM or vulnerability management integration. However, if a `Fn::FindInMap` is used under the `Policies` field, goformation will transform it into a base64-encoded string and then fail to marshal it into an `iam.Policy` [2].

Potential alternative solutions for this problem are:
1. Keep two separate CloudFormation templates, without the use of `Fn::FindInMap` and keep the script. Cons are:
    - WET code that can easily lead to mistakes
    - The current script lead to blockers for an issue already, this means that it could pose a problem in future issues as well
2. Include a new CloudFormation "DevAccess" parameter. Cons are:
    - We are exposing internal development options to our users, making on-boarding more difficult. Because of CloudFormation limitations, this parameter cannot be hidden.
3. Use a templating language like jinja [3] or envsubst to generate separate "production" and "dev" CloudFormation templates. Cons are:
    - Makes developing process more complicated because the original source file is not a valid CloudFormation file anymore and any linter checks will need to run on the generated files.
    - The original CloudFormation template cannot be directly used anymore but has to be generated.
4. Dynamically reference [4] an external value using AWS Systems Manager Parameter Store. Syntax: `{{resolve:service-name:reference-key}}`. Cons are:
    - We have to hardcode a reference to a resource in our template
    - Unclear what happens when the resource does not exist (e.g. in a client's account)
    - Unclear if this will work with `Fn::If`
5. Apply the "production" CloudFormation template and then add a script that modifies the EC2 instance after creation to enable SSH access. Cons are:
    - It's not easy to enable SSH access post-creation and there is no standard way to do it. I worked on this for a while and the script needed at minimum to:
        1. Fetch the instance-id from the cloudformation stack
        2. Fetch and decode the user-data (script that runs on the EC2 instance's startup)
        3. Fetch the public part of the key we want to use for SSH
        4. Stop the EC2 instance
        5. Modify & re-encode the user-data to append the public key to ~/.ssh/authorized_hosts
            - Potentially also the ssh agent needs to be installed / enabled
        6. Modify the security group to allow SSH traffic
        7. Start the instance again
6. Write & commit a `.patch` diff that adds ssh access and apply it to the existing CloudFormation template. Cons are:
    - Maintaining a diff file is probably painful
7. Use CloudFormation Macros [5] to do more complicate processing on our template. Cons are:
    - This will complicate our template even for the non-developer use-case
    - AFAICT, this involves deploying an extra lambda function
8. (Previous approach of this PR) In the CloudFormation template, refer to the `AWS::StackName` pseudo-parameter [6] and if it begins with a specified substring (here `developeraccess`), change the security group to allow access and add a key to the EC2 instance. Cons are:
    - Uses a "magic" string
    - This is still visible to our users if they inspect the template but it doesn't show up in the parameter list and shouldn't complicate on-boarding.
    - The key name *must* be the same as the CloudFormation stack name as there is no viable way of passing additional data in CloudFormation without exposing a configuration option to the user.
        - I have, however, added a convenience script that should take care of creating a key if one is needed
9. (Current approach) use [`yq`](https://github.com/mikefarah/yq) to modify the cloudformation yaml, instead of `goformation`. Not visible to prod users and backwards-compatible with what we have now.

[1]: https://github.com/awslabs/goformation
[2]: https://github.com/awslabs/goformation/blob/81751d04fca47e979bd53b7ca6d61261acf7bedc/cloudformation/iam/aws-iam-policy.go#L14-L55
[3]: https://jinja.palletsprojects.com/en/
[4]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html
[5]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html
[6]: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/596

### Checklist


amirbenun:
Thanks for the detailed comment @orestisfl !
1. Personally, I'm not a fan of implicit solutions, especially when it regards to developer tools. I would prefer to see and understand every single parameter that is relevant. In that case, we have a weird coupling between the stack name and the `allowSSh` (which doesn't exist anymore) and it becomes very implicit where you auto generate an ssh key in case it doesn't have one that match. IMO, we better stop using the goformation lib and just modify the CloudFormation template as a yaml file or even as a text file.
2. There is a plan for having a `ccli` tool for developers, implemented in golang, that will have all the shortcuts and developer commands that we need. If we go with that direction, maybe we can ask chatGPT to translate it to golang?

orestisfl:
Thanks for the feedback @amirbenun 
 
>     1. Personally, I'm not a fan of implicit solutions, especially when it regards to developer tools. I would prefer to see and understand every single parameter that is relevant. In that case, we have a weird coupling between the stack name and the `allowSSh` (which doesn't exist anymore) and it becomes very implicit where you auto generate an ssh key in case it doesn't have one that match. IMO, we better stop using the goformation lib and just modify the CloudFormation template as a yaml file or even as a text file.

I agree. One reason I didn't pursue this was that I was using the wrong "yq" app: https://github.com/kislyuk/yq instead of https://github.com/mikefarah/yq. I have pushed a new commit that should address your feedback.

>     2. There is a plan for having a `ccli` tool for developers, implemented in golang, that will have all the shortcuts and developer commands that we need. If we go with that direction, maybe we can ask chatGPT to translate it to golang?

I am not sure what this `ccli` tool you are referring to is.

deploy/cloudformation/gomain.go
```diff
@@ -98,11 +152,11 @@ func createStack(stackName string, templatePath string, params map[string]string
 	if err != nil {
 		return fmt.Errorf("failed to open template file: %v", err)
 	}
-	filestring := string(file)
+	fileName := string(file)
```
amirbenun, line 155: This is actually the file content and not the name

.pre-commit-config.yaml
```diff
@@ -52,8 +54,17 @@ repos:
           --diff,
         ]
 
+  - repo: local
```
amirbenun, line 57: Why is this section necessary? We check that the template is a valid cloudformation template, isn't that enough?

.pre-commit-config.yaml
```diff
@@ -9,6 +9,8 @@ repos:
         args: [--allow-multiple-documents, --unsafe]
         exclude: (^tests/deploy/k8s-cloudbeat-tests/templates/.*)$|(^deploy/cloudformation/elastic-agent-ec2.yml)$
       - id: check-json
+      - id: check-executables-have-shebangs
+      - id: trailing-whitespace
```
amirbenun, line 13: nice

.pre-commit-config.yaml
orestisfl: cfn-python-lint will check the template semantically, yq will enforce a specific formatting style, like go-fmt. Ok with removing though.

Issue #697, opened 2023-02-02
[Discuss] Bundling vulnerability database into cloudbeat.tar.gz


amirbenun:
**Motivation**
For our vulnerability scanning flow, cloudbeat will use Trivy to scan the available AWS account volumes. In order to get the vulnerabilities it will download at the start of the first scan DB with all the latest vulnerabilities. We want to bundle our binary with the DB so if there's any issue downloading the latest one, we will have a fallback option.

**Definition of done**
- [ ] Propuse the solution
- [ ] ~~In Cloudbeat CI/CD pipeline, before we release a new version, we should download the database and package it to elastic-agent.tar.gz~~
- [ ] Cloudbeat should instruct trivy to read the vulnerabilities database from the downloaded file
- [ ] A failed database download should never cause an exit with an error but rather change cloudbeat to an unhealthy state and keep running

**Related tasks/epics**
- https://github.com/elastic/cloudbeat/issues/624
- https://github.com/elastic/cloudbeat/issues/696

amirbenun:
As for the second part of this task, packaging Trivy DB together with cloudbeat, it means enlarging the elastic-agent artifact with 400 Mb. Is that ok @eyalkraft @oren-zohar ?

oren-zohar:
This is pretty significant, and I think we should discuss it with the agent team before doing anything like that. I would like to hear your thoughts on this, @cmacknz @amitkanfer.

amitkanfer:
can you download on-demand? only if the integration is installed?

amirbenun:
Yes, this is what we do right now, and even if we package the vulnerabilities artifact we will still try to download it first.
This ticket is about having a workaround in case the vulnerabilities artifact won't be available for some reason.

amirbenun:
A continuation tasks for improving the vuln DB download process:
https://github.com/elastic/cloudbeat/issues/1048

PR #959, opened 2023-05-10
Vulnerabilities database cache management

amirbenun:
### Summary of your changes
This PR changes the way we handle trivy vuln db.
Before that PR, if the DB download failed, we remained without any alternative and just failed running.
With this PR, In case the download fails, we try to use the cache that exists from the previous run.

### Related Issues
- Resolves: https://github.com/elastic/cloudbeat/issues/697

### Checklist


vulnerability/db.go
```diff
@@ -0,0 +1,117 @@
...
+		err := f.useCache()
+		if err != nil {
+			return fmt.Errorf("VulnerabilityDBManager: could not use cache: %w", err)
+		}
+	} else {
+		err = f.renewCache()
+		if err != nil {
+			return fmt.Errorf("VulnerabilityDBManager: could not renew cache: %w", err)
+		}
+	}
```
oren-zohar, line 86: If I understand this flow correctly, this case is where there was no init error (the downloaded db is good) but it wasn't able to renew the cache with the new db, so we just exit and there will be no fallback here?

vulnerability/db.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (f VulnerabilityDBManager) useCache() error {
+	client := metadata.NewClient(cacheDir)
+	cache, err := client.Get()
+	if err != nil {
+		return fmt.Errorf("VulnerabilityDBManager: could not find cache: %w", err)
+	}
+
+	// TODO: Change cloudbeat status to unhealthy
```
oren-zohar, line 98: Please open a ticket with more information for this and link it here :pray:

vulnerability/db.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/aquasecurity/trivy-db/pkg/metadata"
+	"github.com/elastic/elastic-agent-libs/logp"
+
+	"github.com/aquasecurity/trivy/pkg/commands/artifact"
+	"github.com/aquasecurity/trivy/pkg/flag"
+	trivy_types "github.com/aquasecurity/trivy/pkg/types"
+)
+
+type VulnerabilityDBManager struct {
```
oren-zohar, line 34: we usually try to avoid starting the name with the name of the package

vulnerability/db.go
```diff
@@ -0,0 +1,117 @@
...
+			Debug:    true,
+			CacheDir: tmpCacheDir,
+		},
+		VulnerabilityOptions: flag.VulnerabilityOptions{
+			VulnType:      []string{trivy_types.VulnTypeOS, trivy_types.VulnTypeLibrary},
+			IgnoreUnfixed: false,
+		},
+		ScanOptions: flag.ScanOptions{
+			Scanners: []trivy_types.Scanner{trivy_types.VulnerabilityScanner},
+			RekorURL: "https://rekor.sigstore.dev",
```
oren-zohar, line 66: what is this? can you please add some documentation?

oren-zohar (changes requested):
Looks good! small comments

vulnerability/db.go
orestisfl: Also, do we want to export these structs if they are only used internally in the `vulnerability` package?

vulnerability/db.go
amirbenun: This also, part of the original [ticket](https://github.com/elastic/cloudbeat/issues/697). I will do that on a different PR.

vulnerability/db.go
amirbenun: This is their default value, since we are skipping their config initialization (where trivy assigns default values) we need to set it ourselves.

vulnerability/db.go
amirbenun: Yes, we will return the error and exit, however, if it happens it means something bad and unexpected happened on the environment and this is our way to raise that flag.

oren-zohar: We can still just log an error and continue to scan using the downloaded DB, no? Why would we want to stop scanning completely?

vulnerability/worker.go
```diff
@@ -41,8 +42,8 @@ type VulnerabilityWorker struct {
 	fetcher       VulnerabilityFetcher
 	replicator    VulnerabilityReplicator
 	verifier      VulnerabilityVerifier
-	evaluator     VulnerabilityScanner
-	runner        VulnerabilityRunner
+	scanner       VulnerabilityScanner
+	manager       VulnerabilityDBManager
```
orestisfl, line 46:
```suggestion
	dbManager       VulnerabilityDBManager
```

vulnerability/db.go
amirbenun: My reasons to stop the scan and not to support this flow.
1. Probability - This code will probably never kick in as these OS operations are pretty safe (deletion of a directory followed by a rename of another directory).
2. Maintenance - In all cases where scanning has started the flow was consistent (downloaded the DB and then replaced it with the old cache) and we don't have to maintain another flow which will ease debugging in the future.
3. Code clarity - Skipping this flow means that the cache directory is static, which keeps the functions signatures more simple.

Since I want to push this pr forward, I added the missing part where the `Download` function also returns what cache directory the `Scanner` should use. But if I convinced you I will be happy to revert this commit

oren-zohar (approved):

PR #961, opened 2023-05-11
CloudFormation: Add CSPM integration option

orestisfl:
### Summary of your changes

Add a CSPM integration option in the CloudFormation template, using a map to change the IAM role permissions depending on which of the two options is used

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/596

### Checklist


deploy/cloudformation/README.md
```diff
@@ -19,6 +19,7 @@ FLEET_URL="<Elastic Agent Fleet URL>"
 ENROLLMENT_TOKEN="<Elastic Agent Enrollment Token>"
 ELASTIC_ARTIFACT_SERVER="https://artifacts.elastic.co/downloads/beats/elastic-agent" # Replace artifact URL with a pre-release version (BC or snapshot)
 ELASTIC_AGENT_VERSION="<Elastic Agent Version>" # e.g: 8.8.0 | 8.8.0-SNAPSHOT
+INTEGRATION=CloudSecurityPostureManagement # Defaults to VulnerabilityManagement if not specified
```
oren-zohar, line 22: nit: Then let's put it here as the default value or comment this out, no?

oren-zohar (changes requested):
Maybe it's not part of this PR or I'm just unfamiliar with CloudFormation syntax, but I don't see where it's being defaulted to `VulnerabilityManagement` if `INTEGRATION` is not being provided

oren-zohar (approved):

PR #971, opened 2023-05-15
S3 provider handles missing public access block config better

ari-aviran:
### Summary of your changes
S3 provider now handles missing public access block configurations properly (logs a debug message instead of an error). Policy version has been updated to one that can handle null public access block configurations properly.

### Screenshot/Data

Screenshot 2023-05-15 at 14 14 05

### Related Issues
Related: https://github.com/elastic/csp-security-policies/pull/234
Related: https://github.com/elastic/csp-security-policies/issues/233

### Checklist

### DIFF
--- a/resources/providers/awslib/s3/provider.go
+++ b/resources/providers/awslib/s3/provider.go
@@ -36,9 +36,10 @@ import (
 )
 
 const (
-	EncryptionNotFoundCode = "ServerSideEncryptionConfigurationNotFoundError"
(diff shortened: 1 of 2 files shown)


oren-zohar (approved):
Approved. @ari-aviran please verify it works after BC4 is released.
@orestisfl I believe this is also relevant for your fix, so if possible can you also test that the issue you're investigating is solved in BC4

PR #957, opened 2023-05-10
Upgrade Go to 1.20

orouz:
### Summary of your changes

WIP

### Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary README/documentation (if appropriate)

PR #973, opened 2023-05-15
[Infra] Add health reporting mechanism

amirbenun:
### Summary of your changes
Launcher listens to health events and report back to elastic-agent

### Related Issues
- Resolves: https://github.com/elastic/cloudbeat/issues/697

### Checklist


health/health.go
```diff
@@ -0,0 +1,35 @@
+package health
+
+import (
+	"errors"
+)
+
+var Reporter = &reporter{
```
olegsu, line 7: Why have it on the global state? it might cause race calling it from multiple routines

amirbenun: I want the reporter to be a singleton.
On one side, every package can consume the reporter and report an health change.
On the other side, the launcher always listens to that reporter and updates the fleet accordingly.
I will also add it as a comment above that section.

.go-version
```diff
@@ -1 +1 @@
-1.19
+1.20
```
uri-weisman, line 1: Do we need to upgrade the hermit dependency as well?

health/health.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	r.ch <- errors.Join(list...)
+}
+
+func (r *reporter) Channel() <-chan error {
+	return r.ch
+}
+
+func (r *reporter) Close() {
```
uri-weisman, line 33: should we close the channel upon a graceful shutdown?

health/health.go
uri-weisman: I think it's correct to make the reporter a singleton, however still worth protecting the global var from "race conditions".

olegsu: Singleton does not mean it will be on the global state. 
Can we add a test that shows your suggestion @amirbenun that is also concurrency safe?

oren-zohar:
there's a separate go upgrade PR
- https://github.com/elastic/cloudbeat/pull/957

health/health.go
```diff
@@ -0,0 +1,40 @@
...
+	defer r.mut.Unlock()
+	r.errors[component] = err
+	list := make([]error, 0, len(r.errors))
+	for _, err := range r.errors {
+		if err != nil {
+			list = append(list, err)
+		}
+	}
+
+	r.ch <- errors.Join(list...)
```
drpaneas, line 31: JFYI this doesn't work with multiple-line errors, as each newline char will be interpreted as new entry. Here's a reproducer: Running the tests, I've noticed that one error message is separated from another one with the `\n` newline character. For example, the current output of tests is 4 errors:

```
component1 went wrong
component1 went wrong
component2 went wrong
component1 went wrong
```

but if I modify (not adding a new test, just modifying a already existing test-case) such as:

```diff
- err:       errors.New("component2 went wrong"),
+ err:       errors.New("component2 went wrong\ncomponent2 went wrong again"),
 ```

then running the tests it produces 5 messages:

```
component1 went wrong
component1 went wrong
component2 went wrong
component2 went wrong again
component1 went wrong
```

PR #981, opened 2023-05-17
k8s CIS 1.1.19: Scan pki/ folder recursively

orestisfl:
### Summary of your changes

This seems to be just a configuration error. Both the rego rules and the `Glob` go code seem to work fine.
I've updated all config files and added new functional tests.

### Screenshot/Data

[finding.json](https://github.com/elastic/cloudbeat/files/11498836/finding.json.txt)

Made by running `docker exec -it kind-multi-worker bash` and then `cd /etc/kubernetes/pki` and creating non-root-owned files under `somedir/` 

### Checklist

### DIFF
--- a/tests/commonlib/io_utils.py
+++ b/tests/commonlib/io_utils.py
@@ -102,7 +102,7 @@ class FsClient:
     """
 
     @staticmethod
-    def exec_command(
+    def exec_command(  # noqa: C901
         container_name: str,
         command: str,
         param_value: str,
@@ -122,6 +122,10 @@ class FsClient:
                 return
             with open(param_value, "a+", encoding="utf-8"):
                 pass
+            return
+
+        if command == "mkdir":
+            os.makedirs(param_value, exist_ok=True)
             return
 
         if command == "cat":
--- a/tests/product/tests/conftest.py
+++ b/tests/product/tests/conftest.py
@@ -2,6 +2,7 @@
 This module provides fixtures and configurations for
 product tests.
 """
+import os.path
 from pathlib import Path
 import time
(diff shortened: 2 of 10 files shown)


tests/commonlib/io_utils.py
```diff
@@ -102,7 +102,7 @@ class FsClient:
     """
 
     @staticmethod
-    def exec_command(
+    def exec_command(  # noqa: C901
```
gurevichdmitry, line 105: @orestisfl, I agree with flake8 that this function is too complex. Could you please create a task in the Automation infrastructure to refactor the function?

tests/product/tests/conftest.py
```diff
@@ -84,6 +86,12 @@ def config_node_pre_test(cloudbeat_start_stop):
         if node.metadata.name != cloudbeat_agent.node_name:
             continue
         for temp_file in temp_file_list:
+            api_client.exec_command(
+                container_name=node.metadata.name,
+                command="mkdir",
+                param_value=os.path.dirname(temp_file),
```
gurevichdmitry, line 92: nit: Since we are already working with `pathlib.Path`, we can utilize it to get the parent directory of a file instead of importing `os.path`. By using `str(Path(temp_file).parent)`, we can directly obtain the parent directory path as a string.

gurevichdmitry (approved):
@orestisfl, great work!

tests/commonlib/io_utils.py
orestisfl: https://github.com/elastic/cloudbeat/issues/1003

Issue #732, opened 2023-02-12
Handle logs from Trivy to be written to our logger instead of output


jeniawhite:

oren-zohar:
Verified:



![Image](https://github.com/elastic/cloudbeat/assets/85433724/01bcab45-67dc-4c2f-82f5-ffa293877f4d)

PR #1041, opened 2023-06-13
Override vulnerability management scanner logger

jeniawhite:
### Summary of your changes
This changes the behavior of the vulnerability scanner logger.
From now on it won't output to the console and the logs will be assigned to the cloudbeat logger.
Screen Shot 2023-06-19 at 13 05 39

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/732

### Checklist


flavors/vulnerability.go
```diff
@@ -18,6 +18,13 @@
 package flavors
 
 import (
+	xlog "github.com/masahiro331/go-xfs-filesystem/log"
+	"go.uber.org/zap"
+
```
orestisfl, line 23: nit (optional): remove whitespace and reorder dependencies. You can add a separate section for third party packages but they usually go after the standard imports.

orestisfl (approved):

PR #1056, opened 2023-06-20
Improve allure report summary

oren-zohar:
### Summary of your changes
Currently, we don't include the allure summary in the post-merge scenario, possibly because it doesn't seem logical to post it as a comment after merging. This pull request aims to address this by ensuring that the summary is at least posted in the workflow summary for easy visibility. I added handling for a case were no tests were run at all (current behavior is a green summary in this case, which should indicate an issue instead)

### Screenshot/Data
image

[Workflow run summary](https://github.com/oren-zohar/cloudbeat/actions/runs/5319663735)


gurevichdmitry (approved):
approved

PR #1058, opened 2023-06-20
add sanity checks action and update mask value

gurevichdmitry:
### Summary of your changes

This pull request resolves the masking issue of the `ec_api_key` and adds a new action in the Sanity Tests flow to perform sanity checks after the deployment.


.github/workflows/test-environment.yml
```diff
@@ -61,6 +59,12 @@ jobs:
         run: ./bin/hermit env -r >> $GITHUB_ENV
         working-directory: ./
 
+      - name: Set API Key
```
oren-zohar, line 62: nit:
```suggestion
      - name: Mask API Key
```

.github/workflows/test-environment.yml
```diff
@@ -142,6 +147,22 @@ jobs:
           aws s3 cp "${{ env.EC2_CSPM_KEY }}" "${{ env.S3_BUCKET }}/cspm.pem"
           aws s3 cp "${{ env.EC2_KSPM_KEY }}" "${{ env.S3_BUCKET }}/kspm.pem"
 
+      - name: Summary
+        if: always()
```
oren-zohar, line 151: if previous steps failed, will this step have any meaning?

dev-docs/sanity-workflow.md
```diff
@@ -75,4 +95,4 @@ The following environment variables are required:
 
 If you want to destroy the provisioned infrastructure, set the `cleanup-env` input to `true` when running the workflow. The cleanup step will be executed at the end.
 
-Note: Destroying the environment is irreversible, so use it with caution.
\ No newline at end of file
+Note: Destroying the environment is irreversible, so use it with caution.
```
oren-zohar, line 98: Let's remove this part, I don't want to discourage people from cleaning their cloud envs. They can always redeploy

tests/integration/tests/test_sanity_checks.py
```diff
@@ -9,7 +9,7 @@
 import pytest
 from commonlib.utils import get_findings
 
-CONFIG_TIMEOUT = 60
+CONFIG_TIMEOUT = 300
```
oren-zohar, line 12: do we still need to change the timeout? (as we excluded `file` type)

tests/integration/tests/test_sanity_checks.py
```diff
@@ -22,7 +22,7 @@
         "cloud-config",
     ],
     "cis_k8s": ["file", "process", "k8s_object"],
-    "cis_eks": ["file", "process", "k8s_object"],
+    "cis_eks": ["process", "k8s_object"],  # Optimize search findings by excluding 'file'.
```
oren-zohar, line 25: let's make sure we track this in the improvements issue

oren-zohar (changes requested):

.github/workflows/test-environment.yml
gurevichdmitry: I agree, it can be replaced by `success()`.

tests/integration/tests/test_sanity_checks.py
gurevichdmitry: In the latest run, all results were retrieved within 2 minutes. I will update the timeout to 120 seconds. However, it's advisable to keep some buffer time due to potential cloud factors that may affect the result. Nevertheless, if the result is retrieved before reaching the timeout, the test run will be considered finished and won't wait.

tests/integration/tests/test_sanity_checks.py
gurevichdmitry: New task is created
- https://github.com/elastic/security-team/issues/6909

oren-zohar (approved):

Issue #1065, opened 2023-06-21
CNVM agent upgrade path is broken from 8.8.0 to 8.8.1


nick-alayil:
**Describe the bug**
Upgrading the elastic agent from 8.8.0 to 8.8.1 is broken for CNVM integration.

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Spin-up a 8.8.0 ESS and install a CNVM integration using CloudFormation.
2. Update the ESS from 8.8.0 to 8.8.1
3. Upgrade the elastic agent used by CNVM integration from 8.8.0 to 8.8.1
4. Agent status will become unhealthy
5. On checking agent logs, it's clear agent upgrade failed due to lack of storage space. 

**Reproduction Video**
https://github.com/elastic/cloudbeat/assets/59900203/1afe1390-a0a4-4604-b51e-61ae827c2849

**Expected behavior**
Elastic agent should be successfully upgraded from 8.8.0 to 8.8.1

Thanks @jamesspi for letting me know about the issue 👍

orouz:
reproduced and fixed in https://github.com/elastic/cloudbeat/pull/1070 by increasing the volume size. 

worth noting, while reproducing, after initiating the upgrade flow the agent status changes were:
`healthy -> updating -> unhealthy -> healthy` - keeping version as 8.8.0 (same logs as in description)

after deploying the fix, the status changes were: 
`healthy -> updating -> unhealthy -> healthy` - version updated to 8.8.1

kfirpeled:
@orouz / @oren-zohar can you please describe how to verify this ticket in our first QA cycle?

mitodrummer:
QA looks good:


![Image](https://github.com/elastic/cloudbeat/assets/16198204/db60bc27-df41-48d2-a5e9-65bfc7efd95a)
![Image](https://github.com/elastic/cloudbeat/assets/16198204/1a9fb79b-25bf-4bbe-8ce6-490fd5802230)

PR #1070, opened 2023-06-22
[CNVM] Use amazon linux 2023 ami in CloudFormation  

orouz:
### Summary of your changes

this changes the AMI we use for defining the EC2 instance OS when deploying our cloud formation template. 

now using [amazon linux 2023](https://docs.aws.amazon.com/linux/al2023/ug/get-started.html)

- [x] [available in all regions](https://aws.amazon.com/about-aws/whats-new/2023/03/amazon-linux-2023/#:~:text=Amazon%20Linux%202023%20is%20generally%20available%20in%20all%20AWS%20Regions%2C)

### Related Issues

- https://github.com/elastic/security-team/issues/6894
- fixes https://github.com/elastic/cloudbeat/issues/1065 for newer versions


deploy/cloudformation/elastic-agent-ec2.yml
```diff
@@ -156,6 +156,10 @@ Resources:
           tar xzvf $ElasticAgentArtifact.tar.gz
           cd $ElasticAgentArtifact
           sudo ./elastic-agent install --non-interactive --url=${FleetUrl} --enrollment-token=${EnrollmentToken}
+      BlockDeviceMappings:
+        - DeviceName: /dev/xvda
+          Ebs:
+            VolumeSize: 10
```
orouz, line 162: got "not enough space on disk" when downloading elastic-agent without it. (default for this AMI looks like 2gb)

oren-zohar: could be related? https://github.com/elastic/cloudbeat/issues/1065

deploy/cloudformation/elastic-agent-ec2.yml
```diff
@@ -4,7 +4,7 @@ Description: Creates an EC2 instance with permissions to run elastic-agent
 Parameters:
   LatestAmiId:
     Type: 'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>'
-    Default: '/aws/service/canonical/ubuntu/server/jammy/stable/current/arm64/hvm/ebs-gp2/ami-id'
+    Default: '/aws/service/ami-amazon-linux-latest/amzn2-ami-minimal-hvm-arm64-ebs'
```
orouz, line 7: > As Amazon Linux AMIs are patched and new versions are released to the public, AWS updates the Parameter Store value with the latest ImageID value for each AMI namespace in all supported Regions.

[source](https://aws.amazon.com/blogs/compute/query-for-the-latest-amazon-linux-ami-ids-using-aws-systems-manager-parameter-store/#:~:text=As%20Amazon%20Linux%20AMIs%20are%20patched%20and%20new%20versions%20are%20released%20to%20the%20public%2C%20AWS%20updates%20the%20Parameter%20Store%20value%20with%20the%20latest%20ImageID%20value%20for%20each%20AMI%20namespace%20in%20all%20supported%20Regions.)

i think we could use a fixed version by date instead of always latest, but not sure we should as latest tag is probably the safest, which is the aim of this PR

deploy/cloudformation/elastic-agent-ec2.yml
orouz: yeah seems like it. i got that error after re-downloading the agent, which is about 2g when unpacked. i assume an upgrade flow does the same

deploy/cloudformation/elastic-agent-ec2.yml
```diff
@@ -1,10 +1,10 @@
 ---
-AWSTemplateFormatVersion: '2010-09-09'
+AWSTemplateFormatVersion: "2010-09-09"
 Description: Creates an EC2 instance with permissions to run elastic-agent
 Parameters:
   LatestAmiId:
     Type: 'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>'
-    Default: '/aws/service/canonical/ubuntu/server/jammy/stable/current/arm64/hvm/ebs-gp2/ami-id'
+    Default: '/aws/service/ami-amazon-linux-latest/al2023-ami-minimal-kernel-default-arm64'
```
orouz, line 7: > AWS provides Amazon Linux 2023 and Amazon Linux 2. If you are migrating from another Linux distribution to Amazon Linux, we recommend that you migrate to Amazon Linux 2023.

[source](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/amazon-linux-ami-basics.html#:~:text=AWS%20provides%20Amazon%20Linux%202023%20and%20Amazon%20Linux%202.%20If%20you%20are%20migrating%20from%20another%20Linux%20distribution%20to%20Amazon%20Linux%2C%20we%20recommend%20that%20you%20migrate%20to%20Amazon%20Linux%202023.)

seems like the recommended image. also both images - [amazon linux 2](https://aws.amazon.com/amazon-linux-2/?amazon-linux-whats-new.sort-by=item.additionalFields.postDateTime&amazon-linux-whats-new.sort-order=desc) and [amazon linux 2023](https://aws.amazon.com/linux/amazon-linux-2023/) have the same scan results

deploy/cloudformation/elastic-agent-ec2.yml
amirbenun: I would change it to 16 or even 32 just to be on the safe side.

amirbenun (approved):

amirbenun:
Don't forget to update the integration as well
https://github.com/elastic/integrations/blob/26a538f/packages/cloud_security_posture/manifest.yml
lines 115 and 177

Issue #1037, opened 2023-06-12
[AWS Orgs] Update cloudbeat.yml config to enable multi-account fetching


orestisfl:
**Motivation**
Add a configuration option in cloudbeat to enable multi-account-specific behavior. I am creating a separate ticket for to be done first so that incremental changes towards the AWS Orgs epic can only be conditionally enabled.

**Definition of done**
What needs to be completed at the end of this task
- [x] Cloudbeat defaults to single-account behavior
- [x] Cloudbeat parses new field correctly
- [x] Cloudbeat panics because it has nothing to do
- [x] Update https://github.com/elastic/integrations to add fields in hbs and manifest file

**Out of scope**
- Actually implement multi-account resource fetching

**Related tasks/epics**
- https://github.com/elastic/cloudbeat/issues/982

PR #1072, opened 2023-06-23
[AWS Orgs] Add config option

orestisfl:
### Summary of your changes

Adds the `aws.account_type` enum option which can take the `single_account` and `organization_account` values. `organization_account` is not implemented yet and leads to an error.

### Related Issues
- Related: https://github.com/elastic/integrations/pull/6682
- Fixes: https://github.com/elastic/cloudbeat/issues/1037

### Checklist


config/config.go
```diff
@@ -70,9 +75,22 @@ func New(cfg *config.C) (*Config, error) {
...
+	case "":
+	case SingleAccount:
+	case OrganizationAccount:
+		fallthrough // TODO: remove when implemented
+	default:
+		return nil, launcher.NewUnhealthyError(fmt.Sprintf(
+			"aws.account_type '%s' is not supported",
+			c.CloudConfig.AwsAccountType,
+		))
+	}
```
amirbenun, line 92: It feels to me that it should check that the value appear in a "supported values" array, same as `isSupportedBenchmark` does.
Also, this validation can already support the `OrganizationAccount` and just point it to the single account flow.

config/config_test.go
```diff
@@ -94,13 +99,13 @@ config:
 	for i, test := range tests {
 		s.Run(fmt.Sprint(i), func() {
 			cfg, err := config.NewConfigFrom(test.config)
-			s.NoError(err)
+			require.NoError(s.T(), err)
```
amirbenun, line 102: Cool! I wasn't familiar with that, I think we should do this change in many of our unit tests🥇

config/config.go
orestisfl: I've changed this to at least log an error, otherwise the configuration is completely ineffectual in the current code base.

config/config.go
```diff
@@ -56,8 +56,9 @@ type Config struct {
 }
 
 type CloudConfig struct {
-	AwsCred aws.ConfigAWS `config:"aws.credentials"`
-	GcpCfg  GcpConfig     `config:"gcp"`
+	AwsCred        aws.ConfigAWS `config:"aws.credentials"`
```
uri-weisman, line 59: Do you mind creating a new `AwsCfg` type to wrap all aws-related configurations?

uri-weisman (approved):

PR #1077, opened 2023-06-25
CloudFormation increase root volume size to 32 Gb

amirbenun:
### Summary of your changes
- increase root volume size to 32 Gb

### Checklist


oren-zohar:
@amirbenun please don't merge before verifying it fixes [this issue](https://github.com/elastic/cloudbeat/issues/1065). Also, why 32gb? will this have any effect on the cost?

amirbenun:
The root volume type is about to change as part of https://github.com/elastic/cloudbeat/pull/1070
We will open another pr if necessary.

Issue #769, opened 2023-02-22
Add a new benchmark type for CIS GCP


uri-weisman:
**Motivation**
Add a new benchmark type - `cis_gcp` to support the new integration.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Merge new input type to [agent spec](https://github.com/elastic/elastic-agent/blob/main/specs/cloudbeat.spec.yml).
- [ ] Cloudbeat supports the new input type
- [ ] Add input to the [manifest](https://github.com/elastic/integrations/blob/main/packages/cloud_security_posture/manifest.yml)

Issue #765, opened 2023-02-22
GCP Authentication util


uri-weisman:
**Motivation**
To ensure consistency across all GCP providers, a shared utility will be developed that utilizes the GCP SDK to authenticate the client using the [specified authentication methods](https://github.com/elastic/security-team/issues/6093). This utility will provide a standardized and efficient way for all providers to obtain the necessary credentials and access GCP resources to retrieve relevant findings.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Restructure cloudbeat config struct to support the agent's updated config.
- [ ] Common utility that supports the relevant GCP auth methods.

**Related tasks/epics**
- [filebeat implementation](https://github.com/elastic/beats/blob/5dd5e889c039f40a75b62374b529bf1c65b6a2f1/x-pack/filebeat/input/httpjson/config_auth.go)

orouz:
seems like we need to wrap the GCP SDK in order to authenticate because our integration sends credentials values as optional (this is how it's done in AWS) so we need to figure out which one to use. 

an alternative approach would be the prevent the user from installing the integration if credentials are required but weren't provided. we already prevent the installation on a different [case](https://github.com/elastic/kibana/blob/0152724d47c9dde677e73b25b579c42f01097d55/x-pack/plugins/cloud_security_posture/server/plugin.ts#L118-L120), so we could do the same here. if we do, cloudbeat could pass the credentials values directly to the GCP SDK with very little need for wrapping. 

when a user will try to install an integration without providing credentials (when they are required), they will get an error notification when trying to install the integration. (like the one they get for "package policy already exists by that name")

@tinnytintin10 thoughts on this?

PR #1078, opened 2023-06-25
[CIS GCP] Add auth methods and new input type

uri-weisman:
### Summary of your changes
1. cloudbeat supports the new input type `cis gcp`.
2. Add a new GCP client auth method to support client configuration coming from fleet cfg.
It's planned that the user will have two options, providing a path to the SA file or passing the JSON content.

### Related Issues
- https://github.com/elastic/cloudbeat/issues/765
- https://github.com/elastic/cloudbeat/issues/769

### Checklist


resources/providers/gcplib/auth/credentials.go
```diff
@@ -0,0 +1,41 @@
...
+	"google.golang.org/api/option"
+)
+
+func GetGcpClientConfig(cfg *config.Config) ([]option.ClientOption, error) {
+	// Create credentials options
+	gcpCred := cfg.CloudConfig.GcpCfg
+	var opt []option.ClientOption
+	if gcpCred.CredentialsFilePath != "" && gcpCred.CredentialsJSON != "" {
+		return nil, errors.New("both credentials_file_path and credentials_json specified, you must use only one of them")
+	} else if gcpCred.CredentialsFilePath != "" {
```
orouz, line 32: could we favor one instead of returning an error?

this would play nicer with the way our integration installation/update works, as the form will send all fields filled by the user, so in the case they do fill both fields, they will end up learning it didn't work from the logs. 

maybe even even try both if one of them happens to not work. 

wdyt?

uri-weisman: @orouz - I was wondering about this.
Do you think we should add a verification to the user input? 
For example, in case the SA file path does not exist, fallback to the other option? or leave this responsibility to the user.
Same as for the JSON content.

orouz (approved):

resources/providers/gcplib/auth/credentials.go
orouz: thinking on this again, maybe we should just favor the JSON? seems like the file path would be used for the actual prod env and the json would be for development, so preferring it would allow a better DX. 

i'm good either way, just figured since we do have values, best to use them

orouz (approved):

resources/providers/gcplib/auth/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+			opts = append(opts, option.WithCredentialsJSON([]byte(gcpCred.CredentialsJSON)))
+		} else {
+			return nil, errors.New("invalid credentials JSON")
+		}
+	}
+
+	return opts, nil
+}
+
+func validateJSONFromFile(log *logp.Logger, filePath string) error {
```
oren-zohar, line 62: the logger is unused

config/config.go
```diff
@@ -57,6 +57,17 @@ type Config struct {
 
 type CloudConfig struct {
 	AwsCred aws.ConfigAWS `config:"aws.credentials"`
+	GcpCfg  GcpClientOpt  `config:"gcp"`
```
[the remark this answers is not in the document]
uri-weisman: Good catch!

config/config.go
```diff
@@ -57,6 +57,17 @@ type Config struct {
...
+}
+
+type GcpConfig struct {
+	ProjectId    string `config:"project_id"`
+	GcpClientOpt `config:"credentials"`
+}
+
+type GcpClientOpt struct {
+	CredentialsJSON     string `config:"credentials_json"`
```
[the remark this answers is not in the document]
uri-weisman: unfortunately no, the expected type is string.

resources/providers/gcplib/auth/credentials.go
uri-weisman: removed

oren-zohar (approved):
Approved. Small comments

resources/providers/gcplib/auth/credentials.go
```diff
@@ -0,0 +1,77 @@
...
+			return nil, errors.New("invalid credentials JSON")
+		}
+	}
+
+	return opts, nil
+}
+
+func validateJSONFromFile(filePath string) error {
+	if _, err := os.Stat(filePath); errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("the file %q cannot be found", filePath)
```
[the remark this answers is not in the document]
orestisfl: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context.

PR #1082, opened 2023-06-26
[CI] Update golangci linter step to avoid timeout

uri-weisman:
### Summary of your changes
Multiple builds are failing due to golangci linter timeout, instead of increasing the timeout let's try and add cache to the linter step, it might reduce the execution time.

- [failed build example](https://github.com/elastic/cloudbeat/actions/runs/5372870113/jobs/9746682963)

### Checklist


.github/workflows/unit-test.yml
```diff
@@ -122,7 +122,9 @@ jobs:
         uses: golangci/golangci-lint-action@v3.6.0
         with:
           version: latest
-          args: --timeout=10m --whole-files
+          args: --timeout=10m --whole-files -v
+          skip-pkg-cache: true
```
orestisfl, line 126: Doesn't this disable caching?

We might want to instead try running this job with the official instructions which claim that caching is enabled by default:

```
jobs:
  golangci:
    name: lint
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: actions/setup-go@v4
        with:
          go-version: '1.20'
          cache: false
      - name: golangci-lint
        uses: golangci/golangci-lint-action@v3
...
```

uri-weisman: I've added `skip-pkg-cache` to address a similar issue to this [one](https://github.com/golangci/golangci-lint-action/issues/23).
I'll try your suggested setup as well, however, I'm still experimenting so I'll make it a draft pr.

orestisfl (approved):

.github/workflows/unit-test.yml
```diff
@@ -103,26 +103,25 @@ jobs:
...
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           repository: ${{ (github.event.pull_request_target || github.event.pull_request).head.repo.full_name }}
           ref: ${{ (github.event.pull_request_target || github.event.pull_request).head.sha }}
 
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: .go-version
+          cache: false
```
oren-zohar, line 117: Seems like this would remove the caching, no?

oren-zohar (changes requested):

.github/workflows/unit-test.yml
orestisfl: Caching is supposed to be handled in https://github.com/golangci/golangci-lint-action

oren-zohar (approved):

PR #1087, opened 2023-06-27
Use beat.Beater interface to avoid warning

orestisfl:
Otherwise, we get the "Exported function with the unexported return type" warning.


oren-zohar (approved):
@orestisfl where did you see the warnings?

orestisfl:
> @orestisfl where did you see the warnings?

GoLand

Issue #1097, opened 2023-06-29
[CNVM] Vulnerability flavor context cancelation causes deadlock


amirbenun:
### Describe the bug
When cloudbeat get a configuration update, the launcher has to kill the current flavor and recreate one instead.
In practice, the vulnerability flavor enters a deadlock and you will never get vulnerabilities.

### Technical explanation
The vulnerability worker is in charge on the asynchronous pipeline of:
Listing EC2s -> Creating Snapshots -> Scan -> Send vulnerabilities.
When the context is canceled, trivy returns an error and the scanner stops listening to the incoming channel.
All other providers are waiting to send their data to the next provider in the pipeline, but since no one reads it causes a deadlock.

### Preconditions
Any version > 8.9.0 of ESS

### To Reproduce
Run Cloudbeat with the CNVM integration installed
Change the integration name to trigger a configuration update

### Expected behavior
To get vulnerabilities

oren-zohar:
@amirbenun did it also happen on 8.8?

amirbenun:
Yes, we should discuss if we want to backport it to 8.8.
@tehilashn what do you think?

tehilashn:
I purpose we'll backport this to 8.8 anyway. If there will be another 8.8 release then the fix will be there.

amirbenun:
Sounds good @tehilashn.
I opened a PR with the backport to 8.8
https://github.com/elastic/cloudbeat/pull/1116

kfirpeled:
Verify by looking for errors in cloudbeat logs after a fresh deployment (0 errors)

Omolola-Akinleye:
I see one error @oren-zohar  @amirbenun 
```
20:08:24.595
elastic_agent.cloudbeat
[elastic_agent.cloudbeat][error] VulnerabilityScanner.scan.ScanVM error: scan error: scan failed: failed analysis: inspection error: post analysis error: post analysis error: walk dir error: walk error: context canceled
```

PR #1099, opened 2023-06-29
Vuln mgmt async providers ctx fix

amirbenun:
### Summary of your changes
The vulnerability management async providers didn't use the `context.Done` channel as they should which caused some bugs.

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/1097

### Checklist


jeniawhite:
Please also handle this case:
```golang
			e.ch <- e.generateEvent(data.reportResult, data.vulnerability, data.snapshot, data.seq)
```

jeniawhite:
This also a case that should be handled:
```golang
	for _, result := range unmarshalledReport.Results {
		for _, vul := range result.Vulnerabilities {
			// TODO: Replace sequence with more generic approach
			f.ch <- Result{
				reportResult:  result,
				vulnerability: vul,
				snapshot:      snap,
				seq:           f.seq,
			}
		}
	}
```

vulnerability/worker.go
```diff
@@ -95,56 +95,55 @@ func NewVulnerabilityWorker(log *logp.Logger, c *config.Config, cdp dataprovider
...
+			if err != nil {
+				f.log.Error("VulnerabilityWorker.work FetchInstances failed: %v", err)
+				return
+			}
+			f.log.Info("VulnerabilityWorker.work FetchInstances finished")
+		}()
+		f.wg.Add(1)
+		go func() {
+			defer f.wg.Done()
+			f.cleaner.GetChan() <- f.replicator.SnapshotInstance(ctx, f.fetcher.GetChan())
```
jeniawhite, line 117: This case should be handled as well

jeniawhite:
I think that this problem exists in flows that are outside Vulnrability as well and we need to make sure that we handle all of our channels in the codebase using either a library or refactoring them to avoid deadlocks.
For example:
```golang
func (f *S3Fetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
	f.log.Info("Starting S3Fetcher.Fetch")
	buckets, err := f.s3.DescribeBuckets(ctx)
	if err != nil {
		f.log.Errorf("failed to load buckets from S3: %v", err)
		return nil
	}

	for _, bucket := range buckets {
		resource := S3Resource{bucket}
		f.log.Debugf("Fetched bucket: %s", bucket.GetResourceName())
		f.resourceCh <- fetching.ResourceInfo{
			Resource:      resource,
			CycleMetadata: cMetadata,
		}
	}

	return nil
}
```

oren-zohar:
@amirbenun don't forget to backport

jeniawhite (approved):

vulnerability/fetcher_test.go
```diff
@@ -64,10 +67,19 @@ func TestVulnerabilityFetcher_FetchInstances(t *testing.T) {
 
 	assert.ElementsMatch(t, expectedInstances, receivedInstances)
 	wg.Wait()
+
+	// Verify no goroutines are leaking. Safest to keep this on top of the function.
+	// Go defers are implemented as a LIFO stack. This should be the last one to run.
+	goleak.VerifyNone(t, goleak.IgnoreCurrent())
```
orestisfl, line 73: I don't think this invocation of goleak is correct here, with this syntax you are ignoring all goroutines. `VerifyNone` is supposed to be called with `defer` before you run any of the code you are testing as `IgnoreCurrent` will be initialized only with the goroutines that existed since before

amirbenun: Yep, I think you are right, I will fix it

PR #1106, opened 2023-07-02
add AWS tags for CNVM deployment

gurevichdmitry:
### Summary of your changes

This pull request adds a script that applies tags to CNVM resources. The script, written in Bash, takes three input parameters: AWS region, stack name, and tags. It uses the AWS CLI to retrieve the instance ID associated with the specified CloudFormation stack. The script waits for the instance ID to be available and then applies the provided tags to the corresponding EC2 instance. To prevent indefinite waiting, a maximum timeout is enforced.


.github/workflows/test-environment.yml
```diff
@@ -229,6 +230,11 @@ jobs:
           aws s3 cp "${{ env.FLEET_API_DIR}}/kspm_eks.yaml" "${{ env.S3_BUCKET }}/kspm_eks.yaml"
           aws s3 cp "${{ env.FLEET_API_DIR}}/cspm-linux.sh" "${{ env.S3_BUCKET }}/cspm-linux.sh"
 
+      - name: Tag CNVM Instance
```
orestisfl, line 233: I am not sure why this needs to be a separate step, can't the tags be added via terraform?

gurevichdmitry: No, CNVM is not deployed through Terraform. After provisioning the infrastructure using Terraform, the next steps involve installing CNVM integration and deploying an agent. However, it's important to note that the agent deployment is an asynchronous operation, and we don't wait until the stack is fully created. To save time and optimize the workflow, we proceed to the next actions and execute the CNVM tagging action separately.

orestisfl: Understood

Then, perhaps we can add it to the script we have to deploy CloudFormation? This will also remove the requirement to add a step that waits for initialization.

justfile
```diff
@@ -122,6 +122,9 @@ logs-cloudbeat:
 deploy-cloudformation:
   cd deploy/cloudformation && go run .
 
+create-cnvm-stack-tags region stack_name tags='Key=division,Value=engineering Key=org,Value=security Key=team,Value=cloud-security-posture':
```
oren-zohar, line 125: Can you put the default common tags into a const instead?

.github/workflows/test-environment.yml
gurevichdmitry: This is exactly the point: the `just deploy-cloudformation` command initiates the deployment of CloudFormation, but it doesn't wait for the stack creation to be completed. At this point in time, the `instance_id `doesn't exist yet because the EC2 instance associated with the agent is still being provisioned. In order to tag the EC2 instance associated with the agent, we need to obtain the instance_id. However, retrieving the `instance_id` is an asynchronous process, and we don't have visibility into when it will be available.

If we were to include the tagging command within the CloudFormation creation step, it would extend the overall workflow duration as we would have to wait for the `instance_id` before proceeding. This delay could potentially add several minutes to the overall runtime.

By providing a separate step for tagging, we can reduce the waiting time. This separation allows the tagging action to be performed "in parallel" (although it's not truly parallel) with other actions, thereby optimizing the overall workflow.

Regarding your mention of a step that waits for initialization, I'm not sure about its context. If you can provide more details, I can provide a more specific response

justfile
oren-zohar: I think it's missing the `project` tag. based on [this](https://docs.google.com/document/d/1FdkJ5EM1tSAqVRvZvmtDSXbBrzS8f7Bw2vENTFes1ek/edit#heading=h.36dv1gu8gyr7)

oren-zohar (changes requested):

oren-zohar:
@gurevichdmitry do you have an example run?

.github/workflows/test-environment.yml
oren-zohar: I would even take it a step further and put the tagging step last. Only if the `cleanup-env` is set to false, as there is no need to tag an env that is supposed to be deleted, and if the delete step failed for some reason, the resources will be auto-deleted since they fail the tagging policy, wdyt?

justfile
gurevichdmitry: justfile is updated

.github/workflows/test-environment.yml
```diff
@@ -36,6 +36,7 @@ env:
   AWS_REGION: "eu-west-1"
   WORKING_DIR: deploy/test-environments
   FLEET_API_DIR: fleet_api/src
+  AWS_DEFAULT_TAGS: "Key=division,Value=engineering Key=org,Value=security Key=team,Value=cloud-security-posture Key=project,Value=test-environments"
```
oren-zohar, line 39: Is there a point in keeping this here if it's the same as the default justfile command value? I think it's a good practice to have our CI running with the default justfile values, so if someone breaks some value there it will fail the CI (obviously not in this case, since it will only affect the tags)

oren-zohar (changes requested):

.github/workflows/test-environment.yml
gurevichdmitry: In CI, we don't use the default value for the tags in `just create-cnvm-stack-tags` because we include the dynamically retrieved `owner` tag `github.actor`.

orestisfl (approved):
As discussed, it is possible to do this with the CloudFormation dev script, without the need to add additional overhead to the CI. However, we can merge this as it is and follow up with a new ticket to address the improvement later.

.github/workflows/test-environment.yml
gurevichdmitry: workflow is updated

gurevichdmitry:
> As discussed, it is possible to do this with the CloudFormation dev script, without the need to add additional overhead to the CI. However, we can merge this as it is and follow up with a new ticket to address the improvement later.

[New task](https://github.com/elastic/security-team/issues/7026) is opened

.github/workflows/test-environment.yml
gurevichdmitry: test flow is updated

PR #1112, opened 2023-07-05
Move fetchers to folders

orouz:
We're going to introduce GCP fetchers, which may overlap with AWS fetchers, so it'd be better if we grouped fetchers by folders.


uri-weisman (approved):

orestisfl (changes requested):
You'll need to rename the package each one is in, which will also probably cause a lot of naming conflicts in other files

transformer/events_creator_test.go
```diff
@@ -34,7 +34,7 @@ import (
 	"github.com/elastic/cloudbeat/dataprovider/types"
 	"github.com/elastic/cloudbeat/evaluator"
 	"github.com/elastic/cloudbeat/resources/fetching"
-	"github.com/elastic/cloudbeat/resources/fetching/fetchers"
+	fetchers "github.com/elastic/cloudbeat/resources/fetching/fetchers/k8s"
```
orestisfl, line 37: The failure with golangci-lint happens because you need to update the exclude rules in https://github.com/elastic/cloudbeat/blob/a89b3eef8aab022324a0b7ce67971338c1aedefc/.golangci.yaml#L40-L41

resources/fetching/factory/eks_factory.go
```diff
@@ -27,7 +27,8 @@ import (
 
 	"github.com/elastic/cloudbeat/resources/fetching"
 	"github.com/elastic/cloudbeat/resources/fetching/condition"
-	"github.com/elastic/cloudbeat/resources/fetching/fetchers"
+	awsFetchers "github.com/elastic/cloudbeat/resources/fetching/fetchers/aws"
+	k8sFetcher "github.com/elastic/cloudbeat/resources/fetching/fetchers/k8s"
```
oren-zohar, line 31: nit: is it fetcher or fetchers? nit2: in any case I think the convention is to avoid capitalization altogether so it should be something like `k8sfetcher` [see here](https://go.dev/blog/package-names)

oren-zohar (changes requested):

resources/fetching/factory/eks_factory.go
```diff
@@ -27,7 +27,8 @@ import (
 
 	"github.com/elastic/cloudbeat/resources/fetching"
 	"github.com/elastic/cloudbeat/resources/fetching/condition"
-	"github.com/elastic/cloudbeat/resources/fetching/fetchers"
+	awsf "github.com/elastic/cloudbeat/resources/fetching/fetchers/aws"
+	k8sf "github.com/elastic/cloudbeat/resources/fetching/fetchers/k8s"
```
oren-zohar, line 31: abbreviations are usually something we avoid, but if you think it's clear I'm fine with it

oren-zohar (approved):

orestisfl (approved):

PR #1113, opened 2023-07-05
Remove unused folder

orestisfl:


oren-zohar:
@orestisfl can you also remove `skaffold.yml` from the root folder while you're at it?

oren-zohar (approved):

PR #1114, opened 2023-07-05
Add es ssl cert auth option to deployment files

orouz:
### Summary of your changes

`cloudbeat-aws.yml` is used by developers to run cloudbeat standalone. we already use some env vars to specify ES configuration, this PR adds an `ES_CERT` env var to be used as a certificate authority by ES, which enables cloudbeat to send data properly to an elastic-package https environment. 

### Checklist


deploy/aws/cloudbeat-aws.yml
```diff
@@ -43,6 +43,9 @@ output.elasticsearch:
 
   # Enable to allow sending output to older ES versions
   allow_older_versions: true
+  ssl:
+    certificate_authorities:
+      - ${ES_CERT}
```
orouz, line 48: when using with elastic-package, the value will be at:
`~/.elastic-package/profiles/default/certs/ca-cert.pem`

deploy/aws/cloudbeat-aws.yml
orouz, line 1: when i want the logs to be easier to follow, i usually:

1. comment out `processors`
2. specify `posture`, `opa`, `launcher` as the log selectors


maybe worth adding as well

deploy/aws/cloudbeat-aws.yml
olegsu: Have you also checked the cloud deployment? any adjuments required?

orouz: when running cloudbeat locally with this and setting `ES_HOST` to something like `https://test-certs.kb.us-west2.gcp.elastic-cloud.com:443`, does it work for you without defining anything related to `ssl` ? asking because when i run it locally against elastic package i need to set `ssl.certificate_authorties` and it doesn't work without it, which turns out is the same (for me) when running against a remote `es`

orouz:
using elastic-package cert with a remote ES deployment doesn't work (unknown authritoy), closing this and will try to solve it on the IDE level. (opened because i keep ignoring this modified file)

PR #1127, opened 2023-07-11
Convert to test suites

amirbenun:
### Summary of your changes
Convert tests into test suites to fix a bug where the `goleak.VerifyNone()` doesn't work.

### Checklist

### DIFF
--- a/vulnerability/fetcher_test.go
+++ b/vulnerability/fetcher_test.go
@@ -47,8 +47,6 @@ func TestFetcherTestSuite(t *testing.T) {
 }
 
 func (s *FetcherTestSuite) TearDownTest() {
-	// Verify no goroutines are leaking. Safest to keep this on top of the function.
-	// Go defers are implemented as a LIFO stack. This should be the last one to run.
 	goleak.VerifyNone(s.T(), s.opts)
 }
 
--- a/vulnerability/fetcher_test.go
+++ b/vulnerability/fetcher_test.go
@@ -37,15 +37,11 @@ import (
 
 type FetcherTestSuite struct {
 	suite.Suite
-
-	log  *logp.Logger
 	opts goleak.Option
 }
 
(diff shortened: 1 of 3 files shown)


vulnerability/fetcher_test.go
```diff
@@ -35,14 +35,31 @@ import (
...
+func TestFetcherTestSuite(t *testing.T) {
+	s := new(FetcherTestSuite)
+	s.opts = goleak.IgnoreCurrent()
+	suite.Run(t, s)
+}
+
+func (s *FetcherTestSuite) TearDownTest() {
+	// Verify no goroutines are leaking. Safest to keep this on top of the function.
+	// Go defers are implemented as a LIFO stack. This should be the last one to run.
+	goleak.VerifyNone(s.T(), s.opts)
```
orestisfl, line 52: I don't mind the approach but FYI you don't need a full test suite, you just need to `defer` the call from the test function.

vulnerability/fetcher_test.go
```diff
@@ -35,14 +35,31 @@ import (
...
+
+func TestFetcherTestSuite(t *testing.T) {
+	s := new(FetcherTestSuite)
+	s.opts = goleak.IgnoreCurrent()
+	suite.Run(t, s)
+}
+
+func (s *FetcherTestSuite) TearDownTest() {
+	// Verify no goroutines are leaking. Safest to keep this on top of the function.
+	// Go defers are implemented as a LIFO stack. This should be the last one to run.
```
orestisfl, line 51:
```suggestion
```

comment doesn't seem relevant

orestisfl (approved):

vulnerability/fetcher_test.go
amirbenun: I know, I just aligned with other tests we have already that call `VerifyNone`

Issue #1118, opened 2023-07-06
[CNVM] Sort EC2 instances in ascending order of disk size


amirbenun:
### Motivation
Scanning EC2 instances can take a lot of time, especially if the first instance being scanned has a large disk size. In order to reduce the time it takes to get the first vulnerability we can sort the EC2 instances that we scan based on their disk size.
Both for tests and real users it can help to get quicker feedback of the environment.

### Definition of done
- [x] EC2 fetcher should sort instances in ascending order
- [ ] Update CF

**Related tasks/epics**
Reference related issues and epics

amirbenun:
With a short research I found that AWS go client function [ec2.DescribeVolumes](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeVolumes) returns a list of volumes with a single request. A request can also filter for volumes attached to specific instances.

In our case, after we get all the instances, with an extra single call (or paginated) we will have all the volumes attached to the instances we want to scan. then, we can compare the instance `RootDeviceName` field with the `InstanceId` and `Device` fields on the volume attachments object.

Another issue that comes to my mind is that we create snapshots for all volumes attached to an instance even though Trivy will only scan the root volume.
After implementing this feature, we will be able to create snapshots only for root volumes and **save even more time and money**.

@nick-alayil @oren-zohar I think it worths prioritizing this task for 8.10.

oren-zohar:
We suspect (@jeniawhite, @amirbenun, @nick-alayil) that not only the size of the disk determines the scan's duration, but also the percentage of it that is used.

amirbenun:
Reopen to track the integration changes

kfirpeled:
@amirbenun/@oren-zohar can you please describe how to verify this ticket in our first QA cycle?

amirbenun:
Create several instances with different disk sizes on the same region (empty region with no instances).
Run CNVM on that region, make sure that the first instance that was scanned is the one with the smallest size.

**Note:** cloudbeat compares the size of the root volume only

mitodrummer:
QA Evidence:
EC2 volumes list sorted size ASC: (highlighting maxcold-8-10-v2-qa-1) size: 20gb
![Image](https://github.com/elastic/cloudbeat/assets/16198204/2b872e1f-4cce-4dd3-b2c8-064f75504136)

Discover sorted timestamp DESC
[...]

PR #1133, opened 2023-07-13
[CNVM] Sort EC2 instances by volume size

amirbenun:
### Summary of your changes
Scan instances based on their disk size to get vulnerabilities quickly

### Related Issues
- Resolves: https://github.com/elastic/cloudbeat/issues/1118

### Checklist


amirbenun:
`TODO:` Need to add `ec2:DescribeVolumes` permission after CloudFormation separation PR is merged

resources/providers/awslib/ec2/volume.go
```diff
@@ -0,0 +1,28 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package ec2
+
+type EC2Volume struct {
+	VolumeId   string
+	InstanceId string
+	Region     string
+	awsAccount string
```
uri-weisman, line 24: Are those fields needed?
can't we just use the native AWS volume type instead of creating a new type? I understand we'll have fields we don't necessarily need.

oren-zohar:
@amirbenun did it improve the runtime?

resources/providers/awslib/ec2/volume.go
amirbenun: I also wondered about that.
The reason I added the struct is to avoid of using AWS SDK structs all over our source code. In my opinion, the AWS providers should encapsulate the SDK calls and return as well a struct defined by our team.
However, I am open to hear opinions and learn the best practices regarding that extra layer and if it brings more benefits than overhead.

uri-weisman: IMO, if customized fields are unnecessary and our focus is on refining the provided structure, I would stick with simplicity and use the AWS struct, though it may not be crucial.
Regarding the `Region` and `awsAccount` fields, are they being used?

amirbenun:
@oren-zohar 
> did it improve the runtime?

On my environment, the first instance cloudbeat scans is 8 Gb and it generates vulnerabilities after 9 minutes. 
In most sanity tests we had cloudbeat took longer than that to generate vulnerabilities.

resources/providers/awslib/ec2/provider.go
```diff
@@ -294,3 +295,49 @@ func (p *Provider) GetRouteTableForSubnet(ctx context.Context, region string, su
...
+
+			result = append(result, &Volume{
+				VolumeId:   *vol.VolumeId,
+				Size:       int(*vol.Size),
+				awsAccount: p.awsAccountID,
+				Region:     region,
+				Encrypted:  *vol.Encrypted,
+				InstanceId: *vol.Attachments[0].InstanceId,
+				Device:     *vol.Attachments[0].Device,
+			})
```
amirbenun, line 338: @uri-weisman 
The `Volume` I defined is different than the one defined by the SDK.
AWS volume is not necessarily attached to an instance, the `Volume` here has the `InstanceId` field that makes it easier to map the relations between instances and volumes.

resources/providers/awslib/ec2/provider.go
```diff
@@ -294,3 +295,49 @@ func (p *Provider) GetRouteTableForSubnet(ctx context.Context, region string, su
 
 	return routeTables.RouteTables[0], nil
 }
+
+func (p *Provider) DescribeVolumes(ctx context.Context, instances []*Ec2Instance) ([]*Volume, error) {
+	instanceFilter := lo.Map(instances, func(ins *Ec2Instance, _ int) string { return *ins.InstanceId })
+	insances, err := awslib.MultiRegionFetch(ctx, p.clients, func(ctx context.Context, region string, c Client) ([]*Volume, error) {
```
uri-weisman, line 301:
```suggestion
	volumes, err := awslib.MultiRegionFetch(ctx, p.clients, func(ctx context.Context, region string, c Client) ([]*Volume, error) {
```

uri-weisman (approved):
Well done, Approved
Just notice the typo on https://github.com/elastic/cloudbeat/pull/1133/files/e2c95136c3295dad477d93e30cfd968a51555323#r1269405227

Issue #1088, opened 2023-06-27
[CNVM] Integration test (cloudbeat binary + AWS)


amirbenun:
### Motivation
As a **premerge** test, an integration test of cloudbeat binary with a test AWS account can get us a great coverage of the cloudbeat functionality.

### Components
- Standalone Cloudbeat binary
- AWS test account
- Elasticsearch

### Test flow
- Cloudbeat runs vulnerability management flow
- Scan a specific region in our AWS account with no instances (to reduce costs)
- Assert for vulnerabilities in ES

### Related tasks/epics
- https://github.com/elastic/cloudbeat/issues/700
- https://github.com/elastic/cloudbeat/issues/1089

amirbenun:
When running the Cloudbeat CNVM on a Github action instance we get the following error:
```
Exiting: beater returned an error: VulnerabilityWorker: could not get init NewVulnerabilityRunner: DB error: failed to download vulnerability DB: database download error: oci download error: download error: failed to download: chtimes db/trivy.db: operation not permitted
```

Which must be related to https://github.com/elastic/cloudbeat/issues/731 and the fact that we have a limitation today running CNVM on AMD processors and it only works on ARM.

oren-zohar:
@amirbenun we can change the runner image, no?

amirbenun:
I will try using `macos-12-xl` but it's also amd based.
![image](https://github.com/elastic/cloudbeat/assets/34831306/de09b059-4687-4a31-8c0c-04972e682dae)

oren-zohar:
are those running on ARM? [maybe they still don't have Apple silicon Github runners](https://github.com/github/roadmap/issues/528)

amirbenun:
Moving the task to block and trying to resolve https://github.com/elastic/cloudbeat/issues/731

PR #1147, opened 2023-07-19
Add CNVM premerge test

amirbenun:
### Summary of your changes
Add a premerge test that:
- Install elasticsearch
- Run cloudbeat CNVM
- Wait for vulnerabilities

### Related Issues
- Resolves: https://github.com/elastic/cloudbeat/issues/1088

### Checklist


amirbenun:
An example of a success run:
https://github.com/amirbenun/cloudbeat/actions/runs/5728082166/job/15521853869?pr=3

deploy/vulnerability/cloudbeat-vuln-mgmt.yml
```diff
@@ -38,9 +38,12 @@ output.elasticsearch:
...
 processors:
-  - add_host_metadata: ~
-  - add_cloud_metadata: ~
-  - add_docker_metadata: ~
+
+# Allow executing system call being used by trivy when os.Chtimes
+seccomp.syscalls:
+  - action: allow
+    names:
+    - utimensat
```
oren-zohar, line 46: :100:

oren-zohar (changes requested):
I'm afraid 22 minutes for apre-merge test is too long :(. Let's see how can we improve the runtime, or set it to a post-merge until then

.github/workflows/cnvm-ci.yml
```diff
@@ -0,0 +1,82 @@
...
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - name: Init Hermit
+        run: ./bin/hermit env -r >> $GITHUB_ENV
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
```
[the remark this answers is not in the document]
amirbenun: Poetry cannot be installed without it, not sure why.
Hermit points to python3.10, when I tried changing it to 3.9 it also failed.

deploy/vulnerability/cloudbeat-vuln-mgmt.yml
```diff
@@ -38,9 +38,12 @@ output.elasticsearch:
 
 # ================================= Processors =================================
 processors:
-  - add_host_metadata: ~
-  - add_cloud_metadata: ~
-  - add_docker_metadata: ~
```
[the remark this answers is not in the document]
amirbenun: Basically the fields added by these processors are irrelevant for CNVM.
Cloudbeat populates the same fields with values of the scanned object rather than the host (the scanner instance).
It also doesn't run in a docker.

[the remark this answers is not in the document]
oren-zohar: will this affect the users? or does this is only for dev purposes and what really matter is the one in /integrations

.github/workflows/cnvm-ci.yml
```diff
@@ -0,0 +1,75 @@
...
+jobs:
+  Run-CNVM-Tests:
+    name: CNVM integration test
+    runs-on: ubuntu-22.04
+    timeout-minutes: 40
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - name: Init Hermit
```
orestisfl, line 23: Do we actually need hermit? It seems that everything is initialized with actions

oren-zohar: It is preferable we init with hermit instead of actions, to make the actions flow as close as possible to our dev environment. Basically to avoid cases where it works in the CI but fails locally

oren-zohar: I still don't get why we need to init python here separately (also the action takes 0 seconds so I'm not sure what it's achieving)

.github/workflows/cnvm-ci.yml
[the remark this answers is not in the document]
oren-zohar: ```poetry run pytest -k "cnvm"``` works locally? with the python that's configured by hermit?

.github/workflows/cnvm-ci.yml
orestisfl: Agreed. In general, it seems that we have many workflows that need to repeat the same steps. Should we try to centralize the setup and share it across different jobs?

.github/workflows/cnvm-ci.yml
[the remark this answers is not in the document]
amirbenun: poetry can't be installed with hermit, also locally I had to do some manual setup

.github/workflows/cnvm-ci.yml
amirbenun: We switch to python3.9 for installing `poetry` that for some reason can't be installed with the python version that provided by hermit.

orestisfl:
> I'm afraid 22 minutes for a pre-merge test is too long :(. Let's see how can we improve the runtime, or set it to a post-merge until then

I wonder if the go build (7m) can be sped up by caching?

oren-zohar:
> I wonder if the go build (7m) can be sped up by caching?

@orestisfl I think this is the only part where we can't cache, as we have to build on every change to cloudbeat, which is most of the PRs to this repo. Unless I'm missing something?

.github/workflows/cnvm-ci.yml
oren-zohar: what errors are you getting? did it work locally?

oren-zohar: > Agreed. In general, it seems that we have many workflows that need to repeat the same steps. Should we try to centralize the setup and share it across different jobs?

which steps? init hermit?

orestisfl:
> > I wonder if the go build (7m) can be sped up by caching?
> 
> @orestisfl I think this is the only part where we can't cache, as we have to build on every change to cloudbeat, which is most of the PRs to this repo. Unless I'm missing something?

Caching for go is based on the hash of the `go.sum` file and should speed up building by restoring dependencies and intermediate build outputs. I think it should work if set up correctly: https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs

.github/workflows/cnvm-ci.yml
orestisfl: > > Agreed. In general, it seems that we have many workflows that need to repeat the same steps. Should we try to centralize the setup and share it across different jobs?
> 
> @orestisfl which steps? init hermit?

hermit, setup go, build cloudbeat. In other pipelines there are also some duplicated docker image builds.

oren-zohar (approved):

PR #1148, opened 2023-07-20
Fix CloudFormation deployment config file

amirbenun:
### Summary of your changes
The `just deploy-cloudformation` command uses a config file to get the deployment configuration.
The config file has recently changed from `.env` to `config.env` (or `config.json`, `config.yml`) and we want to keep it that way as the sanity job is dependent on that.

### Screenshots
image

### Checklist


gurevichdmitry (approved):
Approved

PR #1154, opened 2023-07-25
golangci-lint: Make order of imports deterministic

orestisfl:
### Summary of your changes

Replaces goimports with gci: https://github.com/daixiang0/gci


uri-weisman (approved):
👑

PR #1156, opened 2023-07-25
worker disk space cleanup

oren-zohar:
### Summary of your changes
Fixing K8s CI issue by cleaning up disk space in the runner

### Checklist


oren-zohar:
Seems like it adds around 4 minutes to the CI, I'll see how can I optimize it first

orestisfl:
> Seems like it adds around 4 minutes to the CI, I'll see how can I optimize it first

The `large-packages` option of the action you were using before might help speed up if disabled: https://github.com/jlumbroso/free-disk-space/blob/4d9e71b726748f254fe64fa44d273194bd18ec91/action.yml#L171-L186

oren-zohar:
> > Seems like it adds around 4 minutes to the CI, I'll see how can I optimize it first
> 
> The `large-packages` option of the action you were using before might help speed up if disabled: https://github.com/jlumbroso/free-disk-space/blob/4d9e71b726748f254fe64fa44d273194bd18ec91/action.yml#L171-L186

Saved 5 seconds :metal:

.github/workflows/eks-ci.yml
```diff
@@ -65,7 +65,6 @@ jobs:
           android: true
           dotnet: true
           haskell: true
-          large-packages: true
```
orestisfl: It defaults to `true`, just deleting it won't do anything

oren-zohar: Much faster. Let's see if the freed up space is enough
image

orestisfl (approved):

PR #1166, opened 2023-07-26
fix agent tests

gurevichdmitry:
### Summary of your changes

This pull request (PR) resolves broken and skipped agent tests.

### Checklist


.github/workflows/k8s-ci.yml
```diff
@@ -24,9 +24,21 @@ concurrency:
 jobs:
   Build:
     name: Build
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
```
orestisfl, line 27: Any idea why this is necessary? Apologies if this has been explained before

orestisfl (approved):
I'd suggest finding a solution for 22.04 if necessary but it can also be done in a future task

.github/workflows/k8s-ci.yml
gurevichdmitry: @orestisfl, before running integration tests for agents, we are preparing a new image based on the latest agent's snapshot from artifactory and the most recent cloudbeat binary, which replaces the existing binary in the snapshot. However, when running tests with an image built on Ubuntu 22.04, we encounter cloudbeat errors in the log, stating that the version GLIBC_2.34 and GLIBC_2.32 are not found and required by /usr/share/elastic-agent/data/elastic-agent-8ea940/components/cloudbeat. Ubuntu 22.04 uses GLIBC version 2.35-0ubuntu3, while our packaging workflow runs on top of Ubuntu 20.04 with a different GLIBC version.

orestisfl: @gurevichdmitry thanks, as a guess this is probably because the cloudbeat binary is built on a different system. It might be possible to avoid the problem while building by setting `export CGO_ENABLED=0`

orestisfl: Perhaps a good follow-up task is to update all our pipelines to 22.04

gurevichdmitry: Follow-up [task](https://github.com/elastic/cloudbeat/issues/1173) for updating workflows and pipelines is opened

oren-zohar (approved):

PR #1171, opened 2023-07-28
Update golang-ci-lint settings

orestisfl:
### Summary of your changes

- Increase timeout since we hit the 10m limit when the cache fetching fails
  - The average run should still be shorter than that if caching works
  - On the topic of caching, we are constantly exceeding the 10GB GitHub limit, maybe we should consider doing something for that: https://github.com/elastic/cloudbeat/actions/caches
image
- Enable all `govet` rules and fix issues with shadowing
- Deletes `resources/providers/awslib/iam/test_data.go` since it can contained entirely within `resources/providers/awslib/iam/user_test.go`. Problem with the `test_data.go` file is that since it doesn't end with `_test.go`, those variables are reachable from non-test code.

### Checklist


uri-weisman (approved):

oren-zohar:
> On the topic of caching, we are constantly exceeding the 10GB GitHub limit

@orestisfl lets take if offline when you come back

.github/workflows/unit-test.yml
```diff
@@ -120,7 +120,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3.6.0
         with:
           version: latest
-          args: --timeout=10m --whole-files -v
```
oren-zohar, line 123: I think we need to see which check takes the longest and maybe we should consider moving it post-merge

orestisfl:
> > On the topic of caching, we are constantly exceeding the 10GB GitHub limit
> 
> @orestisfl lets take it offline when you come back

I created https://github.com/elastic/cloudbeat/issues/1175

PR #1169, opened 2023-07-27
 Refactor benchmark.go to remove `Dependencies` struct

orestisfl:
### Summary of your changes

See #1138 for context.

This is based on the #1161 branch, so that PR needs to be merged first.

### Related Issues

- Blocked by: #1161
- Fixes: #1138 
- Related to #1093 

### Checklist
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added the necessary README/documentation (if appropriate)

PR #1172, opened 2023-07-30
Make gcp benchmark dependencies mockable

uri-weisman:
### Summary of your changes
1. Shift gcp auth util and the inventory service to become a dependency in order to be mockable.
2. Modify the credential util to support the case in which no credentials have been provided.
As a result, the GCP SDK will use ADC (application default creds) and will use the compute instance's SA.

### Related Issues
- Blocked by: https://github.com/elastic/cloudbeat/pull/1169 (built on top of this branch)

### Checklist


orestisfl (approved):

oren-zohar:
@uri-weisman test coverage decreased?

flavors/benchmark/benchmark.go
```diff
@@ -69,7 +71,9 @@ func NewBenchmark(cfg *config.Config) (Benchmark, error) {
 		}, nil
 	case config.CIS_GCP:
 		return &GCP{
-			IdentityProvider: &identity.Provider{},
+			IdentityProvider:     &identity.Provider{},
+			CfgProvider:          &auth.ConfigProvider{},
+			inventoryInitializer: &inventory.ProviderInitializer{},
```
orestisfl, line 76: (not actionable in this PR observation) We don't really need `GCP` or any other `Benchmark` implementation to be exported. Until we actually need to do that, we can make everything private.

uri-weisman:
> @uri-weisman test coverage decreased?

@oren-zohar less than 0.1% percent 🙈

oren-zohar:
@uri-weisman I just mentioned this since it is counterintuitive, as this PR is around making more testing possible, no?

uri-weisman:
> @uri-weisman I just mentioned this since it is counterintuitive, as this PR is around making more testing possible, no?

This PR was all about making the utility's API calls mockable, so they wouldn't actually run. Though we wanted to improve code coverage, this PR didn't have many logic changes. The reduction in coverage was mainly due to shifts in code blocks.

Issue #1120, opened 2023-07-09
[CIS GCP] Create a data provider


uri-weisman:
**Motivation**
Enrich events with GCP information, such as organization id, folder id, and project id.
Field names and mapping should follow ECS.

**Definition of done**
What needs to be completed at the end of this task
- [ ] `cloud.account.id` is filled with gcp's organization id.
- [ ] `cloud.account.name` is filled with gcp's org display name.
- [ ] `cloud.project.id` is filled with gcp's project id.
- [ ] `cloud.provider` is filled with "gcp"
- [ ] `cloud.project.name` is filled with gcp's project name

**Related tasks/epics**
- https://www.elastic.co/guide/en/ecs/master/ecs-cloud.html

uri-weisman:
Since we are currently focusing on resource fetching at the project level, I will wait to obtain the data for `cloud.account.id` and `cloud.account.name` until we have well-defined requirements for fetching at the organization level.

PR #1178, opened 2023-08-01
Update GCP's ECS fields

uri-weisman:
### Summary of your changes
Assign GCP fields to `cloud.account.id` and `cloud.account.name` to align with AWS CSPM. 
We communicate with the ECS team to modify and extend the cloud schema in the future

### Screenshot/Data
Screenshot 2023-08-01 at 12 29 18

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/1120

### Checklist

### DIFF
--- a/dataprovider/providers/cloud/data_provider.go
+++ b/dataprovider/providers/cloud/data_provider.go
@@ -32,16 +32,12 @@ const (
 	cloudAccountNameField = "cloud.account.name"
 	cloudProviderField    = "cloud.provider"
 	cloudRegionField      = "cloud.region"
-	cloudProjectIdField   = "cloud.project.id"
-	cloudProjectNameField = "cloud.project.name"
 )
 
 type Identity struct {
 	Provider     string
 	Account      string
 	AccountAlias string
-	ProjectId    string
-	ProjectName  string
 }
 
 type DataProvider struct {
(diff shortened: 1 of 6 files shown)


dataprovider/providers/cloud/data_provider.go
```diff
@@ -32,25 +32,19 @@ const (
 	cloudAccountNameField = "cloud.account.name"
 	cloudProviderField    = "cloud.provider"
 	cloudRegionField      = "cloud.region"
-	cloudProjectIdField   = "cloud.project.id"
```
orestisfl, line 35: From the ECS docs, I think there will be a clear expectation from users that `cloud.project.id` should be populated. WDYT about populating both `cloud.project.*` and `cloud.account.*` with the same data?

uri-weisman: I proposed this in our daily, but @oren-zohar believes it might be redundant.

tinnytintin10: @uri-weisman @oren-zohar let's go with just populating the cloud.account.* fields for now. The purpose of ECS is to normalize heterogeneous data sources. In this case, by having both project and account, it's doing exactly the opposite. I will work with someone on the ECS team to understand why this is and get back to y'all 👍🏾

orestisfl (approved):

PR #1190, opened 2023-08-03
AWS functional tests migration

gurevichdmitry:
### Summary of your changes

AWS functional tests have been updated to make them compatible with the new AWS test account environment.

The most recent test run is [available](https://github.com/gurevichdmitry/cloudbeat/actions/runs/5793882935).


.github/workflows/k8s-ci.yml
```diff
@@ -210,10 +210,9 @@ jobs:
           - test-target: pre_merge
             range: ""
             values_file: tests/deploy/values/ci.yml
-          # Skipping until https://github.com/elastic/security-team/issues/6757 is resolved
-          # - test-target: aws
-          #   range: ""
-          #   values_file: tests/deploy/values/ci-aws-services.yml
+          - test-target: 'aws and not aws_s3'
```
gurevichdmitry, line 213: Implemented skip for AWS S3 functional tests.
A new [task](https://github.com/elastic/security-team/issues/7282) has been created.

amirbenun: Can you elaborate what's the meaning of the value `aws and not aws_s3`?

gurevichdmitry: We run tests using markers (-m) or keyword expressions (-k). 
When using markers, our command looks like `pytest -m "pre_merge"`. 

With keyword expressions, we can define patterns to select specific tests for execution. The value `aws and not aws_s3` is a way to filter tests based on their names. The expression `aws and not aws_s3` is used to execute all tests whose names contain the term `aws` but do not contain the term `aws_s3`.

amirbenun: So if I understand correct, all the values under `test-target` variable are pytest markers?

gurevichdmitry: > So if I understand correct, all the values under `test-target` variable are pytest markers?

exactly

tests/deploy/k8s-cloudbeat-tests/templates/cloudbeat-test-configmap.yml
```diff
@@ -14,7 +14,7 @@ data:
   {{ $key }}: {{ $value }}
   {{- end }}
   MARKER_CMD: {{ .Values.testData.markerCmd}}
-  TEST_MARKER: {{ .Values.testData.marker }}
+  TEST_MARKER: "{{ .Values.testData.marker }}"
```
oren-zohar, line 17: any reason to surround only the test marker by quotes? is it to support the `and` keywords?

oren-zohar (approved):

tests/deploy/k8s-cloudbeat-tests/templates/cloudbeat-test-configmap.yml
gurevichdmitry: Indeed, the reason for surrounding the test marker with quotes is to support keyword expressions like aws and not aws_s3. By enclosing the expression in quotes, it ensures that the entire expression is treated as a single argument. Without the quotes, the command could potentially interpret the expression as multiple arguments, leading to unexpected behavior.

Issue #1132, opened 2023-07-13
CloudFormation template add agent tags


amirbenun:
### Motivation
When deploying a CloudFormation template we have no way to tell in the fleet that the elastic-agent was deployed with the stack creation. Then, it makes it hard to flag the user on an available updates in the CloudFormation template.
The `elastic-agent install` command supports a `--tag` CLI parameter which we can leverage in order to save important information about the elastic-agent.
As a first step, we should tag the elastic-agent with the following:
- `cft-arn:` - CloudFormation stack ARN
- `cft-version:` - CloudFormation template version

### Definition of done
- [ ] Compose the stack ARN from the [pseudo parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html) provided by AWS
- [ ] Define a new parameter for the template version which will be replaced by the CI
- [ ] Tag the agent with those values prefixed with `cft-arn:` and `cft-version:` respectively
- [ ] Changes should apply for CSPM and CNVM
- [ ] Optionally: test the feature as part of the sanity job

### Testing
Make sure agents are properly tagged via FleetUI

### Related tasks/epics
- https://github.com/elastic/security-team/issues/5701

amirbenun:
Example for elastic-agent tags:
<img width="457" alt="image" src="https://github.com/elastic/cloudbeat/assets/34831306/cb10f80b-a3d5-48ad-85b3-da80a3163707">

amirbenun:
### Background
When we publish a CloudFormation template to an S3 bucket we publish it with the date and time suffix to have versioning.

### Problem
While working on the task I separated the cloudformation upload script into a separate file `scripts/publish_cft.sh`.
Since the new script also uses the current date, when it gets called by the CI, the creation date of the two CSPM templates is not necessarily identical.

When I had to update the integration with the new template I realized that there is an assumption that the two templates were named with the exact same time (differs only by the `ACCOUNT_TYPE`).
```
https://elastic-cspm-cft.s3.eu-central-1.amazonaws.com/cloudformation-cspm-ACCOUNT_TYPE-8.10.0-2023-07-24-15-10-19.yml
```

### Available solutions
1. The script will accept the date as a parameter, meaning we will keep publishing the two files with the same date.
2. Instead of having a single variable `cloud_formation_template`, we can hold two different variables: `cloud_formation_template_single_account` and `cloud_formation_template_organization_account`

[...]

PR #1202, opened 2023-08-07
Test CloudFormation agents tags

amirbenun:
### Summary of your changes
Verify on the sanity tests the an agent created using CloudFormation has the proper tags.

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/1132

### Checklist


deploy/test-environments/fleet_api/src/agents_enrolled.py
```diff
@@ -12,6 +13,21 @@
...
+    def __init__(self, count: int, tags: list[str]) -> None:
+        """
+        Args:
+            count (int): Number of agents to be enrolled.
+            tags (list[str]): Tags of the agent.
+        """
+        self.count = count
+        self.tags = tags
+
+
```
gurevichdmitry, line 30: It appears that this class is exclusively utilized for data. Therefore, instead of employing a standard class, you could consider implementing a [dataclass](https://docs.python.org/3/library/dataclasses.html)

We already have an implementation example available [here](https://github.com/elastic/cloudbeat/blob/main/tests/product/tests/data/eks_test_case.py)

deploy/test-environments/fleet_api/src/agents_enrolled.py
```diff
@@ -53,6 +67,40 @@ def verify_agents_enrolled() -> bool:
...
+        expected_tags = expected_agents[agent.policy_id].tags
+    for pattern in expected_tags:
+        pattern_exist = False
+        for tag in agent.tags:
+            if re.match(pattern, tag):
+                pattern_exist = True
+                break
+        if not pattern_exist:
+            logger.info(f"Agent {agent.id} does not have the expected tag {pattern}")
+            return False
```
gurevichdmitry, line 85: nit: Since this behavior is unexpected, it might be better to use `logger.warn()` or `logger.error()`

deploy/test-environments/fleet_api/src/agents_enrolled.py
```diff
@@ -36,23 +48,55 @@ def get_actual_agents() -> dict:
...
-    for policy_id, expected_count in expected.items():
+    for policy_id, expected_agents in expected.items():
         if policy_id not in actual:
             result = False
             logger.info(f"Policy {policy_id} not found in the actual agents mapping")
-        elif actual[policy_id] != expected_count:
+        elif actual[policy_id] != expected_agents.count:
             result = False
-            logger.info(f"Policy {policy_id} expected {expected_count} agents, but got {actual[policy_id]}")
+            logger.info(f"Policy {policy_id} expected {expected_agents.count} agents, but got {actual[policy_id]}")
```
gurevichdmitry, line 62: In the deployment run printout, there is currently no information available regarding the number of agents that are actively running. It could be beneficial to include this information for better context. In the case of a successful deployment, considering adding a log statement like
``` Python
else:
    logger.info(f"Policy id: {policy_id}, expected: {expected_agents.count}, actual: {actual[policy_id]}") would provide valuable insights.
```

gurevichdmitry (approved):
@amirbenun LGTM

deploy/test-environments/fleet_api/src/agents_enrolled.py
```diff
@@ -36,23 +48,57 @@ def get_actual_agents() -> dict:
...
+        expected_tags = expected_agents[agent.policy_id].tags
+    for pattern in expected_tags:
+        pattern_exist = False
+        for tag in agent.tags:
+            if re.match(pattern, tag):
+                pattern_exist = True
+                break
+        if not pattern_exist:
+            logger.warning(f"Agent {agent.id} does not have the expected tag {pattern}")
+            return False
```
oren-zohar, line 83: `any()` returns True if any of the tags in agent.tags match the current pattern.
```suggestion
    for pattern in expected_tags:
        if not any(re.match(pattern, tag) for tag in agent.tags):
            logger.warning(f"Agent {agent.id} does not have the expected tag {pattern}")
            return False
```

deploy/test-environments/fleet_api/src/agents_enrolled.py
```diff
@@ -36,23 +48,57 @@ def get_actual_agents() -> dict:
...
     return result
 
 
+def verify_agent_tags(agent, expected_agents) -> bool:
+    """
+    Verify that the agent has the expected tags
+    """
+    expected_tags = []
+    if agent.policy_id in expected_agents:
+        expected_tags = expected_agents[agent.policy_id].tags
```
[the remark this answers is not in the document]
amirbenun: Will it really work as you suggest?
If the policy id doesn't exist we will get `[]` which doesn't have `.tags`.

PR #1210, opened 2023-08-08
Add merge queue ci

orouz:
this PR makes our `pull_request/target` workflows also run on on `merge_group` event, only for `main` branch

from the [docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#triggering-merge-group-checks-with-github-actions:~:text=If%20your%20repository,and%20push%20events.):

> If your repository uses GitHub Actions to perform required checks on pull requests in your repository, you need to update the workflows to include the merge_group event as an additional trigger. Otherwise, status checks will not be triggered when you add a pull request to a merge queue. The merge will fail as the required status check will not be reported. The merge_group event is separate from the pull_request and push events.

after enabling this new setting we ran into this ^ issue, where CI checks don't run and wait for timeout. this PR is meant to fix that.


gurevichdmitry (approved):
@orouz Approved. This is a great initiative, and in my opinion, it would be a nice addition.

On one hand, it's a good idea because putting your PR in the queue saves time by not requiring checks and merges from the main branch (as it will be done automatically). On the other hand, considering the description's statement that "Using a merge queue is particularly useful on branches that have a relatively high number of pull requests merging each day from many different users," I believe this might not fully apply to our situation. We typically have only a few PRs per day.

Furthermore, we have experienced some flakiness on the testing side, which could become problematic and require customization to determine which tests should be mandatory and which ones can be skipped. Additionally, it's important to note that in case of required test failures, the PR will be rejected and removed from the merge queue. Also, with the merge queue, you will not have the ability to retry failed cases as you might do when merging manually.

orouz:
> We typically have only a few PRs per day.

true, but sometimes we have more and when we do have multiple PRs in a short period of time, every time someone merges before you you need to pay a time tax to upgrade your branch and wait for CI. this seems like it would solve it. 

> Furthermore, we have experienced some flakiness on the testing side, which could become problematic and require customization to determine which tests should be mandatory and which ones can be skipped. Additionally, it's important to note that in case of required test failures, the PR will be rejected and removed from the merge queue. Also, with the merge queue, you will not have the ability to retry failed cases as you might do when merging manually.

i'm ok with failing CI PRs to go back to manual merging. would be best if CI was not flaky. 
also, can you share a reference to where it says we can't retry on `merge_group` events? it's kinda implicit AFAIK

gurevichdmitry:
You're correct. The conclusion is drawn from the [Managing a Merge Queue](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue) section, where no specific option is mentioned for retrying on `merge_group` events. The absence of an explicit provision for this implies that such a feature might not be directly supported.

However, there is an option named **Only merge non-failing pull requests**, which lets you decide whether or not to select the checkbox. According to the documentation, not selecting this checkbox can be advantageous when dealing with intermittent test failures. It prevents false negatives from delaying the queue while still considering the combined set of changes in the merge group.

oren-zohar (commented):
please include [CloudFormation Integration Test](https://github.com/elastic/cloudbeat/pull/1233) too

oren-zohar (changes requested):
I don't understand how `merge_group` will work `pull_request_target`, can you please clarify @orouz?

from the docs:

> the changes in the pull request are grouped into a merge_group with the latest version of the base_branch as **well as changes from pull requests ahead of it in the queue**

orouz:
@oren-zohar 

i'm not sure i see why this would be an issue. for example, merging `cloudbeat/main` as well as `orouz/some_branch` to new PRs shouldn't create a problem as long as there aren't any conflicts. maybe i'm missing smth?

can you provide an example you think wouldn't work and explain why?

.github/workflows/cnvm-ci.yml
```diff
@@ -1,6 +1,8 @@
 name: CNVM-CI
 
 on:
+  merge_group:
+    branches: ['main']
   pull_request_target:
```
oren-zohar, line 6: please move every `pull_request_target` to post-merge instead of adding them to the merge queue

oren-zohar (changes requested):

orouz:
closing this as it's not critical for now. may get back to it in the future.

PR #1234, opened 2023-08-15
Extract resource id

orouz:
### Summary of your changes

currently we're sending a GCP's asset name as both the `resource.name` and `resource.id`. this ends up being displayed in the UI and is repetitive. 

this PR extracts the last bit of the asset name to be used as the ID. 
worth noting: the actual full ID is the asset name, which is made up from the asset's ancestors 

a different solution may remove the resource name/id from the UI. 

an example of how a resource will be displayed after this PR:
`resource.id` - `//bigquery.googleapis.com/projects/elastic-security-test/datasets/test_cspm`
`resource.name` - `test_cspm`

see https://cloud.google.com/apis/design/resource_names#resource_id


resources/fetching/fetchers/gcp/assets_fetcher.go
```diff
@@ -144,6 +144,14 @@ func (r *GcpAsset) GetMetadata() (fetching.ResourceMetadata, error) {
 
 func (r *GcpAsset) GetElasticCommonData() any { return nil }
 
+// a GCP asset name is made up of its ancestors
+// the resource id is the last part of the name
+// see https://cloud.google.com/apis/design/resource_names#resource_id
```
uri-weisman, line 149: One problem I can see is that it might create duplicate resource IDs, leading to clashes in IDs. This could result in missing findings from the latest findings index.

oren-zohar: ancestors?

oren-zohar: yaani parents?

uri-weisman: Do you mean we should represent the direct parent in the resource id?

uri-weisman (approved):
LGTM

orouz:
@oren-zohar this needs to be backported to 8.10, right?

oren-zohar:
> @oren-zohar this needs to be backported to 8.10, right?

Since BC1 was already published I think it's not worth the risk. By the way, can you take a look at the technical index and see if we need to update anything related to this change?

orouz:
> can you take a look at the technical index and see if we need to update anything related to this change?

| Field                  |                    Type                     | Description |                                                                                                                                                                                                                                                  
|----------------|----------------------------|-------------|
| `resource.id`          |             Unique resource ID              | uuid that is generated according to resource type. file-system: uuid(cluster.id, node.id, file.path, file.creation_date). process: uuid(cluster.id, node.id, process.namespace_inode, process.pid, process.start_time). kube-api: k8s object metadata.uuid.   |                               |
| `resource.name`        |                   String                    | A human readable representation of the resource


we're still aligned to this description. `resource.name` is `asset.name` which is the full unique identifier. `resource.name` is `asset.name[last_part]` which is a sort-of a human-readable form for the resource name.

PR #1247, opened 2023-08-17
CNVM CI remove pull_request_target trigger

amirbenun:
### Summary of your changes
CNVM CI - Change trigger from `pull_request_target` to `pull_request`.
`pull_request` trigger doesn't have write secret permissions which is being used by `aws-actions/configure-aws-credentials`.
As part of this PR we also remove this action and use environment variables instead.

### Checklist


amirbenun:
This fails because a github workflow with context of `pull_request` coming from a fork doesn't have access to the repository secrets.
More info: https://github.com/pypa/gh-action-pypi-publish/discussions/49

Issue #1236, opened 2023-08-16
[CIS GCP] Enrich events with project and org info


uri-weisman:
**Motivation**
Until now, we only focused on a single project, so CIS GCP events get linked with that project's ID.
When we support an organization level, we'll gather resources from multiple projects, which means we need to change how this works.
Also, we'll include the relevant organizational details for each event.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Each event is enriched with its project info.
- [ ] Each event is enriched with the org info (`cloud.organization.id` & `cloud.organization.name`)

**Related tasks/epics**
- related: https://github.com/elastic/cloudbeat/issues/1177

kfirpeled:
project info should be `cloud.account.id` and `cloud.account.name`

Omolola-Akinleye:
Events are enriched for project and org info
![Image](https://github.com/elastic/cloudbeat/assets/17135495/c7e3661f-dea3-417a-8ae2-cb37179e1654)

Issue #1232, opened 2023-08-15
[CIS GCP] Make asset inventory's scope configurable


uri-weisman:
**Motivation**
Currently, the asset fetcher supports only [projects scope](https://github.com/elastic/cloudbeat/blob/main/resources/providers/gcplib/inventory/provider.go#L101), we need to make the scope dynamic based on the configuration provided by Fleet.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Cloudbeat is able to fetch resources at the organization and project level.

**Out of scope**
What should not be included in this task

**Related tasks/epics**
Reference related issues and epics

PR #1262, opened 2023-08-22
[GCP] Add support for fetching organization resources

orouz:
### Summary of your changes

- added support for GCP organization
   - added `OrganizationId`  and `AccountType` fields to `GcpConfig` (these will be provided from integration via kibana)
   - scoping-wise, `GcpFactoryConfig` will only include a `Parent` field. 
- removed gcp's `identity_provider` as for org scope we need to enrich events for each project. this is done with caching in place so we don't issue redundant request for asset identification. 

### Screenshot/Data

- enrichment for a single project: 

Screenshot 2023-08-29 at 15 12 14

- enrichment for 2 projects, 1 org:

|Screenshot 2023-08-28 at 16 34 32 	| Screenshot 2023-08-28 at 16 34 25 	|
|------	|-------	|

### Related Issues
- closes https://github.com/elastic/cloudbeat/issues/1232
- closes https://github.com/elastic/cloudbeat/issues/1236

### Checklist


resources/fetching/fetchers/gcp/monitoring_fetcher.go
```diff
@@ -84,7 +84,7 @@ func (f *GcpMonitoringFetcher) Stop() {
 }
 
 func (g *GcpMonitoringAsset) GetMetadata() (fetching.ResourceMetadata, error) {
-	id := fmt.Sprintf("%s-%s", g.subType, g.Asset.ProjectId)
+	id := fmt.Sprintf("%s-%s", g.subType, g.Asset.ParentId)
```
uri-weisman, line 87: We must use a project id here.
In case we'll use an org id we'll get a resource id collision upon assessing multiple projects.

resources/providers/gcplib/auth/credentials.go
```diff
@@ -49,37 +53,47 @@ type ConfigProvider struct {
 }
 
 func (p *ConfigProvider) GetGcpClientConfig(ctx context.Context, cfg config.GcpConfig, log *logp.Logger) (*GcpFactoryConfig, error) {
+	// TODO: is this for development only? if so add a comment
```
uri-weisman, line 56: Not for development, it's for cases where we fetch the ADC from the service account attached to compute instance, the one running elastic agent.

resources/providers/gcplib/inventory/provider.go
```diff
@@ -196,3 +274,112 @@ func mergeAssetContentType(assets []*assetpb.Asset) []*assetpb.Asset {
...
+		return "", "", "", errors.New("failed to get project/organization name")
+	}
+
+	// values in index 0 are the same for all indices.
+	// verified by the lo.FindUniquesBy() above
+	return projectNames[0].ProjectName,
+		organizationIds[0].OrganizationId,
+		organizationNames[0].OrganizationName,
+		nil
+}
```
orouz, line 385: this is a bit excessive, and was done to be extra-safe, but we can alternately just take the data from the first element, it's already guaranteed to be the same for all values, regardless of the `FindUniquesBy`, because we've grouped assets by project

config/config.go
```diff
@@ -65,7 +66,15 @@ type AwsConfig struct {
...
+	ProjectId string `config:"project_id"`
+
+	// empty for SingleAccount
+	OrganizationId string `config:"organization_id"`
+
+	// SingleAccount or OrganizationAccount
+	AccountType string `config:"account_type"`
+
 	GcpClientOpt `config:"credentials"`
 }
```
orouz, line 79: we could remove `account_type` and only rely on the presense of either fields: `project_id` / `organization_id`, but AWS has this field so this is consistent with cloud vendors org support

config/config.go
```diff
@@ -148,3 +150,27 @@ func isSupportedBenchmark(benchmark string) bool {
 	}
 	return false
 }
+
+// TODO: move config validation to libs? (aws/gcp)
+// TODO: why is "" a valid case?
+func isSupportedAccountType(c Config) error {
```
orouz, line 156: 1 - not sure why we're allowing `""` as a valid case
2 - i think we should move this bit so each cloud vendor lib validates its own config, instead of just validating a single field here. (note `account_type` needs to be accompnied by either `project_id` or `organization_id`)

resources/fetching/fetchers/gcp/assets_fetcher.go
```diff
@@ -137,20 +136,33 @@ func (r *GcpAsset) GetMetadata() (fetching.ResourceMetadata, error) {
...
-		ID:      r.Asset.Name,
-		Type:    r.Type,
-		SubType: r.SubType,
-		Name:    getAssetResourceName(r.Asset),
-		Region:  region,
+		ID:        r.Asset.Name,
+		Type:      r.Type,
+		SubType:   r.SubType,
+		Name:      getAssetResourceName(r.Asset),
+		Region:    region,
```
orouz, line 143: we should remove `region` from `fetching.ResourceMetadata` and send it as part of `ecs.cloud`

resources/fetching/fetchers/gcp/monitoring_fetcher.go
```diff
@@ -57,23 +57,25 @@ var monitoringAssetTypes = map[string][]string{
 func (f *GcpMonitoringFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
 	f.log.Info("Starting GcpMonitoringFetcher.Fetch")
 
-	monitoringAsset, err := f.provider.ListMonitoringAssets(monitoringAssetTypes)
+	monitoringAssets, err := f.provider.ListMonitoringAssets(monitoringAssetTypes)
```
orouz, line 60: the main change in the monitoring fetcher is that it returns an array of assets instead of a single asset. this is because when scope is organization, we may have multiple projects, so now we get an array of assets where each asset is a different project. (note each asset consists of 2 asset types and a project id)

resources/fetching/fetchers/gcp/monitoring_fetcher.go
```diff
@@ -86,16 +88,29 @@ func (f *GcpMonitoringFetcher) Stop() {
...
+		"account": map[string]interface{}{
+			"id":   g.Asset.ProjectId,
+			"name": g.Asset.ProjectName,
+		},
+		"organization": map[string]interface{}{
+			"id":   g.Asset.OrganizationId,
+			"name": g.Asset.OrganizationName,
+		},
+	}
+}
```
orouz, line 116: this will change a bit once #1274 is merged

resources/providers/gcplib/auth/credentials_test.go
orouz, line 1: main change in this file is the addition of multiple configs, so each existing test now runs twice - 1 for each account type. also added more tests for config validation

resources/providers/gcplib/inventory/provider.go
```diff
@@ -72,64 +92,94 @@ type ProviderInitializerAPI interface {
...
 		wg.Done()
 	}()
 
 	wg.Wait()
-
-	return mergeAssetContentType(append(resourceAssets, policyAssets...)), nil
+	var assets []*assetpb.Asset
+	assets = append(append(assets, resourceAssets...), policyAssets...)
+	mergedAssets := mergeAssetContentType(assets)
+	extendedAssets := extendAssets(p.ctx, p.crm, mergedAssets)
```
orouz, line 177: this is where we merge assets of the same name (with diff asset type) and then extend them with display names

orouz (commented):
left some comments to highlight changes in this PR

resources/fetching/fetchers/gcp/assets_fetcher.go
```diff
@@ -145,12 +144,26 @@ func (r *GcpAsset) GetMetadata() (fetching.ResourceMetadata, error) {
...
+				"id":   r.Asset.ProjectId,
+				"name": r.Asset.ProjectName,
+			},
+			"organization": map[string]interface{}{
+				"id":   r.Asset.OrganizationId,
+				"name": r.Asset.OrganizationName,
+			},
+		},
+	}, nil
+}
```
orouz, line 161: @amirbenun is this aligned with the new format?

config/config.go
```diff
@@ -148,3 +150,27 @@ func isSupportedBenchmark(benchmark string) bool {
...
+			c.CloudConfig.Aws.AccountType,
+		))
+	}
+
+	if !lo.Contains(collection, c.CloudConfig.Gcp.AccountType) {
+		return launcher.NewUnhealthyError(fmt.Sprintf(
+			"gcp.account_type '%s' is not supported",
+			c.CloudConfig.Gcp.AccountType,
+		))
+	}
```
uri-weisman, line 173: I'm not sure it's needed as we already perform this kind of validation in the credentials BL.

resources/providers/gcplib/inventory/provider.go
```diff
@@ -19,29 +19,54 @@ package inventory
...
 )
 
 type Provider struct {
-	log    *logp.Logger
-	client *GcpClientWrapper
-	ctx    context.Context
-	Config auth.GcpFactoryConfig
+	log       *logp.Logger
+	ctx       context.Context
+	Config    auth.GcpFactoryConfig
```
uri-weisman, line 43: ?
```suggestion
	config    auth.GcpFactoryConfig
```

resources/providers/gcplib/inventory/provider.go
```diff
@@ -189,10 +258,109 @@ func mergeAssetContentType(assets []*assetpb.Asset) []*assetpb.Asset {
...
+	return strings.Split(last, "/")[1]
+}
+
+func getProjectId(ancestors []string) string {
+	return strings.Split(ancestors[0], "/")[1]
+}
+
+func getProjectLogMetricAssets(projectAssets []*ExtendedGcpAsset) []*ExtendedGcpAsset {
+	return lo.Filter(projectAssets, func(asset *ExtendedGcpAsset, _ int) bool {
+		return asset.AssetType == "logging.googleapis.com/LogMetric"
```
uri-weisman, line 346: let's make "logging.googleapis.com/LogMetric" a const

resources/providers/gcplib/inventory/provider.go
```diff
@@ -189,10 +258,109 @@ func mergeAssetContentType(assets []*assetpb.Asset) []*assetpb.Asset {
...
+	})
+}
+
+func getProjectAlertPolicyAssets(projectAssets []*ExtendedGcpAsset) []*ExtendedGcpAsset {
+	return lo.Filter(projectAssets, func(asset *ExtendedGcpAsset, _ int) bool {
+		return asset.AssetType == "monitoring.googleapis.com/AlertPolicy"
```
uri-weisman, line 352: same as above

resources/providers/gcplib/inventory/provider.go
```diff
@@ -189,10 +258,109 @@ func mergeAssetContentType(assets []*assetpb.Asset) []*assetpb.Asset {
...
+	return &fetching.EcsGcpCloud{
+		ProjectId:        keys.projectId,
+		ProjectName:      projectName,
+		OrganizationId:   keys.orgId,
+		OrganizationName: orgName,
+	}
+}
+
+func getOrganizationId(ancestors []string) string {
+	last := ancestors[len(ancestors)-1]
```
uri-weisman, line 336: what happens if we asses a project that is not associated with an organization?

resources/providers/gcplib/inventory/provider.go
```diff
@@ -189,10 +258,109 @@ func mergeAssetContentType(assets []*assetpb.Asset) []*assetpb.Asset {
...
 	for _, asset := range resultsMap {
 		results = append(results, asset)
 	}
 	return results
 }
+
+// extends the assets with the project and organization display name
+func extendAssets(ctx context.Context, crm *ResourceManagerWrapper, assets []*assetpb.Asset) []*ExtendedGcpAsset {
+	var extendedAssets []*ExtendedGcpAsset
+	var cache = make(map[string]*fetching.EcsGcpCloud)
```
uri-weisman, line 271: We create a new cache for every type, consider storing the cache at the provider level.

resources/providers/gcplib/inventory/provider.go
```diff
@@ -189,10 +258,109 @@ func mergeAssetContentType(assets []*assetpb.Asset) []*assetpb.Asset {
...
+	orgId := getOrganizationId(asset.Ancestors)
+	projectId := getProjectId(asset.Ancestors)
+	parentProject := fmt.Sprintf("projects/%s", projectId)
+	parentOrg := fmt.Sprintf("organizations/%s", orgId)
+	return GcpAssetIDs{
+		orgId:         orgId,
+		projectId:     projectId,
+		parentProject: parentProject,
+		parentOrg:     parentOrg,
+		cacheKey:      fmt.Sprintf("%s/%s", parentProject, parentOrg),
```
uri-weisman, line 308: It makes more sense to me to generate the cache key out of this function response and not include it in the `GcpAssetIDs` struct.

uri-weisman (changes requested):
Great job dude!

resources/providers/gcplib/auth/credentials.go
```diff
@@ -48,38 +49,40 @@ type ConfigProvider struct {
 	AuthProvider GoogleAuthProviderAPI
 }
 
+var ErrMissingOrgId = errors.New("organization ID is required for organization account type")
+var ErrInvalidCredentialsJSON = errors.New("invalid credentials JSON")
+var ErrProjectNotFound = errors.New("no project ID was found")
+
 func (p *ConfigProvider) GetGcpClientConfig(ctx context.Context, cfg config.GcpConfig, log *logp.Logger) (*GcpFactoryConfig, error) {
+	// used in cloud shell flow (and development)
```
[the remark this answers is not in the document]
orouz: the suggested comment reads pretty much the same as the function itself: `getDefaultCredentialsConfig`

including _when_ it is used, (cloud shell / development) is more informative. wdyt?

resources/providers/gcplib/inventory/provider.go
orouz: are you sure this is a real secnario ? i don't think users can do that. the `Ancestors` field description on the asset says an organization is always the last item. 

technically, what will happen here if it is a real case is that the project id would be the same as the org id, as both would access the same member

config/config.go
orestisfl: As it is, isn't one of these two always going to be `""`? Since either gcp or aws is going to be enabled.

orouz: i reverted my changes and kept the original aws account type check as it is. the GCP one is done at the credentials file where it also checks for matching IDs. 

> As it is, isn't one of these two always going to be ""? Since either gcp or aws is going to be enabled.

i think so, yeah. which i think is another reason to move this validation logic to libs.

resources/providers/gcplib/inventory/provider.go
orouz: we can add custom logic to omit the `orgId` and `orgName` if `len(ancestors)` is `1`. but i think the juice isn't worth the squeeze, because:
1. we can't use struct tags, so this means sprinkling a bunch of if statements which kinda pollutes the code for this edge case.
2. this scenario is unlikely to happen for companies, as those usually use organizations to scope projects. although, GCP does not require an organization.
3. in the case where there is no organization, IMO it's ok the return the org as the project, hardly an issue. 

wdyt?

resources/providers/gcplib/inventory/provider.go
uri-weisman: @orouz can we make it private?

uri-weisman (approved):
💪

PR #1266, opened 2023-08-24
[Create Environment] Add deployment name validation

gurevichdmitry:
### Summary of your changes

This PR introduces a new GitHub workflow action called "Check Deployment Name" to ensure the validity of deployment names. The action performs two critical checks:

- **Length Check**: It verifies that the deployment name is not longer than 20 characters, as this is the maximum allowed length.

- **Pattern Check**: It ensures that the deployment name adheres to the required pattern, which should start with a letter (a-z or A-Z) and can include letters, numbers, and hyphens.

This action is an important step in maintaining naming conventions and ensuring that deployment names align with our standards, particularly for CNVM deployments.

#### Documentation

Documentation

#### Length Check

Lenght Check

#### CNVM pattern check failure

Pattern Check

### Checklist

### DIFF
--- a/.github/workflows/test-environment.yml
+++ b/.github/workflows/test-environment.yml
@@ -61,6 +61,13 @@ jobs:
       STACK_VERSION: ${{ github.event.inputs.elk-stack-version }}
       CNVM_STACK_NAME: "${{ github.event.inputs.deployment_name }}-cnvm-sanity-test-stack"
     steps:
+      - name: Check out the repo
+        uses: actions/checkout@v3
+
+      - name: Init Hermit
+        run: ./bin/hermit env -r >> $GITHUB_ENV
+        working-directory: ./
+
       - name: Check Deployment Name
         run: |
           deployment_name="${{ github.event.inputs.deployment_name }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-environment.yml
```diff
@@ -10,6 +10,8 @@ on:
         description: |
           Name your environment (Only a-zA-Z0-9 and `-`).
           For example: john-8-7-2-June01'
+          Supported pattern: [a-zA-Z][-a-zA-Z0-9]*
+          Max chars: 20
```
oren-zohar, line 14: can we make this entire section shorter?

oren-zohar (approved):

.github/workflows/test-environment.yml
gurevichdmitry: fixed

.github/workflows/test-environment.yml
```diff
@@ -66,6 +65,22 @@ jobs:
...
+          deployment_name="${{ github.event.inputs.deployment_name }}"
+
+          # Check length
+          if [ ${#deployment_name} -gt 20 ]; then
+            echo "error: Deployment name is too long (max 20 characters)"
+            exit 1
+          fi
+
+          # Check pattern required for CNVM deployment
+          if ! [[ $deployment_name =~ ^[a-zA-Z][-a-zA-Z0-9]*$ ]]; then
```
amirbenun, line 79: You can use `{0,19}` instead of `*` which will also promise the string is no longer than 20 characters.

Issue #1121, opened 2023-07-10
[CIS GCP] Create a manifest file for Google deployment manager


uri-weisman:
**Motivation**
Creating a manifest to be deployed by the deployment manager, will smooth our onboarding process for CIS GCP.

**Definition of done**
- [ ] A configuration file for complete deployment automation (including service account creation and elastic agent deployment with CSPM integration) 
- [ ] A configuration file for only provisioning the necessary service account with proper permissions (users will manually deploy the elastic agent with CSPM integration).
- [ ] Enable the cloud asset inventory.
- [ ] Verify that the chosen instance type is supported across all regions.

**Out of scope**
UI and CI changes.

**Related tasks/epics**
- https://github.com/elastic/security-team/issues/7185

PR #1269, opened 2023-08-24
[CIS GCP] Improve onboarding script

uri-weisman:
### Summary of your changes
1. Use GCP's built-in roles instead of creating a custom role.
2. Add a success print at the end of the script
3. Do not update the default service account permissions in case they already exist.
4. Remove the trailing slash from the artifact server variable if present.

### Related Issues
- https://github.com/elastic/cloudbeat/issues/1121
- https://github.com/elastic/cloudbeat/issues/1270

### Checklist

### Links
- https://cloud.google.com/iam/docs/understanding-roles#cloudasset.viewer
- https://cloud.google.com/iam/docs/understanding-roles#browser


oren-zohar:
We shouldn't backport this to 8.10, removing the label

olegsu (approved):

PR #1272, opened 2023-08-27
Use gotestsum in CI

orouz:
### Summary of your changes

makes CI tests output a bit more readable 
see https://pkg.go.dev/github.com/IstrateM/gotestsum/pkg/gotestsum#section-readme

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -42,6 +42,7 @@ jobs:
 
       - name: Unit-Test
         run: |
+          go install gotest.tools/gotestsum
           GOOS=linux TEST_DIRECTORY=./...  gotestsum --format pkgname -- -race -coverpkg=./... -coverprofile=cover.out.tmp ./...
           cat cover.out.tmp | grep -v "mock_.*.go" > cover.out # remove mock files from coverage report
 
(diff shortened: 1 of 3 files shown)


.github/workflows/unit-test.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
       - name: Unit-Test
         run: |
-          GOOS=linux go test -v -race -coverpkg=./... -coverprofile=cover.out.tmp ./...
+          GOOS=linux TEST_DIRECTORY=./...  gotestsum --format pkgname -- -race -coverpkg=./... -coverprofile=cover.out.tmp ./...
```
oren-zohar, line 45: try adding `go get gotest.tools/gotestsum` here to see if it fixes it
```suggestion
          go get gotest.tools/gotestsum
          GOOS=linux TEST_DIRECTORY=./...  gotestsum --format pkgname -- -race -coverpkg=./... -coverprofile=cover.out.tmp ./...
```

orouz: TIL `go get` is for modules, `go install` for CLIs

oren-zohar (approved):
:rocket::rocket::rocket::rocket:

PR #1274, opened 2023-08-28
ECS common data mapping

amirbenun:
### Summary of your changes
Change `GetElasticCommonData()` to return a map.
Add enrich that accepts a provider with that function and attaches the fields to an event.

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/1284

### Checklist


orouz (approved):
LGTM

amirbenun:
Checking why so many tests are failing before I merge

oren-zohar:
@amirbenun Can you elaborate a bit on the motivation behind changing the signature to return a map?

amirbenun:
Sure @oren-zohar!
1. By changing the method to return a map I'm fixing an old hack we have.
We used to take a field named `ECSField` and put all the data coming back from `GetElasticCommonData` on top of it. There is no real reason for this differentiation between the first key (`process` for example) and the second key (`pid` for example), so instead the map will have a key named `process.pid`.
Another reason for this change is that we couldn't return ECS data from 2 different ECS sections which of course we want. A KMS key should return `x509.not_after` and also `cloud.service.name = AWS KMS`.
2. I tested it locally and I saw findings coming back from K8s, I am checking their structure now to see if it has changed because of this commit.

Issue #869, opened 2023-04-18
Populate vulnerability.reference with a link to NVD rather then aquasec domain


amirbenun:
**Motivation**
`vulnerability.reference` field should point to a website with documentation about the vulnerability.
Today, cloudbeat takes the link from the trivy's field, and that link points to `aquasec` domain.
Instead, we want to replace the link with a link of the form `nvd.nist.gov/vuln/detail/VULNERABILITY_ID` as discussed [here](https://github.com/elastic/kibana/issues/150505#issuecomment-1495600543). 

**Definition of done**
What needs to be completed at the end of this task
- [ ] Compose a new link and populate `vulnerability.reference` before sending to ES.

**Related tasks/epics**
- https://github.com/elastic/kibana/issues/150505

kfirpeled:
Assigned to max as a good first issue on cloudbeat, moved team to Cloud Risk
cc: @oren-zohar

maxcold:
Hey @amirbenun and @kfirpeled , I created a PR to cloudbeat https://github.com/elastic/cloudbeat/pull/1303 to generate `vulnerability.reference` using the same logic as we have in Kibana to generate a link to vulnerability to NVD site. I plan to remove the logic from the frontent and use the `vulnerability.reference` directly. I want to bring up a couple of small concerns related to this change to validate that I should proceed with the change:
1. removing the logic from the frontend means that with the new version of Kibana and with the old data the link will be pointing to `aquasec`. In my opinion this is acceptable taking into account that the only reason I see for this change is to simplify the logic on the frontend.  Do you agree?
2. by having this logic in cloudbeat there is a very small risk that for some users it might be unintended breaking change if they for some reason relied on the `reference` being aquasec before and it changes to nvd (eg. for some automation, but tbh I doubt that we already have such users). An alternative would be to introduce a new field smth like `nvdReference` or smth, but I feel that it would be overcomplication
3. while working on the feature I had a question why do we replace aquasec links with nvd ones in the first place? having this logic in my opinion just adds complexity as we can't always guarantee that the vulnerability is in NVD database and relying on a bit inderect logic of checking "if CVSS contains nvd we guess that it's safe to compose an nvd link out of vuln id". Not a big deal or smth, just wondering why having aquasec links are not suitable for us.

maxcold:
[...]

PR #1303, opened 2023-08-31
Populate vulnerability.reference with a link to NVD

maxcold:
### Summary of your changes

This PR moves the logic existing in Kibana to Cloudbeat of populating the NVD link instead of the Aquasec link in `vulnerability.reference` field.

### Screenshot/Data

the field should look like the following after the change
```
{
...
"reference": "https://nvd.nist.gov/vuln/detail/CVE-2022-3080",
...
}
```

### Related Issues

- Fixes: https://github.com/elastic/cloudbeat/issues/869#top

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ terraform.rc
 # Vulnerability management
 db/
 fanal/
+java-db/
--- a/vulnerability/events_creator.go
+++ b/vulnerability/events_creator.go
@@ -322,7 +322,7 @@ func (e EventsCreator) generateEvent(reportResult trivyTypes.Result, vul trivyTy
 				ID:             vul.VulnerabilityID,
 				Title:          vul.Title,
 				Enumeration:    getIdentifierType(vul.VulnerabilityID),
-				Reference:      vul.PrimaryURL,
+				Reference:      getReference(vul),
 				Description:    vul.Description,
 				Severity:       vul.Severity,
 				Classification: vulScoreSystemClass,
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -73,3 +73,4 @@ terraform.rc
 # Vulnerability management
 db/
 fanal/
+java-db/
```
maxcold, line 76: not sure if it's smth specific to my machine or not, but I had this untracked dir after running cloudbeat locally with vulnerabilities yaml

vulnerability/events_creator.go
```diff
@@ -391,6 +391,20 @@ func (e EventsCreator) getCVSSScore(vul trivyTypes.DetectedVulnerability) float6
...
 }
 
+const NVDVulnDetailBaseURL = "https://nvd.nist.gov/vuln/detail/"
+
+func getReference(vul trivyTypes.DetectedVulnerability) string {
+	if vul.CVSS == nil {
+		return vul.PrimaryURL
+	}
+
+	if _, ok := vul.CVSS[trivyVul.NVD]; ok {
```
maxcold, line 401: I'm assuming that if there is a NVD in CVSS then this vulnerability exists in NVD database and the link will be valid. It's a bit brittle in my opinion, but if we want to replace aquasec links that's the simplest way I could find

amirbenun: Also, as far as I understand this is the same solution we have today in kibana, right?

maxcold: yes, exactly!

amirbenun: nit: invert the `if`
```
if _, ok := vul.CVSS[trivyVul.NVD]; !ok {
  return vul.PrimaryURL
}
```

amirbenun (approved):

maxcold:
@amirbenun updated in case you want to take a look

Issue #1284, opened 2023-08-29
[Infra] ECS data providers


amirbenun:
### Motivation
ECS contain a lot of different common schemas: file, process, cloud, orchestrator.
We want to be able to fill them all before we send the event to ES.
The are different sorts of data from different levels of the benchmark:
1. Top level - Data is shared between all benchmarks and all cloud providers (for example: `cloudbeat.version`)
2. Benchmark - Data is shared between all events of the same benchmark (for example in K8s: `orchestrator.cluster.id`)
3. Cloud provider - Data is shared between all events of the same cloud account (for example: `cloud.account.id`)
4. Resource - Data is relevant for a specific resource (for example: `cloud.service.name`, `orchestrator.resource.id`)

### Definition of done
What needs to be completed at the end of this task
- [x] Introduce `ElasticCommonDataProvider` interface  - Completed in https://github.com/elastic/cloudbeat/pull/1274
- [x] Implement resource level data provider (change existing `GetElasticCommonData` signature) - Completed in https://github.com/elastic/cloudbeat/pull/1274
- [x] Implement `GetElasticCommonData` on `kms`, `network`, `rds`, `s3`, `iam` - Completed in https://github.com/elastic/cloudbeat/pull/1312
- [x] `CommonDataProvider` interface should not have `FetchData` and should populated by its `Options` - Completed in https://github.com/elastic/cloudbeat/pull/1305
- [ ] `EventsCreator` should accept a slice of different providers instead of one
- [ ] Implement different levels of data provider (aws, gcp, k8s, global)
- [x] Introduce a benchmark struct that depends on all the providers to create posture events - Completed in https://github.com/elastic/cloudbeat/pull/1346

### Related tasks/epics
- https://github.com/elastic/security-team/issues/3793
- https://github.com/elastic/security-team/issues/3791
- https://github.com/elastic/security-team/issues/3789

amirbenun:
Completing the rest of the action items requires a refactor that we avoid at the moment.
I will close the issue for now and reopen if we decide to accomplish it in the future.

PR #1305, opened 2023-09-03
Resource id provider

amirbenun:
### Summary of your changes
Separate `CommonDataProvider.FetchData` into two separate interfaces:
1. Id provider - in K8s the resource id should be calculated together with the node id.
2. Common provider - sets versioning data on the event.

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/1284

### Checklist


dataprovider/data_provider.go
```diff
@@ -20,11 +20,13 @@ package dataprovider
 import (
 	"github.com/elastic/beats/v7/libbeat/beat"
 
-	"github.com/elastic/cloudbeat/dataprovider/types"
 	"github.com/elastic/cloudbeat/resources/fetching"
 )
 
 type CommonDataProvider interface {
-	FetchData(resource string, id string) (types.Data, error)
```
uri-weisman, line 28: why not add `GetId` under the `CommonDataProvider` interface?

transformer/events_creator.go
```diff
@@ -92,16 +93,20 @@ func (t *Transformer) CreateBeatEvents(_ context.Context, eventData evaluator.Ev
...
 			},
 		}
 
-		err := t.commonDataProvider.EnrichEvent(&event, resMetadata)
+		err := t.benchmarkDataProvider.EnrichEvent(&event, resMetadata)
+		if err != nil {
+			return nil, fmt.Errorf("failed to enrich event with benchmark context: %v", err)
+		}
+
+		err = dataprovider.NewEnricher(t.commonDataProvider).EnrichEvent(&event)
```
uri-weisman, line 109: Let's create the enrichers before the loop scope.

dataprovider/data_provider.go
amirbenun: tl;dr: Single responsibility.
This interface is about to remove completely and we are going to have the new version:
```
type ElasticCommonDataProvider interface {
	GetElasticCommonData() (map[string]interface{}, error)
}
```
The new interface will have different implementations which will add fields in different layer of the resource. (resource specific, cloud specific, benchmark specific, global).
So while I'm in the process of removing that interface I don't want to add more functionality to it.

uri-weisman (approved):

PR #1310, opened 2023-09-03
Implement Azure fetchers and asset provider

jeniawhite:
### Summary of your changes
Implementation of assets provider using ARG and Azure fetcher in order to fetch resources for evaluation.

### Related Issues
- Fixes: #1250 

### Checklist


resources/providers/azurelib/inventory/provider.go
```diff
@@ -0,0 +1,141 @@
...
+	var resourceAssets []interface{}
+
+	query := armresourcegraph.QueryRequest{
+		Query: to.Ptr(generateQuery(assets)),
+		Options: &armresourcegraph.QueryRequestOptions{
+			ResultFormat: to.Ptr(armresourcegraph.ResultFormatObjectArray),
+		},
+		Subscriptions: []*string{
+			// TODO: Populate from config or query (not sensitive but still don't want to commit)
+			to.Ptr(os.Getenv("AZURE_SUBSCRIPTION_ID"))},
```
orestisfl, line 89: As discussed, this will be handled in a follow-up ticket

resources/fetching/fetcher.go
```diff
@@ -69,6 +69,14 @@ const (
...
 	ProjectManagement      = "project-management"
 	DataProcessing         = "data-processing"
+
+	// Azure resource types
+	AzureVMType               = "azure-vm"
+	AzureStorageAccountType   = "azure-storage-account"
+	AzureNetworkInterface     = "azure-network-interface"
+	AzureApplicationService   = "azure-application-service"
+	AzureLoggingAndMonitoring = "azure-logging-and-monitoring"
+	AzureDatabaseService      = "azure-database-service"
```
orestisfl, line 79:
```suggestion
```

Let's add these one by one as we implement the rules. We can merge this PR without them.

resources/fetching/fetchers/azure/assets_fetcher.go
```diff
@@ -0,0 +1,139 @@
...
+	return &AzureAssetsFetcher{
+		log:        log,
+		resourceCh: ch,
+		provider:   provider,
+	}
+}
+
+func (f *AzureAssetsFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
+	f.log.Info("Starting AzureAssetsFetcher.Fetch")
+	// TODO: Maybe we should use a query per type instead of listing all assets in a single query
```
orestisfl, line 69:
```suggestion
```

Is there any benefit to that?

resources/fetching/fetchers/azure/assets_fetcher.go
```diff
@@ -0,0 +1,139 @@
...
+	Id             string         `json:"id,omitempty"`
+	Name           string         `json:"name,omitempty"`
+	Location       string         `json:"location,omitempty"`
+	Properties     map[string]any `json:"properties,omitempty"`
+	ResourceGroup  string         `json:"resource_group,omitempty"`
+	SubscriptionId string         `json:"subscription_id,omitempty"`
+	TenantId       string         `json:"tenant_id,omitempty"`
+}
+
+// TODO: Implement other types
```
orestisfl, line 53:
```suggestion
```

Can we track this in a new issue?

orestisfl (changes requested):
I think we can merge this after some tests and resolving the TODOs

resources/fetching/fetchers/azure/assets_fetcher.go
jeniawhite: I assume that it might help us If we will need to collect multiple resource types to evaluate a single rule.

resources/fetching/fetchers/azure/assets_fetcher.go
jeniawhite: This is basically the other piece of work regarding the comment above:
```Let's add these one by one as we implement the rules. We can merge this PR without them```

File: resources/fetching/fetchers/azure/assets_fetcher.go [hunk 5]
```diff
@@ -0,0 +1,139 @@
...
+	provider   inventory.ServiceAPI
+}
+
+type AzureAsset struct {
+	Type    string
+	SubType string
+	Asset   AzureAssetInfo `json:"asset,omitempty"`
+}
+
+// TODO: Fill this struct with the required fields
```
orestisfl, line 42: Isn't this done?

```suggestion
```

resources/fetching/fetchers/azure/assets_fetcher.go
```diff
@@ -0,0 +1,139 @@
...
+
+	for _, asset := range assets {
+		select {
+		case <-ctx.Done():
+			f.log.Infof("AzureAssetsFetcher.Fetch context err: %s", ctx.Err().Error())
+			return nil
+		case f.resourceCh <- fetching.ResourceInfo{
+			CycleMetadata: cMetadata,
+			// TODO: Safe guard this conversion
+			Resource: getAssetFromData(asset.(map[string]any)),
```
orestisfl, line 83: Shouldn't `ListAllAssetTypesByName` return the correct type / handle the casting instead of the caller?

resources/fetching/fetchers/azure/assets_fetcher.go
```diff
@@ -0,0 +1,139 @@
...
+		},
+	}
+}
+
+func getString(data map[string]any, key string) string {
+	value, _ := data[key].(string)
+	return value
+}
+
+// TODO: Handle this function
```
orestisfl, line 115: Let's resolve the TODOs, is this relevant?

resources/fetching/fetchers/azure/assets_fetcher.go
```diff
@@ -0,0 +1,139 @@
...
+		Type:    r.Type,
+		SubType: r.SubType,
+		Name:    r.Asset.Name,
+		Region:  r.Asset.Location,
+	}, nil
+}
+
+func (r *AzureAsset) GetElasticCommonData() (map[string]any, error) { return nil, nil }
+
+// TODO: Implement this function
```
orestisfl, line 136: What's the significance of this? Is it a blocker?

resources/providers/azurelib/inventory/provider.go
```diff
@@ -0,0 +1,141 @@
...
+			}
+		} else {
+			break
+		}
+	}
+
+	return resourceAssets, nil
+}
+
+// TODO: Handle this function
```
orestisfl, line 138: Same

orestisfl (commented):
Let's try to merge this relatively clean in `main` unless if it's blocking more important work

I left comments on TODOs and we should definitely guard the casting or avoid casting alltogether

resources/fetching/fetchers/azure/assets_fetcher.go
jeniawhite: It is not a blocker because we did not decide on the subTypes yet.
As of current we only utilize the resource type and do not look at subTypes in the policies.
Future implementation of additional rules might change that, this is why it is left blank as of current.

resources/fetching/fetchers/azure/assets_fetcher_test.go
```diff
@@ -0,0 +1,121 @@
...
+		s.Equal(mockAssets[index].Properties, resource.Properties)
+		s.Equal(mockAssets[index].ResourceGroup, resource.ResourceGroup)
+		s.Equal(mockAssets[index].SubscriptionId, resource.SubscriptionId)
+		s.Equal(mockAssets[index].TenantId, resource.TenantId)
+		s.Equal(mockAssets[index].Type, resource.Type)
+		meta, err := r.GetMetadata()
+		s.NoError(err)
+		s.NotNil(meta)
+		s.NoError(err)
+		s.NotEmpty(meta)
```
orestisfl, line 114:
```suggestion
		require.NoError(s.T(), err)
		s.NotEmpty(meta)
```

resources/fetching/fetchers/azure/assets_fetcher_test.go
```diff
@@ -0,0 +1,121 @@
...
+		s.Equal(mockAssets[index].Id, meta.ID)
+		s.Equal(AzureResourceTypes[mockAssets[index].Type], meta.Type)
+		s.Equal("", meta.SubType)
+		s.Equal(mockAssets[index].Name, meta.Name)
+		s.Equal(mockAssets[index].Location, meta.Region)
```
orestisfl, line 119:
```suggestion
		s.Equal(mockAssets[index].Id, meta.ID)
		s.Equal(AzureResourceTypes[mockAssets[index].Type], meta.Type)
		s.Equal("", meta.SubType)
		s.Equal(mockAssets[index].Name, meta.Name)
		s.Equal(mockAssets[index].Location, meta.Region)
```

I would construct an expected `meta` object and compare it using the `Equal` function, this produces a better diff and it is more feature-proof as we can catch cases where we forget to use new field if we combine it with `exhauststruct`.

orestisfl (approved):
LGTM but the coverails still fails

Issue #1214, opened 2023-08-10
[AWS Orgs] New member accounts need to be detected by already-running cloudbeat


orestisfl:
**Motivation**
When using CloudFormation stack sets, new member accounts in an organization get the necessary roles for cloudbeat to use them. However, since AWS accounts are initialized once in the lifetime of fetchers, new accounts are not detected by an already-running cloudbeat: https://github.com/elastic/cloudbeat/blob/14b4ffd44e12322e3f7aae9cff06ce8de68c65c8/flavors/benchmark/aws_org.go#L61

**Definition of done**
What needs to be completed at the end of this task
- [ ] Accounts need to be re-scanned on each fetcher cycle

orestisfl:
@tinnytintin10 To be included in 8.10 documentation

kfirpeled:
- verify new account is being detected

CohenIdo:
8.11 BC4 QA cycle - verified: 

Before Adding new account:

![image](https://github.com/elastic/cloudbeat/assets/90558359/8c220007-3fe9-402a-981f-83ee45df5cb4)


After adding "Sandbox2" 
<img width="2552" alt="image" src="https://github.com/elastic/cloudbeat/assets/90558359/5810d2db-da63-4583-904f-f2e8960d31d1">

PR #1313, opened 2023-09-04
[AWS Orgs] Update fetchers in each cycle to detect changes in accounts

orestisfl:
### Summary of your changes

This adds a new `Update()` method in the registry interface and an updater in the registry which allows the caller to update the fetchers in each cycle.

### Screenshot/Data

Tests from running cloudbeat locally (without an agent), adding and deleting an account without restarting cloudbeat.

**Account that was created at ~16:05:**
screenshot-2023-09-04T17:07:05 CEST
**Account that was deleted at ~16:15:**
screenshot-2023-09-04T17:07:15 CEST

The accounts also show up correctly in the dashboard, the one that is deleted shows up as "Last evaluated an hour ago"
image

### Related Issues

- Fixes: #1214 

### Checklist


resources/fetching/registry/dynamic.go
```diff
@@ -0,0 +1,107 @@
...
+	d.lock.RLock()
+	defer d.lock.RUnlock()
+	return d.registry.Run(ctx, key, metadata)
+}
+
+func (d *dynamic) Stop() {
+	d.lock.RLock()
+	defer d.lock.RUnlock()
+
+	d.registry.Stop()
```
amirbenun, line 82: `dynamic.Stop` should also stop the scheduling from reoccurring

amirbenun (commented):
Nice!

flavors/benchmark/aws_org.go
```diff
@@ -58,18 +58,29 @@ func (a *AWSOrg) Initialize(ctx context.Context, log *logp.Logger, cfg *config.C
...
-	return registry.NewRegistry(
-			log,
-			factory.NewCisAwsOrganizationFactory(ctx, log, ch, accounts),
-		), cloud.NewDataProvider(
-			cloud.WithLogger(log),
-			cloud.WithAccount(*awsIdentity),
-		), cloud.NewIdProvider(), nil
+	cache := make(map[string]factory.FetchersMap)
+	reg := registry.NewRegistry(log, registry.WithUpdater(
+		func() (factory.FetchersMap, error) {
```
amirbenun, line 63: Nit: extract this function

resources/fetching/registry/registry_test.go
```diff
@@ -256,3 +260,97 @@ func (c *boolFetcherCondition) Condition() bool {
 func (c *boolFetcherCondition) Name() string {
 	return c.name
 }
+
+func Test_registry_Update(t *testing.T) {
```
amirbenun, line 264: Can you add another test where you initial the registry with no `Option`s?
The empty use case should probably just returns an error or hold an empty map that cannot be changed.
I want to make sure it doesn't panic.

amirbenun (approved):

resources/fetching/registry/registry_test.go
orestisfl: That would be pretty much equivalent to the `check nil` case

PR #1330, opened 2023-09-08
go.mod: Organize require blocks and update SDKs

orestisfl:
### Summary of your changes

Organizes go.mod in two `require` blocks: one with direct and one with indirect dependencies. Additionally, bump some of our SDKs in the latest version


oren-zohar (approved):
Nice! I would add a comment above each one (even though it's pretty clear)

go.mod
orouz, line 1: is this grouping done automatically?

orestisfl: It should, but let's see if the ordering is maintained

PR #1332, opened 2023-09-10
Use branches to get bundles

orouz:
- use `defaultBeatVersion`  instead of `PolicyVersion`
 - `BuildOpaBundle` now fetches a *branch* in `elastic/csp-security-policies`, not a tag. 
   - tries to get  major.minor of`defaultBeatVersion`, fallsback to main
   - ensures latest always fallsback to main
 - release manager will open new version branch in `elastic/csp-security-policies` 

---

this needs to be backported to cloudbeat's `8.9.0` and `8.10.0` to make it use branches and not tags. 
[CSP 8.9](https://github.com/elastic/csp-security-policies/tree/8.9)
[CSP 8.10](https://github.com/elastic/csp-security-policies/tree/8.10)

--- 

closes https://github.com/elastic/security-team/issues/5451


version/policy.go
```diff
@@ -17,7 +17,7 @@
 
 package version
 
-const policyVersion = "v1.5.5"
+const policyVersion = "8.10"
```
orouz, line 20: if cloudbeat branches are always aligned with csp-security-policies branches, as in `8.10` branch in cloudbeat will fetch the `8.10` branch from csp-policies to build the bundle, then we can drop this `policyVersion` variable and always use `CloudbeatVersion`

oren-zohar (approved):

PR #1337, opened 2023-09-11
Remove extra build step

orouz:
looks like `Build` already includes a call to `BuildOpaBundle`:
https://github.com/elastic/cloudbeat/blob/48e17af899402c9121ce177379294053e396f482/magefile.go#L81-L84

### DIFF
--- a/.github/workflows/cloudbeat-ci.yml
+++ b/.github/workflows/cloudbeat-ci.yml
@@ -82,12 +82,6 @@ jobs:
           version: latest
           args: build
 
-      - name: build opa bundle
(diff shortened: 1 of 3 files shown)


orestisfl (approved):

orouz:
the bundle is removed after the function is done, looks like it's just there to ensure the build successeds. maybe we can remove the dep from `Build()`? 

https://github.com/elastic/cloudbeat/blob/763e7038b047e1c56567864ce6f36569b4745d4b/magefile.go#L353-L359

@orestisfl wdyt?

PR #1346, opened 2023-09-18
Benchmark interface to create posture events

amirbenun:
### Summary of your changes
Implement parts of the posture structure under a new component named benchmark that will get manager, evaluator, and transformer and return a channel of events.

### Related Issues
- https://github.com/elastic/cloudbeat/issues/1284

### Checklist


flavors/benchmark/aws.go
```diff
@@ -28,16 +28,31 @@ import (
...
 	"github.com/elastic/cloudbeat/dataprovider"
 	"github.com/elastic/cloudbeat/dataprovider/providers/cloud"
+	"github.com/elastic/cloudbeat/flavors/benchmark/builder"
 	"github.com/elastic/cloudbeat/resources/fetching"
 	"github.com/elastic/cloudbeat/resources/fetching/preset"
 	"github.com/elastic/cloudbeat/resources/fetching/registry"
 	"github.com/elastic/cloudbeat/resources/providers/awslib"
 )
 
+const resourceChBuffer = 10000
```
orestisfl, line 38:
```suggestion
const resourceChBufferSize = 10000
```

flavors/benchmark/strategy.go
```diff
@@ -24,24 +24,19 @@ import (
...
-	Initialize(ctx context.Context, log *logp.Logger, cfg *config.Config, ch chan fetching.ResourceInfo) (registry.Registry, dataprovider.CommonDataProvider, dataprovider.IdProvider, error)
-	Stop()
-
+type strategy interface {
+	NewBenchmark(ctx context.Context, log *logp.Logger, cfg *config.Config) (builder.Benchmark, error)
 	checkDependencies() error
 }
 
-func NewBenchmark(cfg *config.Config) (Benchmark, error) {
+func GetStrategy(cfg *config.Config) (strategy, error) {
```
orestisfl, line 39:
```suggestion
func GetStrategy(cfg *config.Config) (Strategy, error) {
```

flavors/benchmark/aws.go
```diff
@@ -28,16 +28,31 @@ import (
...
 )
 
+const resourceChBufferSize = 10000
+
 type AWS struct {
 	IdentityProvider awslib.IdentityProviderGetter
 }
 
+func (a *AWS) NewBenchmark(ctx context.Context, log *logp.Logger, cfg *config.Config) (builder.Benchmark, error) {
+	resourceCh := make(chan fetching.ResourceInfo, resourceChBufferSize)
```
uri-weisman, line 45: wondering whether we should get the `resourceCh` as an input rather than creating it for every benchmark.
wdyt?

flavors/benchmark/builder/builder.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	evaluator, err := evaluator.NewOpaEvaluator(ctx, log, cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	cdp := common.New(log, version.CloudbeatVersionInfo{
```
uri-weisman, line 95: `log` is unused, and can be removed from the `New` method.

flavors/benchmark/builder/builder.go
```diff
@@ -0,0 +1,109 @@
...
+	idp              dataprovider.IdProvider
+	bdp              dataprovider.CommonDataProvider
+	k8sLeaderElector uniqueness.Manager
+}
+
+func New(options ...Option) *Builder {
+	b := &Builder{
+		managerTimeout:   defaultManagerTimeout,
+		idp:              &idProvider{},
+		bdp:              &dataProvider{},
```
uri-weisman, line 59: It's a bit confusing that those struct implementations reside under builder.
consider shifting the "base" implementation to be under the relevant utility.

flavors/publisher.go
```diff
@@ -45,7 +45,7 @@ func NewPublisher(log *logp.Logger, interval time.Duration, threshold int, clien
 	}
 }
 
-func (p *Publisher) HandleEvents(ctx context.Context, ch chan []beat.Event) {
+func (p *Publisher) HandleEvents(ctx context.Context, ch <-chan []beat.Event) {
```
uri-weisman, line 48: 👑

uri-weisman (commented):
Great effort @amirbenun !

flavors/benchmark/builder/builder.go
amirbenun: Since this PR is already big enough I will complete that part on another PR.

flavors/benchmark/builder/builder.go
amirbenun: I agree that it may not should be under the builder package however I think those implementations should remain private. Maybe changing the whole package name?

flavors/benchmark/k8s.go
```diff
@@ -40,6 +41,20 @@ type K8S struct {
...
+	resourceCh := make(chan fetching.ResourceInfo, resourceChBuffer)
+	reg, bdp, idp, err := k.Initialize(ctx, log, cfg, resourceCh)
+	if err != nil {
+		return nil, err
+	}
+
+	return builder.New(
+		builder.WithBenchmarkDataProvider(bdp),
+		builder.WithIdProvider(idp),
+		builder.WithK8sLeaderElector(k.leaderElector),
```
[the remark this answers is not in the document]
uri-weisman: will be handled in a different PR.

uri-weisman (approved):

PR #1348, opened 2023-09-18
Update Cloud Env workflow to default to our Org API key

oren-zohar:
### Summary of your changes
Updating our [cloud env testing](https://github.com/elastic/cloudbeat/blob/main/dev-docs/Cloud-Env-Testing.md) to use our cloud organization API key by default.

### Checklist

### DIFF
--- a/.github/workflows/destroy-environment.yml
+++ b/.github/workflows/destroy-environment.yml
@@ -5,10 +5,6 @@ on:
   # Ability to execute on demand
   workflow_dispatch:
     inputs:
-      ec-api-key:
-        type: string
-        description: "Elastic Cloud API key"
-        required: true
       prefix:
         type: string
         description: "Delete all environments starting with `prefix`"
@@ -16,6 +12,10 @@ on:
       ignore-prefix:
         type: string
         description: "Ignore all environments starting with `ignore-prefix`"
+      ec-api-key:
+        type: string
+        description: "**Optional** To delete env environments on your own organization, enter your Elastic Cloud API key."
+        required: true
 
 env:
   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/.github/workflows/destroy-environment.yml
+++ b/.github/workflows/destroy-environment.yml
@@ -15,7 +15,7 @@ on:
       ec-api-key:
         type: string
         description: "**Optional** To delete env environments on your own organization, enter your Elastic Cloud API key."
-        required: true
+        required: false
 
 env:
   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
@@ -23,6 +23,7 @@ env:
   AWS_REGION: "eu-west-1"
   ENV_PREFIX: ${{ github.event.inputs.prefix }}
   ENV_IGNORE_PREFIX: ${{ github.event.inputs.ignore-prefix }}
+  TF_VAR_ec_api_key: ${{ secrets.EC_API_KEY }}
 
 jobs:
   Destroy:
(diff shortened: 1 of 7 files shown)


.github/workflows/destroy-environment.yml
```diff
@@ -5,17 +5,17 @@ on:
...
         type: string
         description: "Delete all environments starting with `prefix`"
         required: true
       ignore-prefix:
         type: string
         description: "Ignore all environments starting with `ignore-prefix`"
+      ec-api-key:
+        type: string
+        description: "**Optional** To delete env environments on your own organization, enter your Elastic Cloud API key."
+        required: true
```
gurevichdmitry, line 18: I want to clarify the definition of `ec-api-key` as optional. Despite its `required: true` setting, users must manually input values for the required inputs before the workflow can run, as specified in the [GitHub docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs). If we intend to treat it as optional, it should be interpreted as `required: false`. In that case, we would need to develop logic within an action (job step) to use the default API key defined as a secret when no API key is provided.

.github/workflows/test-environment.yml
```diff
@@ -33,6 +29,10 @@ on:
         description: "Cleanup resources after provision"
         default: false
         type: boolean
+      ec-api-key:
+        type: string
+        description: "**Optional** By default, the environment will be created in our Cloud Security Organization. If you want to use your own cloud account, enter your Elastic Cloud API key."
+        required: true
```
gurevichdmitry, line 35: The same applies in this case. Despite its required: true setting, users must manually input values for the required inputs before the workflow can run, as specified in the [GitHub docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs). If we intend to treat it as optional, it should be interpreted as required: false. In that case, we would need to develop logic within an action (job step) to use the default API key defined as a secret when no API key is provided.

dev-docs/Cloud-Env-Testing.md
```diff
@@ -16,29 +16,25 @@ Follow these steps to run the workflow:
...
+      version. Check the available versions [here](https://artifacts-staging.elastic.co/dra-info/index.html).
+      For BC, enter only the version without additions/commit sha, e.g. `8.8.1`.
+      For SNAPSHOT, enter the full version, e.g. `8.8.1-SNAPSHOT`.
 
     - **`ess-region`**: Indicate the Elastic Cloud deployment region. The default value is `gcp-us-west2`, which
       supports
       snapshot and build candidate (BC) versions. Specify a different region only if necessary.
 
-   ![Enter Inputs](https://github.com/elastic/cloudbeat/assets/99176494/06d8144d-13cc-4e13-92fc-19f52ce8206b)
+   ![Required Parameters](https://github.com/oren-zohar/cloudbeat/assets/85433724/6159129e-6d4d-46b1-97a1-f0d3859500fd)
```
gurevichdmitry, line 33: nit: We may need to upload images to the https://github.com/elastic/cloudbeat repo

dev-docs/Cloud-Env-Testing.md
```diff
@@ -47,19 +43,24 @@ Follow these steps to run the workflow:
...
 
-      .
-   ![Adjust Inputs](https://github.com/elastic/cloudbeat/assets/99176494/bac5004d-7cbc-4a34-8127-3acd11acc90e)
+    - **`ec-api-key`** (**optional**): By default, all the new environments will be created in our EC Cloud Security organization.
+      If you want to create the environment on your personal org (`@elastic.co`) you can enter
+      your private [Elastic Cloud](https://cloud.elastic.co/home) API key. Follow the
+      [Cloud API Keys](https://www.elastic.co/guide/en/cloud/current/ec-api-authentication.html) documentation for
+      step-by-step instructions on generating the token.
 
+   ![Optional Parameters](https://github.com/oren-zohar/cloudbeat/assets/85433724/17933589-ee0e-4181-a244-f501f54bda6c)
```
gurevichdmitry, line 58: nit: We may need to upload images to the https://github.com/elastic/cloudbeat repo

gurevichdmitry (approved):
@oren-zohar Great improvement. You may have an example of a run without an API key provided (default API key usage) and also an additional run with an API key provided to ensure that the logic works correctly?

PR #1349, opened 2023-09-18
update version to 8.10.2

CohenIdo:
Update version to 8.10.2 after 8.10.1 was released.


oren-zohar (approved):
Thank you @CohenIdo :heart:

PR #1354, opened 2023-09-20
pre-commit: Add shfmt hook

orestisfl:
### Summary of your changes
Helps us have a more consistent code style in our shell scripts, uses popular https://github.com/mvdan/sh linter.


oren-zohar:
@orestisfl can we use the [`--indent`](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#printer-flags) option to maintain the current indentation? Also, have you tried using [`--simplify`](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#generic-flags) and [`--minify`](https://github.com/mvdan/sh/blob/master/cmd/shfmt/shfmt.1.scd#generic-flags)?

orestisfl:
@oren-zohar
- Indented with 4 spaces instead
- `--minify` seems to result in less readable results in favor of reducing size
- `--simplify` might be beneficial but it changes some things I am not 100% comfortable with. The diff is not that large anyway

oren-zohar (approved):

Issue #1248, opened 2023-08-17
[CIS Azure] Implement deployment flow


jeniawhite:
**Motivation**
We would like to deploy the Azure scanner in a similar manner to our AWS solution.
There are several potential options that are available on Azure for deployments:
- [Azure Powershell](https://learn.microsoft.com/en-us/powershell/azure/what-is-azure-powershell?view=azps-10.2.0)
- [Azure CLI](https://learn.microsoft.com/en-us/cli/azure/what-is-azure-cli)
- [Azure ARM templates](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/overview)

The priority is to have a UI for the user instead of running scripts and writing commands into terminals.
This is why, I've investigated the Azure ARM templates as a potential solution.

There is an option to deploy Azure ARM templates with a button but this only supports the template without propagation of parameter values via the URI.
Syncing with @tinnytintin10 we've agreed that the user will just copy the parameters by hand (for the first drop at least).

I took the time to implement an example template:
<details>
  <summary>template</summary>

```json
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "resources": [
        {
            "type": "Microsoft.Compute/virtualMachines",
            "apiVersion": "2021-04-01",
            "name": "evgbVM",
            "location": "[resourceGroup().location]",
            "dependsOn": [
                "[resourceId('Microsoft.Network/networkInterfaces', 'evgbNic')]"
            ],
            "properties": {
                "hardwareProfile": {
                    "vmSize": "Standard_DS2_v2"
                },
                "storageProfile": {
                    "osDisk": {
                        "createOption": "FromImage",
                        "managedDisk": {
                            "storageAccountType": "Standard_LRS"
                        }
                    },
                    "imageReference": {
                        "publisher": "Canonical",
                        "offer": "UbuntuServer",
                        "sku": "18.04-LTS",
                        "version": "latest"
                    }
                },
                "osProfile": {
                    "computerName": "evgbVM",
                    "adminUsername": "jeniauser",
                    "adminPassword": "evgbJen1@"
                },
                "networkProfile": {
                    "networkInterfaces": [
                        {
                            "id": "[resourceId('Microsoft.Network/networkInterfaces', 'evgbNic')]"
                        }
                    ]
                }
            },
            "identity": {
                "type": "SystemAssigned"
            }
        },
        {
[...]

PR #1357, opened 2023-09-20
Azure: Deploy Elastic Agent

orestisfl:
### Summary of your changes
This adds a script to deploy the Elastic Agent. When the script fails, Azure displays stdout and stderr directly on the deployment page. The script is hosted directly in our github repo.

### Screenshot/Data
image

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/1248

### Checklist


.pre-commit-config.yaml
```diff
@@ -75,6 +75,8 @@ repos:
     hooks:
       - id: shfmt
         args: ["-i", "4"]
+      - id: shellcheck
+        files: deploy/azure/.*sh
```
orestisfl, line 79: We can add this to other scripts but most of them fail so I've added it just for this script as an opportunity to use it in new scripts. We can track improvements on the rest on a new ticket.

This is IMO a good idea, especially for public-facing scripts that are run by users.

deploy/azure/install-agent.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -euxo pipefail
+
+usage="$0 <elastic agent version> <elastic artifact server> <fleet url> <enrollment token>"
+ElasticAgentVersion=${1:?$usage}
+ElasticArtifactServer=${2:?$usage}
+FleetUrl=${3:?$usage}
+EnrollmentToken=${4:?$usage}
+
+ElasticAgentArtifact="elastic-agent-$ElasticAgentVersion-linux-x86_64"
```
orestisfl, line 10: I went with x86 here, should we use arm for azure as well for consistency?

deploy/azure/install-agent.sh
```diff
@@ -0,0 +1,14 @@
...
+curl -L -O "${ElasticArtifactServer}/$ElasticAgentArtifact.tar.gz"
+tar xzf "${ElasticAgentArtifact}.tar.gz"
+cd "${ElasticAgentArtifact}"
+sudo ./elastic-agent install --non-interactive --url="${FleetUrl}" --enrollment-token="${EnrollmentToken}"
```
orestisfl, line 14: This is missing the `--tag` arguments but I am not sure if we have anything to add here yet.

deploy/azure/install-agent.sh
oren-zohar: Most of our test if not all of them are on Linux x86 machines AFAIK, so I think it's better to have it like that

oren-zohar (approved):

PR #1360, opened 2023-09-21
Azure ARM: Fix Role Assignment race condition

orestisfl:
### Summary of your changes
This fixes an error with Azure:
> PrincipalNotFound
> Principal {principalId} does not exist in the directory {tenantId}.
> Check that you have the correct principal ID. If you are creating this
> principal and then immediately assigning a role, this error might be
> related to a replication delay. In this case, set the role assignment
> principalType property to a value, such as ServicePrincipal, User, or
> Group.  See https://aka.ms/docs-principaltype
https://learn.microsoft.com/en-us/azure/role-based-access-control/troubleshooting?tabs=armtemplate#symptom---unable-to-assign-a-role-using-a-service-principal-with-azure-cli

Here, I am following the recommended solution by Azure:
> If you're creating a new user or service principal using the REST API
> or ARM template, set the principalType property when creating the role
> assignment using the Role Assignments - Create API.

Additionally, I am re-naming the role assignment to be based on the resource group and deployment name. This means that there will be one unique role assignment name per deployment and no conflicts should be caused at the subscription level. The previous name, `guid(uniqueString('cloudbeatRoleAssignment'))` deterministically returns the same name every time for these reasons:
1. `guid()` is deterministic:
> The returned value isn't a random string, but rather the result of a
> hash function on the parameters. The returned value is 36 characters
> long. It isn't globally unique.
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string#guid
2. `uniqueString()` is also just a hash function:
> The returned value isn't a random string, but rather the result of a
> hash function. The returned value is 13 characters long
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string#uniquestring

As an additional note, role assignments *need* to be GUIDs:
```
The role assignment ID
'cloudbeatRoleAssignment-4d2a7eb4-45f3-5712-b8f6-4a2d6307a2b8' is not
valid. The role assignment ID must be a GUID. (Code:
InvalidRoleAssignmentId)
```

### Screenshot/Data
image
screenshot-2023-09-21T11:01:21 CEST

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/1248

### Checklist


oren-zohar (approved):
LGTM. Thanks for the informative description

PR #1363, opened 2023-09-21
Azure ARM: Use temporary public key and disable SSH

orestisfl:
### Summary of your changes

This uses the script to kill SSH as an additional precaution.

Public IP to be removed in future change.

The new parameter is needed because the only place where Azure allows randomness is in the `defaultValue` of parameters:
> You can only use this function within an expression for the default
> value of a parameter. Using this function anywhere else in a template
> returns an error. The function isn't allowed in other parts of the
> template because it returns a different value each time it's called.
> Deploying the same template with the same parameters wouldn't reliably
> produce the same results.
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-string#newguid

### Related Issues
- Related: #1248

### Checklist


orestisfl:
I've pushed another commit that generate a public ssh key.

This removes the need for the seed parameter by using a
`Microsoft.Resources/deploymentScripts` resource to generate a temporary
public key. The private key is thrown out and ssh is still disabled by
the deployment script.

Downside of this approach is that the deployment script needs some time
to complete.

Resource output:
screenshot-2023-09-25T15:39:25 CEST
screenshot-2023-09-25T15:40:23 CEST
Resource tree:
screenshot-2023-09-25T15:41:43 CEST

deploy/azure/install-agent.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -euxo pipefail
 
+# Delete password, stop and disable ssh
+sudo passwd --delete cloudbeat &
+sudo systemctl disable --now ssh &
+sudo systemctl mask ssh.service &
+sudo killall sshd &
```
oren-zohar, line 8: What if one of the commands fails? 
Could running all commands in the background concurrently potentially lead to race conditions?

oren-zohar (changes requested):

deploy/azure/install-agent.sh
orestisfl: If they fail, we just have to accept we can't do much. Re-ordered them around to minimize the scenario where something fails.

oren-zohar (approved):

PR #1366, opened 2023-09-26
Azure: Dynamically get subscription ID(s)

orestisfl:
### Summary of your changes
Azure: Dynamically get subscription ID(s)

I've confirmed that with the ARM deployment method, this only retrieves one ID instead of the two I am able to see with my account. This is true both when the role assignment is in the resource group level and the subscription level.

### Checklist


resources/providers/azurelib/inventory/provider.go
```diff
@@ -64,28 +68,65 @@ type ProviderInitializerAPI interface {
...
+		log:           log,
+		client:        wrapper,
+		subscriptions: subscriptions,
+		ctx:           ctx,
+		Config:        azureConfig,
 	}, nil
 }
 
+func (p *ProviderInitializer) getSubscriptionIds(ctx context.Context, azureConfig auth.AzureFactoryConfig) ([]*string, error) {
+	// TODO: mockable
```
orestisfl, line 107: `Init()` is not tested so I consider it out of scope to test this function for now, we should handle this technical debt after the feature freeze though

jeniawhite (approved):

PR #1372, opened 2023-10-01
Support Azure activity logs

jeniawhite:
### Summary of your changes
When implementing rules, we've found out that some rules require all of the activity logs of the subscription to be evaluated as a batch of a single resource.

### Checklist

### DIFF
--- a/flavors/benchmark/azure_test.go
+++ b/flavors/benchmark/azure_test.go
@@ -63,6 +63,7 @@ func TestAzure_Initialize(t *testing.T) {
 			inventoryInitializer: mockAzureInventoryInitializerService(nil),
 			want: []string{
 				"azure_cloud_assets_fetcher",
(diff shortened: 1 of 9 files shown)


orestisfl (approved):
I don't agree with the approach of creating a new fetcher for this, I would have preferred to keep a single query for all resource types and then post-process them according to their type. For example, activity log resources should be batched according to their sub id while vms should just be sent as is, without batching. I believe that that approach will save us future engineering effort as we will need to do the same for other resource type and it also has the benefit of only making one API request instead of multiple.

However, I don't see anything wrong with the implementation so I am approving to unblock this.

jeniawhite:
@orestisfl 
Your approach was one of the options as well.
Due to the FF time constraints, I will merge this in order for us to advance.
But let's reopen this talk and revert if needed.

Issue #1326, opened 2023-09-07
CloudFormation version hardening


amirbenun:
### Background
The CloudFormation template is not backward compatible.
Meaning, when running a 8.10.x stack and deploying an agent using CloudFormation, you cannot modify the version to 8.9.x.
More on that in the [ticket](https://github.com/elastic/cloudbeat/issues/1208).

### Motivation
The current experience will let you deploy the stack but cloudbeat will produce errors and will fail to report vulnerabilities.
Instead we want to have a better UX that will block the user from creating that setup.
The most straight forward solution is to validate that the CloudFormation parameter is of the form 8.10.x  and fail fast otherwise with a readable description.

### Definition of done
- [ ] Validate the form of the CloudFormation parameter `ElasticAgentVersion`
- [ ] When fails to deploy, the CloudFormation error message should be clear and relevant

### Related tasks/epics
- https://github.com/elastic/cloudbeat/issues/1208

PR #1381, opened 2023-10-03
Elastic agent version should be >= cloud formation template version

moukoublen:
### Summary of your changes

This PR implements a (forward compatible) version hardening between the cloud formation template version and the configurable elastic agent  version. 

The check is implemented by extracting the elastic stack version from the cloud formation template version string and check that the requested elastic agent version is grater than or equal to that. 

### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/1326

### Checklist


deploy/cloudformation/elastic-agent-ec2-cnvm.yml
```diff
@@ -142,14 +142,25 @@ Resources:
...
           #!/bin/bash -x
+          signal() {
+            sudo yum install -y aws-cfn-bootstrap || { sudo apt-get update && sudo apt-get -y install python3-pip && sudo pip3 install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz; }
+            cfn-signal --exit-code "$1" --stack ${AWS::StackName} --resource ElasticAgentEc2Instance --region ${AWS::Region}
+          }
+          ## version_gte expects 2 version strings and returns code 0 if the first one is greater than or equal to the second.
+          version_gte() {
+            printf '%s\n%s\n' "$2" "$1" | sort --check=quiet --version-sort
+          }
+          AGENT_MIN_VERSION="$(echo 'CFT_VERSION' | sed --regexp-extended  's/^[a-zA-Z-]+-(.*)-[[:digit:]]{4}(-[[:digit:]]{2}){5}.yml/\1/g')"
```
orestisfl, line 153: (Optional) It might be more readable to do the version checking with an inline python script

moukoublen: Sure, I guess you mean to use something like the [semver](https://python-semver.readthedocs.io/en/latest/usage/compare-versions.html) package to write something like that `semver.compare("1.0.0", "2.0.0")` ? 

That would be more readable. 

Can we safely assume that the semver would be installed in the default vm? If not what could be the approach?

Is there any native way to do it in python? (I am not super familiar with python 😅)

orestisfl: semver is not part of the standard library, so I don't think we should rely on it.

Something like this could be a starting point:
```shell
python -c '
import sys
from packaging.version import Version
a = sys.argv[1]
b = sys.argv[2]
print(Version(a) < Version(b))
' 1.0 2.0
```

orestisfl: Nvm, `packaging` is also non-standard

deploy/cloudformation/elastic-agent-ec2-cnvm.yml
```diff
@@ -167,14 +167,25 @@ Resources:
...
+          signal() {
+            sudo yum install -y aws-cfn-bootstrap || { sudo apt-get update && sudo apt-get -y install python3-pip && sudo pip3 install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz; }
+            cfn-signal --exit-code "$1" --stack ${AWS::StackName} --resource ElasticAgentEc2Instance --region ${AWS::Region}
+          }
+          is_version_valid() {
+            local min_version="$(echo "$2" | sed --regexp-extended 's/^[a-zA-Z-]+-(.*)-[[:digit:]]{4}(-[[:digit:]]{2}){5}.yml/\1/g')"
+            [[ "$min_version" = CFT_* ]] && return 0 # if the file is not published do not check
+            printf '%s\n%s\n' "$min_version" "$1" | sort --version-sort --check=quiet
+          }
+          if ! is_version_valid "${ElasticAgentVersion}" 'CFT_VERSION' &> /dev/null; then
```
amirbenun, line 179: The `CFT_VERSION` placeholder is being replaced with something in the form of `cloudformation-cnvm-8.11.0.yml`, which I don't think is what you expect.
We have two options:
1. Add another parameter, `ELASTIC_VERSION` to be replaced with `8.11.0`.
2. Replace `CFT_VERSION` with `8.11.0` without the extra data that I see there and put that data hard coded on the template.

Hope the two options are clear, I would go with 2 but it's up to you.

amirbenun (changes requested):

deploy/cloudformation/elastic-agent-ec2-cnvm.yml
moukoublen: Thanks for the feedback. 
I went with the 2nd option, `CFT_VERSION` is being rendered having only the version string, and the remote (s3) file name is being created inside `scripts/publish_cft.sh` and also in cloudformation files while setting the tag:

e.g.
```
--tag=cft_version:cloudformation-cspm-organization-account-CFT_VERSION.yml
```

amirbenun (approved):

PR #1396, opened 2023-10-04
Azure: Populate cloud.account.name with DisplayName

### LINKED ISSUE #1395: [CIS Azure] Use human-friendly subscription name in ECS fields
**Motivation**
Subscription name is currently set to subscription ID. This is not user-friendly and it has been requested by the Kibana team to either be empty or actually be the subscription name as provided by azure.

**Related tasks/epics**
https://github.com/elastic/cloudbeat/issues/1390

orestisfl:
### Summary of your changes
Changes provider initialization to map subscription ids to names

### Screenshot/Data
image

### Related Issues
- Blocked by #1394 (based on branch of that PR)
- Fixes #1395

### Checklist


resources/providers/azurelib/inventory/provider.go
```diff
@@ -108,10 +110,10 @@ func (p *ProviderInitializer) Init(ctx context.Context, log *logp.Logger, azureC
 	}, nil
 }
 
-func (p *ProviderInitializer) getSubscriptionIds(ctx context.Context, azureConfig auth.AzureFactoryConfig) ([]*string, error) {
+func (p *ProviderInitializer) getSubscriptionIds(ctx context.Context, azureConfig auth.AzureFactoryConfig) (map[string]string, error) {
 	// TODO: mockable
```
oren-zohar, line 114: is this TODO still relevant?

resources/providers/azurelib/inventory/provider.go
```diff
@@ -212,3 +203,20 @@ func (p *Provider) runPaginatedQuery(query armresourcegraph.QueryRequest) ([]Azu
 
 	return resourceAssets, nil
 }
+
+func (p *Provider) getAssetFromData(data map[string]any) AzureAsset {
+	subId := getString(data, "subscriptionId")
+	properties, _ := data["properties"].(map[string]any)
```
oren-zohar, line 209: Should we have error handling if type assertions fail here?

resources/providers/azurelib/inventory/provider.go
```diff
@@ -124,8 +126,8 @@ func (p *ProviderInitializer) getSubscriptionIds(ctx context.Context, azureConfi
 			return nil, err
 		}
 		for _, subscription := range page.Value {
-			if subscription != nil {
-				result = append(result, subscription.SubscriptionID)
+			if subscription != nil && subscription.SubscriptionID != nil {
+				result[*subscription.SubscriptionID] = strings.Dereference(subscription.DisplayName)
```
oren-zohar, line 130: Can this lead to a panic?

resources/providers/azurelib/inventory/provider.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
 	"github.com/elastic/elastic-agent-libs/logp"
 	"github.com/samber/lo"
+	"golang.org/x/exp/maps"
```
oren-zohar, line 31: fyi [it was approved to be added to stdlib in Go 1.21](https://github.com/golang/go/issues/57436#issuecomment-1412566002)

resources/providers/azurelib/inventory/provider.go
orestisfl: yes :+1: we are missing tests for these paths (whole `Init()`)

resources/providers/azurelib/inventory/provider.go
orestisfl: No, I check the `subscription.SubscriptionID` for `nil`ness in the `if` above

resources/providers/azurelib/inventory/provider.go
orestisfl: Not much we can do if these are missing since that would either mean the specific asset doesn't have "properties" (I don't know if that's possible) or the output was malformed. Since the code was just moved around, I'd argue to keep the code as is for now and decide on what to do for these cases (also for `getString` which ignores missing values / wrong types).

The code is safe from panics.

jeniawhite (approved):

Issue #1108, opened 2023-07-03
Listen to ctx.Done() when sending to channels


amirbenun:
## Background
A bug that we found in CNVM but can be relevant all across:
- https://github.com/elastic/cloudbeat/issues/1097

## Describe the bug
The pattern of sending an event on a channel without knowing who or when it is going to be consumed.
Examples outside of the vulnerability flavor in `pipeline.go`:
```go
	go func() {
		defer close(outputCh)
		defer cancel()

		for s := range inputChannel {
			val, err := fn(ctx, s)
			if err != nil {
				log.Error(err)
				continue
			}
			outputCh <- val
		}
	}()
```
Since `outputCh` is consumed outside of this section, sending to `outputCh` is not safe.
If for example the consumer will stop listening to this channel (because the context canceled) it will result in deadlock.
An alternative is to replace `outputCh <- val` with 
```go
select {
    case <- ctx.Done():
        return
    case outputCh <- val:
}
```

## Definition of done
- Map the sections in our source code who are in risk
- Suggest a fix where possible
- Add unit tests that fail in the current state and pass after the fix

uri-weisman:
Thanks @amirbenun for the detailed explanation.
This bug only emphasizes the need for a shared "pipeline" utility across the project.

Issue #691, opened 2023-02-01
[Discuss] Replace pipeline package with RX


amirbenun:
**Motivation**
Today in cloudbeat, we have our own custom implementation for handling asynchronous flow.
`Step` is our thin generic function to support a multilayer of asynchronous pipelines implemented over go channels.
Instead, we can use a well-tested library that will allow us to go forward without the concern of missing functionality or bugs.
[RxGo](https://github.com/ReactiveX/RxGo) has 4.5K stars, and the organization is also popular among other coding languages.

**Definition of done**
- [ ] A POC to check if the library fits our requirements
- [ ] A summary of the pros and cons of using it
- [ ] Create relevant tasks based on the decision

jeniawhite:
I would like us to preserve the way we are treating the different steps of the vulnerability as a sort of micro-services that works on the channel queue in the background and are easily scalable.
If you look at the initial solution each step works on a single entity from a different channel and you can easily scale them with goroutines without doing any code changes increasing the concurrency to our needs.

This feels like an improvement that we can definitely prioritize and perform a quick POC to ensure viability, but I think that we have more important things to cover as of right now in the immediate time.
Once we will have a stable flow of the pipeline (including reclaiming of snapshots), then it will be probably easier to refactor the pipeline to our tailored needs and we will have more knowledge on the areas of focus for improvements.

oren-zohar:
our CNVM EKS implementation could be a good opportunity for this - we can see how stable it is, and then refactor our original implementation

orestisfl:
## Suggestion for safer channel usage in fetchers

This is a toy implementation of my suggestion in the retro.

As a demo, I replaced the channel in the IAM fetcher with this structure. Notice how the fetcher doesn't have access to the channel anymore so it's not possible to end up in a goroutine leak.

If we refactor everything like this, we are going to end with

```
func NewCisAwsFactory(log *logp.Logger, cfg aws.Config, ch pipeline.ContextAwareChannel[fetching.ResourceInfo], identity *awslib.Identity) FetchersMap {
```

making it impossible to use the raw channel in new / old fetchers.

### Full diff

```diff
diff --git a/pipeline/pipeline.go b/pipeline/pipeline.go
index 55c7686..bcec271 100644
--- a/pipeline/pipeline.go
+++ b/pipeline/pipeline.go
@@ -47,3 +47,20 @@ func Step[In any, Out any](ctx context.Context, log *logp.Logger, inputChannel c
 
[...]

PR #1399, opened 2023-10-05
Remove context from structs

orestisfl:
### Summary of your changes
Contexts in structs are not a good practice in Go: https://go.dev/blog/context-and-structs

Even though this shouldn't affect the current code, we should make sure we avoid this practice so that we don't run in problems in the future.

Enable the [containedctx](https://github.com/sivchari/containedctx) linter.

`posture` and `Manager` are exceptions since one is called by the libbeat API and the other needs it for goroutine canceling (code could be refactored there as well but is a bigger change).

I propose to not immediately merge this as it might introduce conflicts for the 8.11 branch during the QA cycle.

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/691
- Related: https://github.com/elastic/cloudbeat/issues/1108
- Related: https://github.com/elastic/cloudbeat/issues/1218

### Checklist


resources/providers/awslib/ec2/provider_test.go
```diff
@@ -270,119 +268,6 @@ func TestProvider_GetEbsEncryptionByDefault(t *testing.T) {
 	}
 }
 
-func TestProvider_DescribeInstances(t *testing.T) {
```
orestisfl, line 273: Removed these tests as they are misleadingly empty

resources/providers/azurelib/inventory/provider_test.go
```diff
@@ -145,29 +142,16 @@ func (s *ProviderTestSuite) TestGetString() {
 	}
 }
 
-func (s *ProviderTestSuite) TestProviderInit() {
```
orestisfl, line 148: Test only checks mock

resources/providers/gcplib/inventory/provider_test.go
```diff
@@ -68,8 +67,8 @@ func (s *ProviderTestSuite) TestProviderInit() {
 		ClientOpts: []option.ClientOption{},
 	}
 
-	initMock.On("Init", s.ctx, s.logger, gcpConfig).Return(&Provider{}, nil).Once()
-	provider, err := initMock.Init(s.ctx, s.logger, gcpConfig)
+	initMock.On("Init", mock.Anything, s.logger, gcpConfig).Return(&Provider{}, nil).Once()
```
amirbenun, line 70: `EXPECT`

amirbenun (approved):
💯

Issue #1391, opened 2023-10-04
[CIS Azure] Batch Fetcher: Group resources by Subscription ID


orestisfl:
**Motivation**
Currently, batch asset types like bastions are sent together regardless of their subscription ID, making the assumption that they are all from the same subscription. We need to change this.

This is necessary for organization support but might be beneficial for the single account case as well

**Definition of done**
What needs to be completed at the end of this task
- [x] Maybe use https://github.com/samber/lo#groupby
- [x] Resources are grouped according to subscription id

**Out of scope**
Expanding support of multi-account onboarding

PR #1400, opened 2023-10-05
Azure: Batch Fetcher: Group resources by subscription IDs and types

orestisfl:
### Summary of your changes
This has the side-benefit of only requiring one query for all types.

Utilizes [`lo.GroupBy`](https://github.com/samber/lo#groupby)

Can be backported to 8.11 if seen as beneficial.

### Screenshot/Data

Can't really E2E test without having access to multiple subscriptions.

### Related Issues
- Fixes: #1391 

### Checklist


resources/fetching/fetchers/azure/batch_fetcher.go
```diff
@@ -50,21 +52,26 @@ func NewAzureBatchAssetFetcher(log *logp.Logger, ch chan fetching.ResourceInfo,
...
 
-		if len(assets) == 0 {
-			continue
-		}
+	if len(allAssets) == 0 {
+		return nil
+	}
+
+	assetGroups := lo.GroupBy(allAssets, func(item inventory.AzureAsset) string {
+		return fmt.Sprintf("%s-%s", item.SubscriptionId, item.Type)
```
jeniawhite, line 65: We just need to make sure that we do not have rules that require all resources of a subscription for evaluation (of multiple types grouped).

jeniawhite (approved):

resources/fetching/fetchers/azure/batch_fetcher.go
orestisfl: Yes :+1: But I think we will notice that while implementing a rule that has that requirement.

PR #1415, opened 2023-10-16
CloudFormation: Fix missing tags in cloud-security account

orestisfl:
### Summary of your changes

This is a temporary fix that tags required resources for the build to succeed. We can revert it after the QA cycle to find a more elegant solution.


maxcold:
The change to Cloudformation looks good to me, but I don't know much about `just`. Can approve it, if you are blocked @orestisfl , otherwise would leave it to `cloudbeat` team to approve

orestisfl:
> The change to Cloudformation looks good to me, but I don't know much about `just`. Can approve it, if you are blocked @orestisfl , otherwise would leave it to `cloudbeat` team to approve

Yes, no worries, mostly just CCing you :)

gurevichdmitry (commented):
@orestisfl I'm unsure if it's a good idea to address this at the template level. Given that we're making these templates available for public use, it may not be appropriate to include data specific to our AWS cloud specifications.

orestisfl:
@gurevichdmitry, agreed, that's why we can revert this after the QA cycle. However, a clean solution is not easily apparent right now:

1. Adding the tags only in the dev scripts (`just`) will mean we cannot truly E2E test CF deployment
2. Adding a parameter for tags means that product is also affected

We are also investigating if we can work-around the requirements in a different way in the meantime.

oren-zohar (approved):

orestisfl:
Pipeline passed at https://github.com/orestisfl/cloudbeat/actions/runs/6536219417/job/17747423571

PR #1425, opened 2023-10-16
Dependabot: Run for all go modules

orestisfl:
### Summary of your changes
We can start this as an experiment and see how it works out


oren-zohar (approved):
Let's give it a shot. Do you think we should also have auto approve for this kind of PRs?

orestisfl:
> Let's give it a shot. Do you think we should also have auto approve for this kind of PRs?

Perhaps? Let's test it like this first

Issue #1320, opened 2023-09-05
CloudFormation: Upload to static link parametrized only by version


orestisfl:
**Motivation**
Currently, whenever we make a change to CloudFormation, we need to wait for the deployment to complete and then update the integrations [manifest file](https://github.com/elastic/integrations/blob/main/packages/cloud_security_posture/manifest.yml) with the new url in a new PR. This creates extra friction on each change and it's not necessary for our versioning model: New stack releases update the integration as well, so the current integration should always use its version's CloudFormation template.

In https://github.com/elastic/cloudbeat/blob/5cfd955370dc9d3411f0750bed6cb1ba4f3fe3e3/.github/workflows/publish-cloudformation.yml#L31 we use `$DATE` for the filename. I propose to only use `$VERSION` instead.

Alternatively, we can upload to both `$VERSION-$DATE` and `$VERSION` in case we need to freeze the file in a specific version for the integration. This should be the exception though, not the rule. 

**Definition of done**
What needs to be completed at the end of this task
- [x] Update url to depend only on `$VERSION`
- [x] Check how the s3 api reacts to duplicates, do we need to remove the old file first or is it automatic?
- [ ] Update integration url to link to current version
  - [ ] Ideally, re-use the version from the manifest variables 
- [ ] (Optional) Also upload to `$VERSION-$DATE`

Issue #1364, opened 2023-09-21
Lint `deploy/` shell scripts with shellcheck


orestisfl:
**Motivation**
[Shellcheck](https://www.shellcheck.net/) is a very popular shell script linter, its usage is recommended as a best practice when developing shell scripts.

Expand the shellcheck linter usage to other scripts in our repo, especially the ones that are used by users, like [deploy.sh](https://github.com/elastic/cloudbeat/blob/main/deploy/deployment-manager/deploy.sh).

For example, one potential problem I see in that script is that the `set -e` option is not set which means that the script will continue even when commands start failing. But, enabling that option is not enough because we are affected by rule [SC2155](https://www.shellcheck.net/wiki/SC2155) in our variable exports.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Expand pre-commit file regex to include more files
- [ ] Evaluate and fix or ignore failures

**Related tasks/epics**
Shellcheck introduced in https://github.com/elastic/cloudbeat/pull/1357

oren-zohar:
Closed by: https://github.com/elastic/cloudbeat/pull/1534

orestisfl:
Not all scripts are linted yet, re-opening

romulets:
- [ ] Include `deploy/deployment-manager/` scripts as well

romulets:
It will be covered by https://github.com/elastic/cloudbeat/pull/1568

PR #1427, opened 2023-10-17
CloudFormation: Upload to link parametrized only by version

orestisfl:
### Summary of your changes
Also tidies-up script and enables shellcheck for it

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/1320

### Checklist


.pre-commit-config.yaml
```diff
@@ -76,7 +76,7 @@ repos:
       - id: shfmt
         args: ["-i", "4"]
       - id: shellcheck
-        files: deploy/azure/.*sh
+        files: (deploy/azure/.*sh|scripts/publish_cft.sh)
```
amirbenun, line 79: Not for this PR, but we should probably apply it on all `*.sh` files right?

amirbenun (approved):

.pre-commit-config.yaml
orestisfl: I agree, https://github.com/elastic/cloudbeat/issues/1364

PR #1435, opened 2023-10-17
Dependabot: Fix group syntax, add label, bump limit

orestisfl:
### Summary of your changes
- Groups required a `patterns` key
- Allow up to 10 PRs at a time for go mod updates
- Add label to all PRs

image
https://github.com/elastic/cloudbeat/network/updates

See fix working in https://github.com/orestisfl/cloudbeat-tmp/network/updates


.github/dependabot.yml
```diff
@@ -9,18 +9,32 @@ updates:
...
+          - "github.com/elastic/*"
       k8s:
-        "k8s.io/*"
+        patterns:
+          - "k8s.io/*"
+    labels:
+      - "backport-skip"
+      - "dependency"
+      - "go"
+    open-pull-requests-limit: 10
```
oren-zohar, line 30: seems a bit much

oren-zohar (approved):

.github/dependabot.yml
orestisfl: It should be a one-off and we probably don't want to miss on updates because of PR limits. Let's see how it plays out.

PR #1438, opened 2023-10-17
go.mod: Update beats and elastic libraries

orestisfl:
### Summary of your changes
Done with:
```
go get -u github.com/elastic/beats/v7@10b198c985eb95c16405b979c63847881a199aba
```

This updates to gRPC to a version that should not be flagged by Snyk


oren-zohar (approved):

oren-zohar (approved):
We should automate this too. I think APM does it with a script

Issue #1474, opened 2023-10-25
[CIS Azure] Fix 7.1 Ensure an Azure Bastion Host Exists


orestisfl:
https://github.com/elastic/cloudbeat/issues/1258#issuecomment-1777127858

**Describe the bug**
The 7.1 rule should fail when no bastions exist

**To Reproduce**
Delete Bastion in Azure, check for failed finding

orestisfl:
This should be fixed again for 8.12+ in the `main` branch, https://github.com/elastic/cloudbeat/pull/1475 is only applicable for 8.11.

kfirpeled:
We miss instructions how to remediate and how make sure it fails in order to verify it

cc: @orestisfl , @oren-zohar

orestisfl:
> are there any pre-requisites that should be done in advnace before we start to test these rules?

Not really, bastions can be freely deleted and re-created in the team's subscription.

If needed, I can elaborate further on what is expected and what is not but that would perhaps diminish the effect of QAing a rule since I will instruct how to QA this rule according to my own understanding, which might be different from the QAers' understanding.

orestisfl:
Verified as broken by @maxcold, working on a fix to be re-tested in next QA batch

romulets:
Tests:

| Scenario | Evaluation Result | Expected |
|---|---|----|
| Standard Bastions present | passed | ✅ |
| No Bastions present | failed | ✅ |
| Basic Bastion present | pass | ✅ |

<img width="1461" alt="image" src="https://github.com/elastic/cloudbeat/assets/5350001/1208b6af-e513-461e-9ec1-b5ec8d147836">


LGTM 👍

Issue #1258, opened 2023-08-21
[CIS Azure] Virtual Machines rules


jeniawhite:
We would like to implement the following rules:
- [ ] 7.1 Ensure an Azure Bastion Host Exists
- [x] 7.2 Ensure Virtual Machines are utilizing Managed Disks
- [x] 7.3 Ensure that 'OS and Data' disks are encrypted with Customer Managed Key (CMK)
- [x] 7.4 Ensure that 'Unattached disks' are encrypted with 'Customer Managed Key' (CMK)

orestisfl:
@kfirpeled Please only verify the 7.1 rule

maxcold:
@animehart pls check this one with @jeniawhite

animehart:
had a chat with @jeniawhite , this will be verified on next BC (BC5)

PR #1475, opened 2023-10-25
[8.11] Azure: Send empty batch resources when missing

orestisfl:
### Screenshot/Data
image

### Related Issues
- Fixes: #1474 
- Related: #1258 

### Checklist

### DIFF
--- a/resources/fetching/fetchers/azure/batch_fetcher.go
+++ b/resources/fetching/fetchers/azure/batch_fetcher.go
@@ -57,27 +57,27 @@ func (f *AzureBatchAssetFetcher) Fetch(ctx context.Context, cMetadata fetching.C
 		}
 
 		if len(assets) == 0 {
-			if assetType == inventory.BastionAssetType {
-				// Hack backported into 8.11:
-				// Get the first subscription (should only be one), get the sub id and name and populate a mock resource
-				// with an empty array.
-				subId, subName := func() (string, string) {
-					for subId, subName := range f.provider.GetSubscriptions() {
-						return subId, subName
-					}
-					return "", ""
-				}()
-
-				success := f.write(ctx, fetching.ResourceInfo{
-					Resource: &EmptyBastionResource{
-						subId:   subId,
-						subName: subName,
-					},
-					CycleMetadata: cMetadata,
-				})
-				if !success {
-					return nil
+			// Hack backported into 8.11:
+			// Get the first subscription (should only be one), get the sub id and name and populate a mock resource
+			// with an empty array.
+			subId, subName := func() (string, string) {
+				for subId, subName := range f.provider.GetSubscriptions() {
+					return subId, subName
 				}
+				return "", ""
+			}()
+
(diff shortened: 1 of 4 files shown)


resources/fetching/fetchers/azure/batch_fetcher.go
```diff
@@ -57,29 +57,58 @@ func (f *AzureBatchAssetFetcher) Fetch(ctx context.Context, cMetadata fetching.C
 		}
 
 		if len(assets) == 0 {
+			if assetType == inventory.BastionAssetType {
```
jeniawhite, line 60: I think that we need the same behavior for Activity Logs as well.
Because if we do not have any Activity Logs configured we would like to fail rules that should verify that Activity Logs are configured.

jeniawhite (changes requested):

resources/fetching/fetchers/azure/batch_fetcher.go
```diff
@@ -57,29 +57,58 @@ func (f *AzureBatchAssetFetcher) Fetch(ctx context.Context, cMetadata fetching.C
...
 
 		if len(assets) == 0 {
+			// Hack backported into 8.11:
+			// Get the first subscription (should only be one), get the sub id and name and populate a mock resource
+			// with an empty array.
+			subId, subName := func() (string, string) {
+				for subId, subName := range f.provider.GetSubscriptions() {
+					return subId, subName
+				}
+				return "", ""
```
jeniawhite, line 67: This might create problems in the UI.
Since this should be used to create unique IDs.
Maybe if we cannot pull anything we should just skip? Sending empty strings might break some things in the UI. We need to verify that.

resources/fetching/fetchers/azure/batch_fetcher.go
```diff
@@ -57,29 +57,58 @@ func (f *AzureBatchAssetFetcher) Fetch(ctx context.Context, cMetadata fetching.C
...
+				Resource: &EmptyBatchResource{
+					Type:    pair.Type,
+					SubType: pair.SubType,
+					SubId:   subId,
+					SubName: subName,
+				},
+				CycleMetadata: cMetadata,
+			})
+			if !success {
+				return nil
```
jeniawhite, line 80: Maybe if we cannot send a batch we should continue and try to send other batches?
The same goes for other cases of returning nil on errors.
Also the API calls for listing types as well since if we get an error there we essentially skip the whole batch fetcher, but we can at least try to query other batches.

orestisfl: This is just for case of context canceled, where we return (as we did before)

resources/fetching/fetchers/azure/batch_fetcher.go
orestisfl: This should never happen, it would mean we have no subscriptions, so we can't scan anything anyway. Added a sanity check anyway.

jeniawhite (approved):

PR #1525, opened 2023-11-06
update-beats: Set -x flag to print commands

orestisfl:
### Summary of your changes
Easier to debug and follow what commands are running in the CI.


.ci/scripts/update-beats.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euo pipefail
+set -euxo pipefail
```
oren-zohar, line 2: should we have this in all of our scripts?

oren-zohar (approved):

PR #1528, opened 2023-11-07
update-beats: Use hermit for go

orestisfl:
### Summary of your changes
Use hermit for proper go version.

For some reason, running `git diff` triggers the PR correctly. I haven't determined exactly why.

### Screenshot/Data
Example PR: https://github.com/orestisfl/cloudbeat/pull/14. Note that this is done without the apm CI actions because I don't have vault access. Instead, I called updatecli directly from the GH action: https://github.com/orestisfl/cloudbeat/blob/d517e53a6d27679912f6a89fd55660317b98df68/.github/workflows/update-beats.yml

CC @v1v

### DIFF
--- a/.ci/scripts/update-beats.sh
+++ b/.ci/scripts/update-beats.sh
@@ -4,3 +4,4 @@ set -euxo pipefail
 BEATS_VERSION=${1:?Missing version argument}
 go get "github.com/elastic/beats/v7@$BEATS_VERSION"
 go mod tidy
+git diff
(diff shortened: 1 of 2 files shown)


v1v:
>  I haven't determined exactly why.

IIRC, updatecli needs some specific stdout when using the `shell target`, see https://www.updatecli.io/docs/plugins/resource/shell/#_shell_target.

v1v (approved):

oren-zohar (approved):

PR #1544, opened 2023-11-13
Updatecli: Update pre-commit as well

orestisfl:
### Summary of your changes
Added to hermit pipeline to reduce PR spam and because some of the hooks
are inter-connected with hermit packages. For example, this way we are
going to have the same version of `regal` in hermit and pre-commit.


oren-zohar (approved):
LGTM

PR #1549, opened 2023-11-14
update-hermit: Consider major versions as well

orestisfl:
### Summary of your changes

`hermit upgrade` will only upgrade minor versions for packages that use semantic versioning. Not all of our hermit packages use semantic versioning and even for those that do, we want to be at least aware of new major versions and update to those if they don't cause too much of a breakage.
Uninstalling and re-installing will always install to the latest major version.

This fixes an issue with `pre-commit autoupdate` not being able to update our `black` hook because it's too old to understand newer syntax.

Additionally, skip the poetry-based checks in pre-commit as that is not installed.

### Screenshot/Data
Example PR: https://github.com/orestisfl/cloudbeat/pull/23

### DIFF
--- a/.ci/updatecli/scripts/update-hermit.sh
+++ b/.ci/updatecli/scripts/update-hermit.sh
@@ -1,27 +1,38 @@
 #!/bin/bash
 set -euxo pipefail
 
-bin/hermit upgrade \
-    aws-iam-authenticator \
-    awscli \
-    elastic-package \
-    gcloud \
-    gh \
-    golangci-lint \
-    jq \
-    just \
-    kind \
-    mage \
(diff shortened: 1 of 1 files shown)


oren-zohar:
> Additionally, skip the poetry-based checks in pre-commit as that is not installed.

@orestisfl can you elaborate on this part?

orestisfl:
> > Additionally, skip the poetry-based checks in pre-commit as that is not installed.
> 
> @orestisfl can you elaborate on this part?

Yes, poetry is not available via hermit because it needs a more complicated installation. Additionally, these checks don't have any value here as the version updates are not affecting them.

PR #1555, opened 2023-11-14
Add Azure credentials type in configuration 

moukoublen:
### Summary of your changes

Based on the discussion in [this comment](https://github.com/elastic/integrations/pull/8376#discussion_r1390451815), this PR adds the Azure credentials type in the config and uses it to identify which Azure `TokenCredential` to use.

Enumerator values are on par with [this PR](https://github.com/elastic/kibana/pull/171069).

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/resources/providers/azurelib/auth/credentials_test.go
+++ b/resources/providers/azurelib/auth/credentials_test.go
@@ -48,9 +48,10 @@ func TestConfigProvider_GetAzureClientConfig(t *testing.T) {
 			name: "Should return a ClientSecretCredential",
 			config: config.AzureConfig{
 				Credentials: config.AzureClientOpt{
-					TenantID:     "tenant_a",
-					ClientID:     "client_id",
-					ClientSecret: "secret",
+					ClientCredentialsType: config.AzureClientCredentialsTypeSecret,
+					TenantID:              "tenant_a",
+					ClientID:              "client_id",
+					ClientSecret:          "secret",
 				},
 			},
 			authProviderInitFn: func(m *MockAzureAuthProviderAPI) {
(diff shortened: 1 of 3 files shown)


resources/providers/azurelib/auth/credentials_test.go
```diff
@@ -117,6 +121,7 @@ func TestConfigProvider_GetAzureClientConfig(t *testing.T) {
 			name: "Should return a ClientCertificateCredential",
 			config: config.AzureConfig{
 				Credentials: config.AzureClientOpt{
+					ClientCredentialsType:     config.AzureClientCredentialsTypeCertificate,
```
orestisfl, line 124: I would also add tests to cover the cases where credentials type is empty / unknown value.

moukoublen: The first test case has empty credentials type value.  I added one with `unknown client credentials type`.

resources/providers/azurelib/auth/credentials_test.go
```diff
@@ -44,13 +44,26 @@ func TestConfigProvider_GetAzureClientConfig(t *testing.T) {
...
 		},
+		{
+			name: "Should return a DefaultAzureCredential on unknown client credentials type",
+			config: config.AzureConfig{
+				Credentials: config.AzureClientOpt{
+					ClientCredentialsType: "unknown",
+				},
+			},
+			authProviderInitFn: initDefaultCredentialsMock(nil),
+			want: &AzureFactoryConfig{
```
orestisfl, line 55: Sorry for being nitpicky but this should probably return an error, it signifies a mismatch between integration and cloubeat

moukoublen: Sure,  I added an error in that case. Not in an empty string, though. Empty string fallbacks to default.

orestisfl (approved):
:+1:

PR #1559, opened 2023-11-15
Azure management groups

### LINKED ISSUE #1300: [CIS Azure] Support organizations in ARG (Azure resource graph)
**Motivation**
We are querying the resources with the help of [ARG](https://learn.microsoft.com/en-us/azure/governance/resource-graph/overview), it has a support for querying multiple subscriptions and management groups.

```golang
// QueryRequest - Describes a query to be executed.
type QueryRequest struct {
	// REQUIRED; The resources query.
	Query *string

	// An array of facet requests to be computed against the query result.
	Facets []*FacetRequest

	// Azure management groups against which to execute the query. Example: [ 'mg1', 'mg2' ]
	ManagementGroups []*string

	// The query evaluation options
	Options *QueryRequestOptions

	// Azure subscriptions against which to execute the query.
	Subscriptions []*string
}
```

**Definition of done**
What needs to be completed at the end of this task
- [ ] `cloudbeat` should query the relevant subscriptions and management groups

**Out of scope**
What should not be included in this task

**Related tasks/epics**
Reference related issues and epics

orestisfl:
### Related Issues
- Fixes #1300

### Checklist


resources/providers/azurelib/governance/management_group.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/samber/lo"
+
+	"github.com/elastic/cloudbeat/resources/fetching"
+	"github.com/elastic/cloudbeat/resources/providers/azurelib/inventory"
```
moukoublen, line 29: [Optional/Quesation] Since this package imports the same-layer package inventory and also seems pretty small; why not merge them inside inventory (having a `Provider` and a `GovernanceProvider` ?

orestisfl: I feel it is better to start separating things instead of continuously enlarging one single package, it feels that it will go against go [best practices](https://go.dev/blog/package-names).

For example, `governance.Provider` is a better name than `inventory.GovernanceProvider`.

moukoublen: In general, I agree, in this case, since they are both "inventory" in a way, might be good enough. I mostly have in mind the [tiny package syndrome](https://about.sourcegraph.com/blog/go/idiomatic-go#tiny-package-syndrome) pitfall.  (_For example we now have another provider `azurelib.Provider` that hides these two providers that they hide the azure SDK calls_). 

Which might be ok. I just got a feeling of a bit of complexity when first looked at it and thus the question. 

Totally optional.

moukoublen (approved):

Issue #1220, opened 2023-08-13
Fix periodic CI


gurevichdmitry:
**Motivation**

The periodic CI is currently failing due to a redundant "Build" step, which needs refactoring and updating to resolve the issue.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Fix Build job steps
- [ ] Uncomment workflow scheduler

**Out of scope**
What should not be included in this task

**Related tasks/epics**
Reference related issues and epics

oren-zohar:
It was decided to remove the Periodic CI, as it's not needed anymore. cc @gurevichdmitry

PR #1568, opened 2023-11-18
shell fmt all files

oren-zohar:
### Summary of your changes
Running `shellcheck` on all files
I auto-applied the fixes from `shellcheck` by using the following command:
```sh
find . -type f -name '*.sh' | sed 's|^\./||' | xargs shellcheck -f diff | git apply
```

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/1364
- Closes: https://github.com/elastic/cloudbeat/issues/1220


romulets:
To close https://github.com/elastic/cloudbeat/issues/1364 please also include `deploy/deployment-manager/` scripts in the linter by dropping line https://github.com/elastic/cloudbeat/blob/main/.pre-commit-config.yaml#L81

PR #1578, opened 2023-11-22
Update regal and enable use-if and use-contains rules

orestisfl:
### Summary of your changes

Issue https://github.com/StyraInc/regal/issues/468 explains the rationale of this rule, it's going to be required in rego v1.

Big diff here but the change is pretty trivial in all files.

### Related Issues

This was initially done in https://github.com/elastic/cloudbeat/pull/1570 but mergify insists on closing that PR for some reason.


oren-zohar (approved):
🫣

Issue #1418, opened 2023-10-16
[CIS Azure] Logging and Monitoring rules


jeniawhite:
We would like to implement the following rules:
- [x] 5.3.1 Ensure Application Insights are Configured
- [ ] 5.1.4 Ensure the storage account containing the container with activity logs is encrypted with Customer Managed Key
- [x] 5.1.2 Ensure Diagnostic Setting captures appropriate categories
- [x] 5.1.3 Ensure the Storage Container Storing the Activity Logs is not Publicly Accessible
- [ ] 5.1.5 Ensure that logging for Azure Key Vault is 'Enabled'

PR #1583, opened 2023-11-23
CIS Azure rule 5.1.2

moukoublen:
### Summary of your changes

This PR introduces a new fetcher that uses the azure `armmonitor` SDK in order to fetch diagnostic settings. 
It also implements the rule `CIS 5.1.3`.

### Screenshot/Data

Screenshot 2023-11-28 at 3 14 33 PM

### Related Issues

Related: https://github.com/elastic/cloudbeat/issues/1418

### Checklist

#### Introducing a new rule?

- [x] Add relevant unit tests

### DIFF
--- a/resources/fetching/fetchers/azure/insights_batch_fetcher.go
+++ b/resources/fetching/fetchers/azure/insights_batch_fetcher.go
@@ -1,0 +1,92 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package fetchers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/elastic/elastic-agent-libs/logp"
+	"github.com/samber/lo"
+
+	"github.com/elastic/cloudbeat/resources/fetching"
+	"github.com/elastic/cloudbeat/resources/fetching/cycle"
+	"github.com/elastic/cloudbeat/resources/providers/azurelib"
+	"github.com/elastic/cloudbeat/resources/providers/azurelib/inventory"
+)
+
+type AzureInsightsBatchAssetFetcher struct {
+	log        *logp.Logger
+	resourceCh chan fetching.ResourceInfo
+	provider   azurelib.ProviderAPI
+}
+
+func NewAzureInsightsBatchAssetFetcher(log *logp.Logger, ch chan fetching.ResourceInfo, provider azurelib.ProviderAPI) *AzureInsightsBatchAssetFetcher {
+	return &AzureInsightsBatchAssetFetcher{
+		log:        log,
+		resourceCh: ch,
+		provider:   provider,
+	}
+}
+
+func (f *AzureInsightsBatchAssetFetcher) Fetch(ctx context.Context, cycleMetadata cycle.Metadata) error {
+	f.log.Info("Starting AzureInsightsBatchAssetFetcher.Fetch")
+	subscriptions, err := f.provider.GetSubscriptions(ctx, cycleMetadata)
+	if err != nil {
+		return fmt.Errorf("azure insights fetcher: error while receiving subscriptions: %w", err)
+	}
+
+	assets, err := f.provider.ListDiagnosticSettingsAssetTypes(ctx, cycleMetadata, lo.Keys(subscriptions))
+	if err != nil {
+		return fmt.Errorf("azure insights fetcher: error while receiving diagnostic settings: %w", err)
+	}
+
+	// group and send by subscription id
+	subscriptionGroups := lo.GroupBy(assets, func(item inventory.AzureAsset) string {
+		return item.SubscriptionId
+	})
+
+	for subId, subscription := range subscriptions {
+		batchAssets := subscriptionGroups[subId]
+		if batchAssets == nil {
+			batchAssets = []inventory.AzureAsset{} // Use empty array instead of nil
+		}
+
+		select {
+		case <-ctx.Done():
+			err := ctx.Err()
+			f.log.Infof("AzureInsightsBatchAssetFetcher.Fetch context err: %s", err.Error())
+			return err
+		case f.resourceCh <- fetching.ResourceInfo{
+			CycleMetadata: cycleMetadata,
+			Resource: &AzureBatchResource{
+				typePair: typePair{
+					Type:    fetching.MonitoringIdentity,
+					SubType: fetching.AzureDiagnosticSettingsType,
+				},
+				Subscription: subscription,
+				Assets:       batchAssets,
+			},
+		}:
+		}
+	}
+
+	return nil
+}
+
+func (f *AzureInsightsBatchAssetFetcher) Stop() {}
--- a/resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
+++ b/resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
@@ -63,6 +63,7 @@ func TestAzureInsightsBatchAssetFetcher(t *testing.T) {
 		mockInventoryError  error
 		cycleMetadata       cycle.Metadata
 		expected            []fetching.ResourceInfo
+		expectedMetaData    []fetching.ResourceMetadata
 		expectedErr         bool
 	}{
 		"error": {
@@ -71,6 +72,7 @@ func TestAzureInsightsBatchAssetFetcher(t *testing.T) {
 			mockInventoryError:  errors.New("mock error"),
 			cycleMetadata:       newCycle(123),
 			expected:            nil,
+			expectedMetaData:    []fetching.ResourceMetadata{},
 			expectedErr:         true,
 		},
 
@@ -86,6 +88,21 @@ func TestAzureInsightsBatchAssetFetcher(t *testing.T) {
 						typePair:     expectedPair,
 						Subscription: subscriptionOne,
 						Assets:       []inventory.AzureAsset{},
+					},
+				},
+			},
+			expectedMetaData: []fetching.ResourceMetadata{
+				{
+					ID:      "azure-diagnostic-settings-sub1",
+					Type:    "monitoring",
+					SubType: "azure-diagnostic-settings",
+					Name:    "azure-diagnostic-settings-sub1",
+					Region:  "global",
+					CloudAccountMetadata: fetching.CloudAccountMetadata{
+						AccountId:        "sub1",
+						AccountName:      "subName1",
+						OrganisationId:   "",
+						OrganizationName: "",
 					},
 				},
 			},
@@ -121,6 +138,34 @@ func TestAzureInsightsBatchAssetFetcher(t *testing.T) {
 						Assets: []inventory.AzureAsset{
 							{Id: "id3", Name: "name3", SubscriptionId: "sub2"},
 						},
+					},
+				},
+			},
+			expectedMetaData: []fetching.ResourceMetadata{
+				{
+					ID:      "azure-diagnostic-settings-sub2",
+					Type:    "monitoring",
+					SubType: "azure-diagnostic-settings",
+					Name:    "azure-diagnostic-settings-sub2",
+					Region:  "global",
+					CloudAccountMetadata: fetching.CloudAccountMetadata{
+						AccountId:        "sub2",
+						AccountName:      "subName2",
+						OrganisationId:   "",
+						OrganizationName: "",
+					},
+				},
+				{
+					ID:      "azure-diagnostic-settings-sub1",
+					Type:    "monitoring",
+					SubType: "azure-diagnostic-settings",
+					Name:    "azure-diagnostic-settings-sub1",
+					Region:  "global",
+					CloudAccountMetadata: fetching.CloudAccountMetadata{
+						AccountId:        "sub1",
+						AccountName:      "subName1",
+						OrganisationId:   "",
+						OrganizationName: "",
 					},
 				},
 			},
@@ -167,6 +212,19 @@ func TestAzureInsightsBatchAssetFetcher(t *testing.T) {
 			sortResourceInfoSlice(got)
 
 			assert.Equal(t, tc.expected, got)
+
+			for _, resourceInfo := range got {
+				elm, err := resourceInfo.GetElasticCommonData()
+				require.NoError(t, err)
+				assert.Equal(t, map[string]any(nil), elm)
+			}
+
+			metadata := lo.Map(got, func(item fetching.ResourceInfo, _ int) fetching.ResourceMetadata {
+				metadata, err := item.GetMetadata()
+				require.NoError(t, err)
+				return metadata
+			})
+			assert.Equal(t, tc.expectedMetaData, metadata)
 		})
 	}
 }
--- a/resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
+++ b/resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
@@ -224,7 +224,7 @@ func TestAzureInsightsBatchAssetFetcher(t *testing.T) {
 				require.NoError(t, err)
 				return metadata
 			})
-			assert.Equal(t, tc.expectedMetaData, metadata)
+			assert.ElementsMatch(t, tc.expectedMetaData, metadata)
 		})
 	}
 }
--- a/resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
+++ b/resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
@@ -20,6 +20,7 @@ package fetchers
 import (
 	"context"
 	"errors"
+	"reflect"
 	"sort"
 	"testing"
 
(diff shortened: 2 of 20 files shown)


security-policies/bundle/compliance/cis_azure/rules/cis_5_1_2/rule.rego
```diff
@@ -0,0 +1,37 @@
...
+		{"Resource": data_adapter.resource},
+	)
+}
+
+required_categories_enabled if {
+	diagnostic_settings_category_enabled("Administrative") == true
+	diagnostic_settings_category_enabled("Alert") == true
+	diagnostic_settings_category_enabled("Policy") == true
+	diagnostic_settings_category_enabled("Security") == true
+} else = false
```
orestisfl, line 24: Prefer `default` over `else` https://docs.styra.com/regal/rules/style/default-over-else

resources/providers/azurelib/inventory/provider.go
```diff
@@ -224,3 +261,75 @@ func getString(data map[string]any, key string) string {
...
+	js, err := d.MarshalJSON()
+	if err != nil {
+		return nil, err
+	}
+
+	m := map[string]any{}
+	err = json.Unmarshal(js, &m)
+	if err != nil {
+		return nil, err
+	}
```
orestisfl, line 318:
```suggestion
	js, err := d.MarshalJSON()
	if err != nil {
		return nil, fmt.Errorf("failed to marshal json: %w", err)
	}

	m := map[string]any{}
	err = json.Unmarshal(js, &m)
	if err != nil {
		return nil, fmt.Errorf("failed to unmarshal json: %w", err)
	}
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -224,3 +261,75 @@ func getString(data map[string]any, key string) string {
...
+	assets := make([]AzureAsset, 0, 8)
+
+	for i := range res {
+		for _, v := range res[i].Value {
+			if v == nil {
+				continue
+			}
+			a, err := transformDiagnosticSettingsResource(v, subID, subName)
+			if err != nil {
+				return nil, fmt.Errorf("error wile parsing azure asset model: %w", err)
```
orestisfl, line 275:
```suggestion
				return nil, fmt.Errorf("error while parsing azure asset model: %w", err)
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -224,3 +261,75 @@ func getString(data map[string]any, key string) string {
...
+	err = json.Unmarshal(js, &m)
+	if err != nil {
+		return nil, err
+	}
+
+	return m, nil
+}
+
+func readPager[T any](ctx context.Context, pager *runtime.Pager[T]) ([]T, error) {
+	res := make([]T, 0, 8)
```
orestisfl, line 324: Why 8?

orestisfl, line 324:
```suggestion
	var res []T
```

let's go for simpler and consistent

resources/providers/azurelib/inventory/provider_test.go
```diff
@@ -211,3 +213,217 @@ func Test_generateQuery(t *testing.T) {
...
 }
+
+func TestTransformDiagnosticSettingsClientListResponses(t *testing.T) {
+	response := func(v []*armmonitor.DiagnosticSettingsResource) armmonitor.DiagnosticSettingsClientListResponse {
+		return armmonitor.DiagnosticSettingsClientListResponse{
+			DiagnosticSettingsResourceCollection: armmonitor.DiagnosticSettingsResourceCollection{Value: v},
+		}
+	}
+	s := func(s string) *string { return &s }
+	b := func(b bool) *bool { return &b }
```
orestisfl, line 224: There is `to.Ptr()`

resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		aj, _ := (&r[j]).Resource.(*AzureBatchResource)
+		if len(aj.Assets) > 0 {
+			y = aj.Assets[0].SubscriptionId
+		}
+
+		return x > y
+	})
+}
```
orestisfl, line 173:
```suggestion
			assert.ElementsMatch(t, tc.expected, got)
		})
	}
}
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -37,14 +40,15 @@ import (
 type Provider struct {
 	log           *logp.Logger
 	client        *AzureClientWrapper
-	subscriptions map[string]string
+	subscriptions map[string]string // Key: subscription ID, value: display name
```
orestisfl, line 43:
```suggestion
	subscriptions map[string]string
```

This is going to be refactored in https://github.com/elastic/cloudbeat/pull/1559 anyway

resources/providers/azurelib/inventory/provider.go
```diff
@@ -74,17 +79,26 @@ type ProviderInitializerAPI interface {
 func (p *ProviderInitializer) Init(ctx context.Context, log *logp.Logger, azureConfig auth.AzureFactoryConfig) (ServiceAPI, error) {
 	log = log.Named("azure")
 
-	clientFactory, err := armresourcegraph.NewClientFactory(azureConfig.Credentials, nil)
+	// init azure resource graph client
```
orestisfl, line 82:
```suggestion
```

I feel the code is already clear enough here

resources/providers/azurelib/inventory/provider.go
```diff
@@ -74,17 +79,26 @@ type ProviderInitializerAPI interface {
...
+	resourceGraphClientFactory, err := armresourcegraph.NewClientFactory(azureConfig.Credentials, nil)
 	if err != nil {
 		return nil, err
 	}
+	resourceGraphClient := resourceGraphClientFactory.NewClient()
 
-	client := clientFactory.NewClient()
+	// init azure monitor client
```
orestisfl, line 89:
```suggestion
```

I feel the code is already clear enough here

resources/providers/azurelib/inventory/provider.go
```diff
@@ -74,17 +79,26 @@ type ProviderInitializerAPI interface {
 func (p *ProviderInitializer) Init(ctx context.Context, log *logp.Logger, azureConfig auth.AzureFactoryConfig) (ServiceAPI, error) {
 	log = log.Named("azure")
 
-	clientFactory, err := armresourcegraph.NewClientFactory(azureConfig.Credentials, nil)
+	// init azure resource graph client
+	resourceGraphClientFactory, err := armresourcegraph.NewClientFactory(azureConfig.Credentials, nil)
 	if err != nil {
 		return nil, err
```
orestisfl, line 85:
```suggestion
		return nil, fmt.Errorf("failed to init resource graph client: %w", err)
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -74,17 +79,26 @@ type ProviderInitializerAPI interface {
...
 	}
+	resourceGraphClient := resourceGraphClientFactory.NewClient()
 
-	client := clientFactory.NewClient()
+	// init azure monitor client
+	diagnosticSettingsClient, err := armmonitor.NewDiagnosticSettingsClient(azureConfig.Credentials, nil)
+	if err != nil {
+		return nil, err
```
orestisfl, line 92:
```suggestion
		return nil, fmt.Errorf("failed to init monitor client: %w", err)
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -157,6 +171,29 @@ func (p *Provider) ListAllAssetTypesByName(ctx context.Context, assetsGroup stri
 	return resAssets, nil
 }
 
+func (p *Provider) ListDiagnosticSettingsAssetTypes(ctx context.Context) ([]AzureAsset, error) {
+	p.log.Infof("Listing Azure Diagnostic Monitor Settings")
+
+	assets := make([]AzureAsset, 0, 8)
```
orestisfl, line 177:
```suggestion
	var assets []AzureAsset
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -157,6 +171,29 @@ func (p *Provider) ListAllAssetTypesByName(ctx context.Context, assetsGroup stri
 	return resAssets, nil
 }
 
+func (p *Provider) ListDiagnosticSettingsAssetTypes(ctx context.Context) ([]AzureAsset, error) {
+	p.log.Infof("Listing Azure Diagnostic Monitor Settings")
```
orestisfl, line 175:
```suggestion
	p.log.Info("Listing Azure Diagnostic Monitor Settings")
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -224,3 +261,75 @@ func getString(data map[string]any, key string) string {
 	value, _ := data[key].(string)
 	return value
 }
+
+func transformDiagnosticSettingsClientListResponses(res []armmonitor.DiagnosticSettingsClientListResponse, subID, subName string) ([]AzureAsset, error) {
+	assets := make([]AzureAsset, 0, 8)
```
orestisfl, line 266:
```suggestion
	var assets []AzureAsset
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -157,6 +171,29 @@ func (p *Provider) ListAllAssetTypesByName(ctx context.Context, assetsGroup stri
...
+		if err != nil {
+			return nil, err
+		}
+		a, err := transformDiagnosticSettingsClientListResponses(res, subID, subName)
+		if err != nil {
+			return nil, err
+		}
+		if len(a) > 0 {
+			assets = append(assets, a...)
+		}
```
orestisfl, line 191:
```suggestion
		assets = append(assets, a...)
```

Should be redundant

resources/providers/azurelib/inventory/provider.go
```diff
@@ -74,17 +79,26 @@ type ProviderInitializerAPI interface {
...
+	}
 
 	// We wrap the client, so we can mock it in tests
 	wrapper := &AzureClientWrapper{
 		AssetQuery: func(ctx context.Context, query armresourcegraph.QueryRequest, options *armresourcegraph.ClientResourcesOptions) (armresourcegraph.ClientResourcesResponse, error) {
-			return client.Resources(ctx, query, options)
+			return resourceGraphClient.Resources(ctx, query, options)
+		},
+		AssetDiagnosticSettings: func(subID string, options *armmonitor.DiagnosticSettingsClientListOptions) *runtime.Pager[armmonitor.DiagnosticSettingsClientListResponse] {
+			return diagnosticSettingsClient.NewListPager(fmt.Sprintf("/subscriptions/%s/", subID), options)
```
orestisfl, line 101:
```suggestion
		AssetDiagnosticSettingsPager: func(subID string, options *armmonitor.DiagnosticSettingsClientListOptions) *runtime.Pager[armmonitor.DiagnosticSettingsClientListResponse] {
			return diagnosticSettingsClient.NewListPager(fmt.Sprintf("/subscriptions/%s/", subID), options)
```

resources/fetching/fetchers/azure/insights_batch_fetcher.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (f *AzureInsightsBatchAssetFetcher) Fetch(ctx context.Context, cMetadata fetching.CycleMetadata) error {
+	f.log.Info("Starting AzureInsightsBatchAssetFetcher.Fetch")
+	subscriptions := f.provider.GetSubscriptions()
+
+	assets, err := f.provider.ListDiagnosticSettingsAssetTypes(ctx)
+	if err != nil {
+		f.log.Errorf("AzureInsightsBatchAssetFetcher.Fetch failed to fetch diagnostic settings: %s", err.Error())
+		return err
```
orestisfl, line 51:
```suggestion
		return err
```

Since we return the error, let's avoid double-logging it

resources/providers/azurelib/inventory/provider.go
moukoublen: When we initialize the slice with nil and then append, we have some reallocation happening (check [this](https://go.dev/play/p/oQ0xdHsDt9k)).

Each capacity expansion (reallocation) doubles the previous capacity value. 

So its very common for example if you append 5 elements, to have 4 reallocation. 

So performance-wise is might be better to avoid that with a capacity (not fixed length) that might be large "enough" from the beginning. 

what do you think?

orestisfl (commented):
Some more comments

resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
moukoublen: We have 2 levels of slices. One is the resource and the second is the asset of each resource.

resources/providers/azurelib/inventory/provider.go
orestisfl: I feel that when we are fetching information that potentially has to travel over a wire in the atlantic, optimizing for allocation of a couple of bytes is premature and makes reading the code slightly worse as the reader wonders "why 8". Still optional though.

moukoublen: I will change it in this pr to nil and open a ticket/proposal to revisit our slices initialization in general, perhaps and we can discuss it there.

resources/providers/azurelib/inventory/provider.go
orestisfl: WDYT about this?

resources/providers/azurelib/inventory/provider.go
```diff
@@ -157,6 +169,29 @@ func (p *Provider) ListAllAssetTypesByName(ctx context.Context, assetsGroup stri
...
+
+	var assets []AzureAsset
+
+	for subID, subName := range p.subscriptions {
+		pager := p.client.AssetDiagnosticSettingsPager(subID, nil)
+		res, err := readPager(ctx, pager)
+		if err != nil {
+			return nil, err
+		}
+		a, err := transformDiagnosticSettingsClientListResponses(res, subID, subName)
```
orestisfl, line 183: Maybe a better name than `a`?

resources/providers/azurelib/inventory/provider.go
```diff
@@ -224,3 +259,75 @@ func getString(data map[string]any, key string) string {
 	value, _ := data[key].(string)
 	return value
 }
+
+func transformDiagnosticSettingsClientListResponses(res []armmonitor.DiagnosticSettingsClientListResponse, subID, subName string) ([]AzureAsset, error) {
+	var assets []AzureAsset
+
```
orestisfl, line 265:
```suggestion
func transformDiagnosticSettingsClientListResponses(assets []AzureAsset, response []armmonitor.DiagnosticSettingsClientListResponse, subID, subName string) error {
```

instead of having another azure assets array, WDYT about re-using the previous one. Will also lead to one less `append` call.

Additionally, slightly better name than `res`

resources/providers/azurelib/inventory/provider.go
```diff
@@ -224,3 +259,75 @@ func getString(data map[string]any, key string) string {
...
+	for i := range res {
```
orestisfl, line 266:
```suggestion
	for _, values := range res {
```

resources/providers/azurelib/inventory/provider.go
```diff
@@ -224,3 +259,75 @@ func getString(data map[string]any, key string) string {
...
+	m := map[string]any{}
+	err = json.Unmarshal(js, &m)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal json: %w", err)
+	}
+
+	return m, nil
+}
+
+func readPager[T any](ctx context.Context, pager *runtime.Pager[T]) ([]T, error) {
```
orestisfl, line 321: I am not sure what's the value of flattening the pager? Instead of `for pager.More()` we have a `for _, value := res`, so same level of indentation but without the benefits of a pager + slightly more code.

moukoublen: I am not sure I understood; this is a helper function to avoid performing this loop inside the loop of subscriptions. 

I feel by doing that the subscription loop is more readable.

resources/providers/azurelib/inventory/provider.go
moukoublen: If you want to append/mutate the slice given in the first argument, you will have to pass a pointer to the slice otherwise, you might or you might not actually mutate it (depending if reallocation happens or not during append). 

And having passed a pointer to slice and appending like this `*slice = append(*slice, ...)` feels kind of difficult reading. 

If we need to improve performance, initializing with capacity would be much more effective that this I think.

resources/providers/azurelib/inventory/provider.go
moukoublen: I tend to follow [this](https://research.swtch.com/names) naming philosophy where the larger the scope of the variable, the more descriptive the name should be. 

So something that is used directly after initialization could be named single letter e.t.c.

But sure if we don't follow that rule, I could change it.

resources/providers/azurelib/inventory/provider.go
moukoublen: Done

resources/fetching/fetchers/azure/insights_batch_fetcher_test.go
```diff
@@ -0,0 +1,195 @@
...
+				require.NoError(t, err)
+			}
+
+			got := testhelper.CollectResources(ch)
+
+			// sort to ensure assertion
+			sortResourceInfoSlice(tc.expected)
+			sortResourceInfoSlice(got)
+
+			assert.Equal(t, tc.expected, got)
```
orestisfl, line 169: Can you also please check the `GetMetadata()` and `GetElasticCommonData()` calls?

moukoublen: Sure, done.

resources/fetching/fetchers/azure/insights_batch_fetcher.go
```diff
@@ -0,0 +1,92 @@
...
+	for subId, subscription := range subscriptions {
+		batchAssets := subscriptionGroups[subId]
+		if batchAssets == nil {
+			batchAssets = []inventory.AzureAsset{} // Use empty array instead of nil
+		}
+
+		select {
+		case <-ctx.Done():
+			err := ctx.Err()
+			f.log.Infof("AzureInsightsBatchAssetFetcher.Fetch context err: %s", err.Error())
```
orestisfl, line 73:
```suggestion
```

if we are going to return the error anyway

orestisfl (approved):

resources/fetching/fetchers/azure/insights_batch_fetcher.go
moukoublen: will remove it in the next pr with the next rule

PR #1599, opened 2023-11-27
Add slack notification

gurevichdmitry:
### Summary of your changes

This PR introduces Slack notifications for both the `create-environment` and `delete-environment` workflows.

### Screenshot/Data
Create Environment message
Create Environment

Destroy environment message
Destroy Environment

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,189 @@
...
+        return "#36a64f"
+    if status == "failure":
+        return "#D40E0D"
+    return ""
+
+
+class BuildSlackException(Exception):
+    """
+    Custom exception class for errors related to building Slack notifications.
+    """
```
oren-zohar, line 70: Why do we need an empty wrapper exception? who's going to raise it?

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,189 @@
...
+    Retrieve the value of the specified environment variable.
+
+    Parameters:
+        env_var (str): The name of the environment variable to retrieve.
+
+    Returns:
+        str: The value of the specified environment variable.
+    """
+    value = os.environ.get(env_var)
+    if value is None or value == "":
```
oren-zohar, line 22: would that be the same?
```suggestion
    if not value
```

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Returns:
+        str: The Slack color corresponding to the GitHub Actions job status.
+             Possible values: "good" for success, "danger" for failure, or an empty string.
+    """
+    if status == "success":
+        return "#36a64f"
+    if status == "failure":
+        return "#D40E0D"
+    return ""
```
oren-zohar, line 64: This function could be replaced by a dict

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,189 @@
...
+        github_actor = check_env_var("GITHUB_ACTOR")
+        github_run_url = check_env_var("RUN_URL")
+        job_status = check_env_var("JOB_STATUS")
+        kibana_url = check_env_var("KIBANA_URL")
+        s3_bucket = check_env_var("S3_BUCKET")
+        deployment_name = check_env_var("DEPLOYMENT_NAME")
+        stack_version = check_env_var("STACK_VERSION")
+        docker_image = os.getenv("DOCKER_IMAGE_OVERRIDE", "N/A")
+        if docker_image == "":
+            docker_image = "N/A"
```
oren-zohar, line 92: The approach to check for the existence of an environment variable is inconsistent. For `DOCKER_IMAGE_OVERRIDE`, os.getenv with a default value is used directly in the run function. It might be more consistent to use the `check_env_var` function for all environment variable retrievals or refactor `check_env_var` to handle optional environment variables with default values.

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,189 @@
...
+    return ""
+
+
+class BuildSlackException(Exception):
+    """
+    Custom exception class for errors related to building Slack notifications.
+    """
+
+
+def run():
```
oren-zohar, line 73: This function is quite long and handles multiple aspects of the script. Consider breaking it down into smaller functions for better readability and maintainability. For example, creating a separate function to generate the Slack payload.

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,189 @@
...
+                                },
+                            ],
+                        },
+                    ],
+                },
+            ],
+        }
+        set_output("payload", json.dumps(slack_payload))
+
+    except BuildSlackException as err:
```
oren-zohar, line 184: could other exceptions occur? (e.g., issues with network, JSON operations). Consider adding a generic exception handler or specific handlers for known exception types that could arise. What is going to throw `BuildSlackException`?

.github/actions/slack-notification/slack_payload.py
```diff
@@ -0,0 +1,155 @@
...
+
+def main():
+    """
+    Main function to process environment variables and generate a Slack-compatible message.
+    """
+    if os.environ.get("MESSAGE"):
+        json_data = process_message_env()
+    elif os.environ.get("PAYLOAD"):
+        payload = replace_user_mentions(os.environ["PAYLOAD"], load_users_from_env())
+        json_data = json.loads(payload)
```
oren-zohar, line 144: try-catch json.load for safety?

.github/actions/slack-notification/slack_payload.py
```diff
@@ -0,0 +1,155 @@
...
+
+def load_users_from_env():
+    """
+    Load a dictionary of users from the 'USERS' environment variable.
+
+    Returns:
+    dict: A dict of users.
+    """
+    users = os.environ.get("USERS")
+    if users is None or users == "":
```
oren-zohar, line 111: would that be the same?
```suggestion
    if not users
```

.github/actions/slack-notification/slack_payload.py
```diff
@@ -0,0 +1,155 @@
...
+                      "type": "section",
+                      "text": {
+                          "type": "mrkdwn",
+                          "text": processed_message
+                      }
+                  }
+              ]
+          }
+    """
+    message = os.environ.get("MESSAGE", "No message")
```
oren-zohar, line 80: if there's no message you'll keep doing all the logic on the `"No message"` string

.github/actions/slack-notification/slack_payload.py
```diff
@@ -0,0 +1,155 @@
...
+    Parameters:
+    - message (str): The message to be processed.
+    - users (dict): A dict of users.
+
+    Returns:
+    str: The message with user mentions replaced with Slack user IDs.
+    """
+    for user in users:
+        message = message.replace(f"{user}", f"<@{users[user]}>")
+    return message
```
oren-zohar, line 133: instead of trying to replace all user's handles with their ID one by one, we can just map the user to it's Slack handle. Btw, how do we handle a case where the mapping doesn't contain the user?

.github/actions/slack-notification/README.md
```diff
@@ -0,0 +1,83 @@
+## About
```
oren-zohar, line 1: Can you add some information about the Slack handles mapping in vault and how to update it?

oren-zohar (changes requested):
There is some inconsistency in error handling, in some cases we print and in other we raise an exception.

.ci/scripts/prepare_slack_data.py
gurevichdmitry: fixed

.ci/scripts/prepare_slack_data.py
gurevichdmitry: fixed

.github/actions/slack-notification/slack_payload.py
gurevichdmitry: fixed

.github/actions/slack-notification/slack_payload.py
gurevichdmitry: Added an `if` statement to skip logic for `'no message'`.

.github/actions/slack-notification/slack_payload.py
gurevichdmitry: I'm not sure I understand how to map the user to their Slack handle. Could you provide an example? The payload is loaded as a string, and users is a dictionary from the vault. This dictionary contains a mapping of GitHub user to Slack user id. In the case where no user is found, the payload will contain the GitHub username, indicating that the user is not mapped to their Slack id.

.ci/scripts/prepare_slack_data.py
gurevichdmitry: The `run()` function, which executes the main flow, contains a try-except block. Providing a general Exception to catch is flagged as a warning by `pylint`. To address this, a custom BuildSlackException class was created to serve as an empty wrapper exception.

.ci/scripts/prepare_slack_data.py
gurevichdmitry: As mentioned earlier, `BuildSlackException` is an empty wrapper. If other exceptions occur, it will be raised as a general exception. If issues related to JSON operations arise, the corresponding JSON exception will be raised. Since this script is basic, providing specific handlers or a generic exception handler is not necessary in this case.

.github/actions/slack-notification/slack_payload.py
gurevichdmitry: I'm not convinced we need it there. If json.loads fails, an exception will occur, and we shouldn't proceed with payload generation.

.github/actions/slack-notification/slack_payload.py
gurevichdmitry: refactored

.github/actions/slack-notification/README.md
gurevichdmitry: docs were updated

.ci/scripts/prepare_slack_data.py
gurevichdmitry: The run() function is long because of the complex structure of the Slack payload, which involves many parameters. To address this, we could create a `dataclass` to hold these parameters for better organization. However, since this script is standalone, breaking it into smaller functions may not provide substantial benefits.

.github/actions/slack-notification/slack_payload.py
```diff
@@ -0,0 +1,144 @@
...
+
+    Parameters:
+    - message (str): The message to be processed.
+
+    Returns:
+    str: The message with user mentions replaced with Slack user IDs.
+    """
+    github_user = os.environ.get("GITHUB_ACTOR")
+    if not github_user:
+        print(f"::warning::Invalid value for GITHUB_ACTOR: {github_user}")
```
oren-zohar, line 116:
```suggestion
        print(f"::warning::GitHub user information is missing. Falling back to the original message")
```

.github/actions/slack-notification/slack_payload.py
```diff
@@ -0,0 +1,144 @@
...
+        return message
+    slack_user_id = os.environ.get("SLACK_USER")
+    if not slack_user_id:
+        print(f"::warning::Invalid value for SLACK_USER: {slack_user_id}")
```
oren-zohar, line 120:
```suggestion
        print(f"::warning::Slack user ID is missing. Falling back to the original message")
```

oren-zohar (changes requested):

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,175 @@
...
+    Parameters:
+        env_var (str): The name of the environment variable to retrieve.
+
+    Returns:
+        str: The value of the specified environment variable.
+    """
+    value = os.environ.get(env_var)
+    if not value:
+        if env_var == "DOCKER_IMAGE_OVERRIDE":
+            return "N/A"
```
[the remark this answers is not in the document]
gurevichdmitry: Docker override is provided only when using a BC version; for snapshots, this value will not be applicable

.github/actions/slack-notification/slack_payload.py
```diff
@@ -0,0 +1,144 @@
...
+                  }
+              ]
+          }
+    """
+    message = os.environ.get("MESSAGE", "No message")
+    if message != "No message":
+        message = replace_user_mentions(message)
+        if os.environ["URL_ENCODED"] == "true":
+            message = unquote(message)
+        message = "\n".join(line.strip() for line in message.splitlines())
```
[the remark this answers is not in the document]
gurevichdmitry: Sure. This section is relevant to sending a simple message. In the action, we are utilizing `GITHUB_OUTPUT` to output the message for Slack notification. The challenge arises when the message is a multiline string. When attempting to write to `GITHUB_OUTPUT`, GitHub Actions notifies about formatting problems with the string. There are several solutions to work around this issue. One approach is to use delimiters between lines, as suggested by [GitHub Workflows](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings). However, the GitHub solution warns that the delimiter chosen should not occur on a line of its own within the value, and this can be problematic for arbitrary multiline messages. To avoid this, another solution involves encoding the string. Therefore, if a multiline encoded string is provided, it should be decoded first. In this case, the encoding of the string was selected to handle the multiline format, and the last line of code normalizes the string by removing leading and trailing whitespaces from each line.

.ci/scripts/prepare_slack_data.py
gurevichdmitry: Removed empty wrapper and improved code for better clarity and error handling.

.ci/scripts/prepare_slack_data.py
gurevichdmitry: Refactored code using `dataclass` implementation.

.github/actions/slack-notification/slack_payload.py
gurevichdmitry: removed "No message"

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def run():
+    """
+    Main function to run the Slack notification workflow.
+
+    This function is responsible for validating environment variables, generating a Slack payload,
+    setting GitHub Action outputs, and handling exceptions related to building Slack notifications.
+    """
+    try:
```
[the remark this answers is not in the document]
gurevichdmitry: Simplified code with improved exception handling.

.ci/scripts/prepare_slack_data.py
```diff
@@ -0,0 +1,175 @@
...
+def set_output(name: str, value: str):
+    """
+    Set an output variable for the GitHub Actions workflow.
+
+    Parameters:
+        name (str): The name of the output variable.
+        value (str): The value to set for the output variable.
+    """
+    with open(os.environ["GITHUB_OUTPUT"], "a", encoding="utf-8") as fh:
+        print(f"{name}={value}", file=fh)
```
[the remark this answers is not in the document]
gurevichdmitry: This code is designed to be straightforward and standalone within its script. Keeping each script independent makes it easy to understand without introducing unnecessary complexity or dependencies on shared functions across scripts.

.github/actions/slack-notification/slack_payload.py
orestisfl:
```
message = os.environ.get("MESSAGE", "No message")
if message != "No message":
    ...
```

I suggest moving this code and the following lines in a new function and adding the informative comment you typed in the other thread as a docstring:
> Sure. This section is relevant to sending a simple message. In the action, we are utilizing `GITHUB_OUTPUT` to output the message for Slack notification. The challenge arises when the message is a multiline string. When attempting to write to `GITHUB_OUTPUT`, GitHub Actions notifies about formatting problems with the string. There are several solutions to work around this issue. One approach is to use delimiters between lines, as suggested by [GitHub Workflows](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#multiline-strings). However, the GitHub solution warns that the delimiter chosen should not occur on a line of its own within the value, and this can be problematic for arbitrary multiline messages. To avoid this, another solution involves encoding the string. Therefore, if a multiline encoded string is provided, it should be decoded first. In this case, the encoding of the string was selected to handle the multiline format, and the last line of code normalizes the string by removing leading and trailing whitespaces from each line.

gurevichdmitry: The code already has a dedicated function for handling messages. The documentation has been updated.

oren-zohar (approved):
Let's just revert the dataclass to a dict following our discussion

PR #1625, opened 2023-12-06
[Cloud Security] bump version of main to 8.13 (for 8.12 FF)

maxcold:
Bumping stack versions for 8.12 Feature Freez


orestisfl (commented):
Can you also please update:
- deploy/azure/ARM-for-single-account.json
- deploy/azure/ARM-for-organization-account.json

maxcold:
@orestisfl done

bin/hermit.hcl
```diff
@@ -1,4 +1,4 @@
 env = {
-  "CLOUDBEAT_VERSION": "8.12.0",
+  "CLOUDBEAT_VERSION": "8.13.0",
```
orestisfl, line 2: I think this needs to be done separately because the snapshot builds for 8.13 are not available yet

PR #1649, opened 2023-12-11
RULES.md collapsible CIS sections

oren-zohar:
### Summary of your changes
Per @tehilashn's request - making the CIS benchmark table collapsible

### Screenshot/Data
https://github.com/oren-zohar/cloudbeat/blob/collapsible-rules-md/security-policies/RULES.md


orestisfl (approved):
`poetry.lock` update seems unnecessary but no problem merging that as well.

PR #1652, opened 2023-12-12
Azure Assets Enricher

moukoublen:
### Summary of your changes
This PR introduces Azure asset "enricher" interface and creates a slice of enrichers in Azure asset fetcher.

After the assets are fetched from the provider, azure assset fetcher runs all enrichers serially and enriches the assets. 

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/resources/fetching/fetchers/azure/assets_enricher.go
+++ b/resources/fetching/fetchers/azure/assets_enricher.go
@@ -29,12 +29,12 @@ import (
 	"github.com/elastic/cloudbeat/resources/utils/strings"
 )
 
-type AssetsEnricherAPI interface {
+type AssetsEnricher interface {
 	Enrich(ctx context.Context, cycleMetadata cycle.Metadata, assets []inventory.AzureAsset) error
 }
 
-func initEnrichers(provider azurelib.ProviderAPI) []AssetsEnricherAPI {
-	var enrichers []AssetsEnricherAPI
+func initEnrichers(provider azurelib.ProviderAPI) []AssetsEnricher {
+	var enrichers []AssetsEnricher
 
 	enrichers = append(enrichers, storageAccountEnricher{provider: provider})
 
--- a/resources/fetching/fetchers/azure/assets_enricher.go
+++ b/resources/fetching/fetchers/azure/assets_enricher.go
@@ -1,0 +1,88 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package fetchers
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/samber/lo"
+
+	"github.com/elastic/cloudbeat/resources/fetching/cycle"
+	"github.com/elastic/cloudbeat/resources/providers/azurelib"
+	"github.com/elastic/cloudbeat/resources/providers/azurelib/inventory"
+	"github.com/elastic/cloudbeat/resources/utils/strings"
+)
+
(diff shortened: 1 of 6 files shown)


resources/fetching/fetchers/azure/assets_enricher.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"github.com/samber/lo"
+
+	"github.com/elastic/cloudbeat/resources/fetching/cycle"
+	"github.com/elastic/cloudbeat/resources/providers/azurelib"
+	"github.com/elastic/cloudbeat/resources/providers/azurelib/inventory"
+	"github.com/elastic/cloudbeat/resources/utils/strings"
+)
+
+type AssetsEnricherAPI interface {
```
romulets, line 32: Any specific reason to append it with `API`? I think `AssetsEnricher` is good enough

romulets (commented):
Given the horizon of more enrichers and potentially complex one, I think it's a good idea to move the `assets_enricher.go` to its own package `enricher/` and have `storageAccountEnricher` in a separate file `storage_account_enricher.go`.

But overall, I like it!

resources/fetching/fetchers/azure/assets_enricher.go
moukoublen: Indeed, and in general, I agree. I just saw the `API` postfix in some other interfaces in cloudbeat, and I decided to follow the pattern.

orestisfl: We usually add the `API` suffix when there is a struct and interface with the same name.

romulets: I don't think that's the case here, is it @moukoublen ?

moukoublen: Cool, I renamed it! 👍

moukoublen:
@romulets my suggestion would be (since `fetchers/azure` is still small enough) to not create an inner package. And put enrichers inside of `fetchers/azure`. If, at some point, we have a lot of enrichers / a lot of different functionalities, we can then extract them into a separate package.

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
```diff
@@ -0,0 +1,76 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
```
romulets, line 1: I'm used as pattern to have the "type" prepended to the file name, not appeneded. So `storage_account_assets_enricher`. It reads a bit more natural to me.

I'll follow what you decide @moukoublen, but wanted to bring the point up just in case you agree with me :D

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
```diff
@@ -0,0 +1,76 @@
...
+	provider azurelib.ProviderAPI
+}
+
+func (e storageAccountEnricher) Enrich(ctx context.Context, cycleMetadata cycle.Metadata, assets []inventory.AzureAsset) error {
+	subscriptions, err := e.provider.GetSubscriptions(ctx, cycleMetadata)
+	if err != nil {
+		return fmt.Errorf("storageAccountEnricher: error while getting subscription: %w", err)
+	}
+
+	diagSettings, err := e.provider.ListDiagnosticSettingsAssetTypes(ctx, cycleMetadata, lo.Keys(subscriptions))
```
romulets, line 42: Why use `lo` and not [exp/maps](https://pkg.go.dev/golang.org/x/exp/maps#Keys)? 

PS: I still don't understand why this wasn't move from exp to [std](https://pkg.go.dev/maps) with go 1.21 😢

romulets (approved):
Two small non blocker comments :)

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
moukoublen: Yes I agree `exp/maps` might be better but cloudbeat codebase uses `lo` so far, so perhaps we can open an issue to replace it cloudbeat-wide?

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
moukoublen: Indeed. The "advantage" of using names with a hierarchy format is that you have a "grouped" sorting by default as a side effect. 

So having many `assets_enricher_***.go` will be sorted together makes it easier to navigate. 

I could change it, though, if that's not useful.

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
romulets: Jah sure let's stick to lo if that's the pattern. Just curious why :)

romulets (approved):

PR #1657, opened 2023-12-14
golangci-lint: Make unused config stricter

orestisfl:
### Summary of your changes
Some of these changes seem suspicious. The changes are not wrong, they functionally change nothing, however, it could be an indicator that something is not working exactly as expected.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -18,6 +18,7 @@ linters:
     - testifylint
     - unconvert
     - unparam
+    - unused
     - whitespace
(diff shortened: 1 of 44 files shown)


romulets (commented):
Very neat the check on fields too! It seems that there are still quite a few unused fields tho

orestisfl:
> Very neat the check on fields too! It seems that there are still quite a few unused fields tho

Thanks, I think I needed to run the command once more to catch additional fields

romulets (approved):

Issue #1654, opened 2023-12-12
Clean up Azure Classical VMs


romulets:
**Motivation**
We are pulling Classical VM resources from ARG. [And according to Microsoft from September 6th 2023 no active Classical VM should exist](https://learn.microsoft.com/en-gb/azure/virtual-machines/classic-vm-deprecation).

It's good to investigate if this is a hard definition and indeed 0 Classical VMs exist, or if there are exceptions to the rule. And then make a judgement call if we need to keep scanning it or not. If not, clean it up

**Definition of done**
- [x] Clean up Classical VMs references and specific rules

orestisfl:
They were added in https://github.com/orestisfl/cloudbeat/commit/53e8ec1d9b11bd391f4bfbe86e8106d35d1fc441 but I don't see them being used anywhere, I think we can delete them :+1:

romulets:
There are no specific rules and I could not find any exceptions to the rule. Classical VMs are deprecated.

PR #1660, opened 2023-12-14
Clean up Azure Classical VMS

romulets:
### Summary of your changes

Clean up Azure Classical VMs pull, [they are completely deprecated](https://learn.microsoft.com/en-gb/azure/virtual-machines/classic-vm-deprecation)

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/1654


orestisfl (approved):
As a note, `azure-classic-vm` is not used anywhere in rego code

PR #1662, opened 2023-12-14
Azure: Fix wrong usage of subscription ID

orestisfl:
### Summary of your changes
There are two forms of subscription IDs:
- Short: e.g. 8d65815f-a5b6-402f-9298-045155da7d74
- Fully Qualified: e.g. /subscriptions/8d65815f-a5b6-402f-9298-045155da7d74

The batch fetcher was grouping assets using the first one and then was using the second one to access the keys from the group map.

This commit fixes that and also re-names the fields to avoid future confusion. A `ShortId` field has also been added to be used for accessing the group maps and formatting batch asset names.

### Screenshot/Data
**Before:**
screenshot-2023-12-14T18:07:06 EET
**After:**
screenshot-2023-12-14T18:06:58 EET

### Related Issues
Fixes https://github.com/elastic/cloudbeat/issues/1474

### Checklist


orestisfl:
@jeniawhite @moukoublen notice the resource id/name change in the screenshots

resources/providers/azurelib/governance/management_group.go
```diff
@@ -105,9 +112,10 @@ func (p *provider) scan(ctx context.Context) (map[string]Subscription, error) {
 
 		mg := managementGroups[strings.FromMap(parent, "name")]
 		subscriptions[asset.SubscriptionId] = Subscription{
```
moukoublen, line 114: Question: what will be the format of the **key** of the returned map `GetSubscriptions(...) (map[string]Subscription, error)`?

Because this key is used in `resources/fetching/fetchers/azure/insights_batch_fetcher.go` and is copied to `AzureAsset.SubscriptionId` while transforming azure SDK models to `AzureAsset`

orestisfl: Key map === asset.SubscriptionId === Subscription.ShortID

orestisfl: Key map === asset.SubscriptionId === Subscription.ShortID

moukoublen: [Not thoroughly thought idea] Perhaps we can create two string aliases like `type SubscriptionIDFull string`  and `type SubscriptionIDShort`, which might help in each case to know what kind of ID we have (even as map keys or everywhere else).

moukoublen (approved):

Issue #1602, opened 2023-11-29
Agentless CSPM sanity test


amirbenun:
### Background:
Besides of the serverless tests we have today that assert for regular agents findings we want to also verify that agentless also produces findings as expected.

### Definition of done:
- [x] Create a serverless project using our terraform
- [ ] Install all 3 cloud provider CSPM integrations
  - [x] AWS
  - [ ] GCP
  - [x] Azure
- [x] Assert for findings

### Related:
- https://github.com/elastic/cloudbeat/issues/1524

amirbenun:
Since I experienced issues with installing GCP and Azure, the first iteration will only test Agentless CSPM for AWS.
- https://github.com/elastic/cloudbeat/pull/1664

amirbenun:
Update:
- Azure will be testable after elastic-agent 8.12 release.
- GCP is blocked by https://github.com/elastic/kibana/issues/172687

PR #1664, opened 2023-12-14
Add agentless CSPM AWS findings assertion in serverless mode

amirbenun:
### Summary of your changes
Our sanity tests has already a serverless mode.
On this PR, the sanity tests install CSPM AWS integration on the agentless policy and verify it produces findings.

### Related Issues
- Resolves: https://github.com/elastic/cloudbeat/issues/1602

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


deploy/test-environments/fleet_api/src/install_agentless_integrations.py
```diff
@@ -0,0 +1,98 @@
...
+
+AGENT_POLICY_ID = "agentless"
+
+
+def generate_aws_integration_data():
+    """
+    Generate data for creating CSPM AWS integration
+    """
+    access_key_id = os.getenv("AWS_ACCESS_KEY_ID", "")
+    secret_access_key = os.getenv("AWS_SECRET_ACCESS_KEY", "")
```
gurevichdmitry, line 28: AWS keys are already available through `cnfg.aws_config`, so you can use:
```
aws_config = cnfg.aws_config
```
and then
```
...
"access_key_id": aws_config.access_key_id,
"secret_access_key": aws_config.secret_access_key
...
```

deploy/test-environments/fleet_api/src/install_agentless_integrations.py
```diff
@@ -0,0 +1,98 @@
...
+    return {
+        "name": "cspm_aws",
```
gurevichdmitry, line 30: May be add some randomization there. We have a function `generate_random_name` which is part of the `package_policy` module and can do that. You can use it like this:
`"name": generate_random_name("cspm_aws")`

deploy/test-environments/fleet_api/src/install_agentless_integrations.py
```diff
@@ -0,0 +1,98 @@
...
+            "azure.credentials.type": "manual",
+        },
+    }
+
+
+def generate_gcp_integration_data():
+    """
+    Generate data for creating CSPM GCP integration
+    """
+    application_credentials = os.getenv("GOOGLE_APPLICATION_CREDENTIALS", "")
```
gurevichdmitry, line 63: I think it would be better to retrieve credentials from the `cnfg` as well.

deploy/test-environments/fleet_api/src/install_agentless_integrations.py
```diff
@@ -0,0 +1,98 @@
...
+
+        logger.info(f"Created {package_policy}")
+
+        create_cspm_integration(
+            cfg=cnfg.elk_config,
+            pkg_policy=package_policy,
+            agent_policy_id=AGENT_POLICY_ID,
+            cspm_data={},
+        )
+        logger.info(f"Installation of {NAME} integration is done")
```
gurevichdmitry, line 98: `create-environment` workflow supports different elastic versions installation, and the main branch is always used for deployment. To prevent some misconfigurations, such as policy field discrepancies, and ensure integration availability, logic based on the package version has been implemented for each integration. It might be better to retrieve the current package version and prevent the installation of integrations if the version is less than required.

tests/agents_map.py
```diff
@@ -0,0 +1,60 @@
+"""
+Generate agent parameterization for pytest.
+"""
+import os
+import sys
+from loguru import logger
+from munch import Munch
+from configuration import elasticsearch
```
gurevichdmitry, line 8: I'm not sure the location of the` tests/agents_map.py` file is ideal. The tests folder serves as the root for the tests project and should contain all relevant project files and configurations. A better approach would be to store this file under the `tests/commonlib` path.

tests/agents_map.py
```diff
@@ -0,0 +1,60 @@
...
+CIS_AZURE_COMPONENT = "cloudbeat/cis_azure"
+
+
+class AgentExpectedMapping:
+    """
+    This class is used to map expected number of agents that are running each component.
+    """
+
+    def __init__(self):
+        agentless = os.getenv("TEST_AGENTLESS", "false") == "true"
```
gurevichdmitry, line 24: The same applies here; the processing of environment variables should be centralized in the `configuration.py` file.

deploy/test-environments/fleet_api/src/install_agentless_integrations.py
amirbenun: With serverless it is a bit different, you will always deploy the latest elastic stack so this check is redundant.

tests/integration/tests/test_sanity_checks.py
```diff
@@ -7,19 +7,23 @@
...
 from loguru import logger
+from configuration import elasticsearch
+from commonlib.utils import get_findings
+from commonlib.agents_map import CIS_AWS_COMPONENT, AgentExpectedMapping, AgentComponentMapping
 
 CONFIG_TIMEOUT = 120
 GCP_CONFIG_TIMEOUT = 600
 CNVM_CONFIG_TIMEOUT = 3600
+COMPONENTS_BACKOFF = 10
+COMPONENTS_TIMEOUT = 180
```
oren-zohar, line 20: Can you please add a comment? The names used here are pretty generic IMO

tests/commonlib/agents_map.py
```diff
@@ -0,0 +1,59 @@
...
+Generate agent parameterization for pytest.
+"""
+import os
+import sys
+from loguru import logger
+from munch import Munch
+from configuration import elasticsearch, agent
+
+sys.path.append(os.path.relpath("../deploy/test-environments/fleet_api/src"))
+from api.agent_policy_api import get_agents  # pylint: disable=wrong-import-position # noqa: E402
```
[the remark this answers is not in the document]
amirbenun: This import only works after the `sys.path.append` command as it loads code from another python project.

gurevichdmitry (approved):
@amirbenun great work, LGTM. 
Could you please verify that running the regular workflow (not agentless) is also functioning correctly?

Issue #1669, opened 2023-12-18
Enable revive linter


orestisfl:
**Motivation**
The [revive](https://github.com/mgechev/revive) linter offers various checks for Go best-practices and common bugs and can help with code consistency and code review. It is currently disabled in our golangci.yml.

**Definition of done**
What needs to be completed at the end of this task
- [x] Enable the revive linter
- [x] Investigate which rules are beneficial and enable them gradually, fixing any findings in the process

Rules:
- [x] ~~[add-constant](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#add-constant)~~
- [x] [argument-limit](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#argument-limit)
- [x] [atomic](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#atomic)
- [x] ~~[banned-characters](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#banned-characters)~~
- [x] ~~[bare-return](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bare-return)~~
- [ ] [blank-imports](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#blank-imports)
- [x] [bool-literal-in-expr](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#bool-literal-in-expr)
- [x] [call-to-gc](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#call-to-gc)
- [x] [cognitive-complexity](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#cognitive-complexity)
- [x] [comment-spacings](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#comment-spacings)
- [ ] [confusing-naming](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#confusing-naming)
- [x] [confusing-results](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#confusing-results)
- [x] [constant-logical-expr](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#constant-logical-expr)
- [x] [context-as-argument](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#context-as-argument)
[...]

PR #1670, opened 2023-12-18
golangci: Enable revive linter

orestisfl:
### Summary of your changes
Enable revive linter and manually disable most rules that fail except some of the most obvious ones that we should fix and prevent.

This enables all rules that don't fail to prevent future bugs.

### Screenshot/Data
Fixes the following:

```text
launcher/launcher_test.go:300:16: datarace: datarace: range value tt is captured (by-reference) in goroutine (revive)
				time.Sleep(tt.delay)
				           ^
resources/providers/awslib/multi_region_test.go:27:2: duplicated-imports: Package "github.com/aws/aws-sdk-go-v2/aws" already imported (revive)
	awssdk "github.com/aws/aws-sdk-go-v2/aws"
	^
resources/providers/awslib/multi_region.go:28:2: duplicated-imports: Package "github.com/aws/aws-sdk-go-v2/aws" already imported (revive)
	awssdk "github.com/aws/aws-sdk-go-v2/aws"
	^
resources/providers/awslib/cached_region_selector.go:34:40: var-declaration: should drop = 0 from declaration of var currentRegionCacheTTL; it is the zero value (revive)
	currentRegionCacheTTL time.Duration = 0
	                                      ^
resources/providers/awslib/ec2/nacl.go:38:2: comment-spacings: no space between comment delimiter and comment text (revive)
	//arn:aws:ec2:region:account-id:network-acl/network-acl-id
	^
resources/providers/awslib/ecr/provider.go:32:2: comment-spacings: no space between comment delimiter and comment text (revive)
	/// When repoNames is nil, it will describe all the existing Repositories
	^
resources/fetching/fetchers/aws/elb_fetcher.go:124:1: receiver-naming: receiver name res should be consistent with previous receiver name r for ElbResource (revive)
func (res ElbResource) GetElasticCommonData() (map[string]any, error) {
	return map[string]interface{}{
		"cloud.service.name": "ELB",
	}, nil
}
flavors/benchmark/builder/data_provider.go:28:35: unused-parameter: parameter 'event' seems to be unused, consider removing or renaming it as _ (revive)
func (d dataProvider) EnrichEvent(event *beat.Event, _ fetching.ResourceMetadata) error { return nil }
                                  ^
flavors/benchmark/builder/id_provider.go:22:28: unused-parameter: parameter 'resourceType' seems to be unused, consider removing or renaming it as _ (revive)
func (p *idProvider) GetId(resourceType string, resourceId string) string {
                           ^
resources/fetching/preset/aws_org_preset_test.go:49:13: flag-parameter: parameter 'drain' seems to be a control flag, avoid control coupling (revive)
func subtest(t *testing.T, drain bool) {
            ^
```

### Related Issues
- Related https://github.com/elastic/cloudbeat/issues/1669

### Checklist


.golangci.yaml
```diff
@@ -72,6 +71,60 @@ linters-settings:
...
+        disabled: true
+      - name: import-alias-naming
+        disabled: true
+      - name: import-shadowing
+        disabled: true
+      - name: indent-error-flow
+        disabled: true
+      - name: line-length-limit
+        disabled: true
+        arguments: [160]
```
oren-zohar, line 109: why disable it if you set the line length?

.golangci.yaml
```diff
@@ -72,6 +71,60 @@ linters-settings:
...
+      - name: redundant-import-alias
+        disabled: true
+      - name: unchecked-type-assertion
+        disabled: true
+      - name: unexported-naming
+        disabled: true
+      - name: unexported-return
+        disabled: true
+      - name: unhandled-error
+        disabled: true
```
oren-zohar, line 121: why skipping it?

.golangci.yaml
```diff
@@ -72,6 +71,60 @@ linters-settings:
...
+      - name: blank-imports
+        disabled: true
+      - name: cognitive-complexity
+        disabled: true
+      - name: confusing-naming
+        disabled: true
+      - name: confusing-results
+        disabled: true
+      - name: cyclomatic
+        disabled: true
```
oren-zohar, line 90: I think this is a good way to check for readability. Why are we disabling it?

launcher/launcher_test.go
```diff
@@ -289,6 +289,7 @@ func (s *LauncherTestSuite) TestWaitForUpdates() {
 	}
 
 	for _, tt := range testcases {
+		tt := tt
```
oren-zohar, line 292: ?

.golangci.yaml
orestisfl: Rules like this need larger changes so we can enable them one-by-one to make reviewing easier.

.golangci.yaml
orestisfl: Rules like this need larger changes so we can enable them one-by-one to make reviewing easier.

launcher/launcher_test.go
orestisfl: Common golang idiom, the `tt` variable needs to be captured when used in goroutines.

See https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/

This is fixed in 1.22 but we are still not in that version.

oren-zohar (approved):

Issue #1416, opened 2023-10-16
[CIS Azure] Storage Accounts rules


jeniawhite:
We would like to implement the following rules:
- [x] 3.11 Ensure Soft Delete is Enabled for Azure Containers and Blob Storage
- [ ] 3.5 Ensure Storage Logging is Enabled for Queue Service for 'Read', 'Write', and 'Delete' requests
- [x] 3.13 Ensure Storage logging is Enabled for Blob Service for 'Read', 'Write', and 'Delete' requests
- [x] 3.14 Ensure Storage Logging is Enabled for Table Service for 'Read', 'Write', and 'Delete' Requests

orestisfl:
## 3.5 Failed to verify :x: 

![screenshot-2024-02-20T13:43:04 CET](https://github.com/elastic/cloudbeat/assets/5778622/097a7a2c-7b5b-4915-b602-5e7deb59e349)
![image](https://github.com/elastic/cloudbeat/assets/5778622/6a1d4d63-16b7-46aa-809f-9c595a925a09)

## 3.11 verified :heavy_check_mark: 
![screenshot-2024-02-20T12:51:28 CET](https://github.com/elastic/cloudbeat/assets/5778622/bc6983ba-5c2d-4642-8990-f22faf09945b)


## 3.13 verified :heavy_check_mark: 
![screenshot-2024-02-20T13:10:00 CET](https://github.com/elastic/cloudbeat/assets/5778622/6b3297f2-be47-4675-a7d8-1cbc02a53db3)


## 3.14 verified :heavy_check_mark: 
![screenshot-2024-02-20T13:11:44 CET](https://github.com/elastic/cloudbeat/assets/5778622/1cb7a8ab-5446-40f6-9fa5-207f85da2bd4)

moukoublen:
@orestisfl I think to enable queue diagnostic settings, you have to follow this path:

  1. Storage Account
  2. Monitoring -> Diagnostic settings (not classic)
  3. Queue Diagnostic status should say `Enable`
  4. If not, click on it and in next screen `+ add diagnostic settings`

<img width="1932" alt="Screenshot 2024-02-20 at 4 38 26 PM" src="https://github.com/elastic/cloudbeat/assets/1380039/a9cdae2a-3a88-4346-b0e9-dc380f566863">
[...]

PR #1671, opened 2023-12-18
[CIS] Azure 3.11

moukoublen:
### Summary of your changes

Rule: Ensure Soft Delete is Enabled for Azure Containers and Blob Storage

Storage Account enriched extended by fetching blob service for each storage account (one request per storage account).
Blob Service is added at the `extension` field of each storage resource on which the rego rules are based.
 
### Screenshot/Data

Screenshot 2023-12-27 at 5 01 39 PM

### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/1416

### Checklist

#### Introducing a new rule?

- [x] Add relevant unit tests

### DIFF
--- a/resources/fetching/fetchers/azure/assets_enricher_storage_account.go
+++ b/resources/fetching/fetchers/azure/assets_enricher_storage_account.go
@@ -19,6 +19,7 @@ package fetchers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/samber/lo"
@@ -26,6 +27,7 @@ import (
 	"github.com/elastic/cloudbeat/resources/fetching/cycle"
 	"github.com/elastic/cloudbeat/resources/providers/azurelib"
 	"github.com/elastic/cloudbeat/resources/providers/azurelib/inventory"
+	"github.com/elastic/cloudbeat/resources/utils/maps"
 	"github.com/elastic/cloudbeat/resources/utils/strings"
 )
 
@@ -45,10 +47,14 @@ func (e storageAccountEnricher) Enrich(ctx context.Context, cycleMetadata cycle.
 	}
 	e.addUsedForActivityLogsFlag(assets, diagSettings)
 
+	if err := e.addStorageAccountBlobServices(ctx, assets); err != nil {
+		return fmt.Errorf("storageAccountEnricher: error while getting data protection settings: %w", err)
+	}
+
 	return nil
 }
 
-func (*storageAccountEnricher) addUsedForActivityLogsFlag(assets []inventory.AzureAsset, diagSettings []inventory.AzureAsset) {
+func (storageAccountEnricher) addUsedForActivityLogsFlag(assets []inventory.AzureAsset, diagSettings []inventory.AzureAsset) {
 	usedStorageAccountIDs := map[string]struct{}{}
 	for _, d := range diagSettings {
 		storageAccountID := strings.FromMap(d.Properties, "storageAccountId")
@@ -67,10 +73,54 @@ func (*storageAccountEnricher) addUsedForActivityLogsFlag(assets []inventory.Azu
 			continue
 		}
 
-		if a.Extension == nil {
-			a.Extension = make(map[string]any)
-		}
-		a.Extension["usedForActivityLogs"] = true
+		a.AddExtension(inventory.ExtensionUsedForActivityLogs, true)
+
 		assets[i] = a
 	}
+}
+
+func (e storageAccountEnricher) addStorageAccountBlobServices(ctx context.Context, assets []inventory.AzureAsset) error {
+	storageAccounts := lo.Filter(assets, func(item inventory.AzureAsset, index int) bool {
+		return item.Type == inventory.StorageAccountAssetType
+	})
+	if len(storageAccounts) == 0 {
+		return nil
+	}
+
+	dataProtection, err := e.provider.ListStorageAccountBlobServices(ctx, storageAccounts)
+	if err != nil {
+		return err
+	}
+
+	if len(dataProtection) == 0 {
+		return nil
+	}
+
+	perStorageAccountID := map[string]inventory.AzureAsset{}
+	for _, d := range dataProtection {
+		perStorageAccountID[strings.FromMap(d.Extension, inventory.ExtensionStorageAccountID)] = d
+	}
+
+	var errAgg error
+	for i, a := range assets {
+		if a.Type != inventory.StorageAccountAssetType {
+			continue
+		}
+
+		dataProtectionSettings, exist := perStorageAccountID[a.Id]
+		if !exist {
+			continue
+		}
+
+		blobService, err := maps.AsMapStringAny(dataProtectionSettings)
+		if err != nil {
+			errAgg = errors.Join(errAgg, err)
+		}
+
+		a.AddExtension(inventory.ExtensionBlobService, blobService)
+
+		assets[i] = a
+	}
+
+	return errAgg
 }
--- a/resources/providers/azurelib/inventory/asset.go
+++ b/resources/providers/azurelib/inventory/asset.go
@@ -26,6 +26,7 @@ const (
 	ActivityLogAlertAssetType          = "microsoft.insights/activitylogalerts"
 	ApplicationInsights                = "microsoft.insights/components"
 	BastionAssetType                   = "microsoft.network/bastionhosts"
+	BlobServiceAssetType               = "microsoft.storage/storageaccounts/blobservices"
 	ClassicStorageAccountAssetType     = "microsoft.classicstorage/storageaccounts"
 	DiagnosticSettingsAssetType        = "microsoft.insights/diagnosticSettings"
 	DiskAssetType                      = "microsoft.compute/disks"
@@ -47,6 +48,13 @@ const (
 	// Azure Resource Graph table groups
 	AssetGroupResources              = "resources"
 	AssetGroupAuthorizationResources = "authorizationresources"
+
+	// Extension keys
+	ExtensionBlobService         = "blobService"
+	ExtensionNetwork             = "network"
+	ExtensionUsedForActivityLogs = "usedForActivityLogs"
+	ExtensionStorageAccountID    = "storageAccountId"
+	ExtensionStorageAccountName  = "storageAccountName"
 )
 
 type AzureAsset struct {
@@ -62,6 +70,13 @@ type AzureAsset struct {
 	Type           string         `json:"type,omitempty"`
 	Sku            map[string]any `json:"sku,omitempty"`
 	Identity       map[string]any `json:"identity,omitempty"`
+}
+
+func (a *AzureAsset) AddExtension(key string, value any) {
+	if a.Extension == nil {
+		a.Extension = map[string]any{}
+	}
+	a.Extension[key] = value
 }
 
 func getAssetFromData(data map[string]any) AzureAsset {
--- a/resources/providers/azurelib/inventory/provider.go
+++ b/resources/providers/azurelib/inventory/provider.go
@@ -20,28 +20,33 @@ package inventory
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"fmt"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/monitor/armmonitor"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resourcegraph/armresourcegraph"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage"
 	"github.com/elastic/elastic-agent-libs/logp"
 
 	"github.com/elastic/cloudbeat/resources/fetching/cycle"
+	"github.com/elastic/cloudbeat/resources/utils/maps"
 	"github.com/elastic/cloudbeat/resources/utils/strings"
 )
 
 type azureClientWrapper struct {
 	AssetQuery              func(ctx context.Context, query armresourcegraph.QueryRequest, options *armresourcegraph.ClientResourcesOptions) (armresourcegraph.ClientResourcesResponse, error)
 	AssetDiagnosticSettings func(ctx context.Context, subID string, options *armmonitor.DiagnosticSettingsClientListOptions) ([]armmonitor.DiagnosticSettingsClientListResponse, error)
+	AssetBlobServices       func(ctx context.Context, subID string, clientOptions *arm.ClientOptions, resourceGroup, storageAccountName string, options *armstorage.BlobServicesClientListOptions) ([]armstorage.BlobServicesClientListResponse, error)
 }
 
 type ProviderAPI interface {
 	// ListAllAssetTypesByName List all content types of the given assets types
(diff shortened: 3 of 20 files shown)


resources/utils/maps/maps.go
```diff
@@ -0,0 +1,38 @@
...
+// under the License.
+
+package maps
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+func AsMapStringAny(item any) (map[string]any, error) {
```
romulets, line 25: I get this method makes life easier, but it's quite non performant for something that should be simple.

If the map will have something around 7 or so entries not sure if this function gives real gains, it's not that painful to create such a map. Wdyt?

security-policies/RULES.md
```diff
@@ -390,9 +390,9 @@
 
 ## AZURE CIS Benchmark
 
-### 44/151 implemented rules (29%)
+### 45/151 implemented rules (30%)
 
-#### Automated rules: 44/77 (57%)
+#### Automated rules: 45/77 (58%)
```
romulets, line 395: I thing this needs to be generated again

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
```diff
@@ -0,0 +1,31 @@
+package compliance.cis_azure.rules.cis_3_11
+
+import data.compliance.lib.common
+import data.compliance.policy.azure.data_adapter
+import future.keywords.if
+
+finding = result if {
+	# filter
+	data_adapter.is_storage_account
+	_ = data_adapter.resource.extension.blobService
```
romulets, line 10: Why do you do this blank assignment? Is it to prevent the rule being executed if it doesn't have a blobService extension?

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
```diff
@@ -0,0 +1,31 @@
...
+
+ensure_soft_delete_is_enabled if {
+	is_policy_valid(data_adapter.resource.extension.blobService.deleteRetentionPolicy)
+	data_adapter.resource.extension.blobService.deleteRetentionPolicy.allowPermanentDelete == false
+	is_policy_valid(data_adapter.resource.extension.blobService.containerDeleteRetentionPolicy)
+}
+
+is_policy_valid(policy) if {
+	policy.enabled == true
+	policy.days >= 7
```
romulets, line 29: When reading the policy it seems to me the only check needed is if days isn't empty or null. It's mentioned the 7 to 365 rule, but in the checks just say:
image

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
```diff
@@ -0,0 +1,31 @@
...
+		{"Resource": data_adapter.resource},
+	)
+}
+
+default ensure_soft_delete_is_enabled = false
+
+ensure_soft_delete_is_enabled if {
+	is_policy_valid(data_adapter.resource.extension.blobService.deleteRetentionPolicy)
+	data_adapter.resource.extension.blobService.deleteRetentionPolicy.allowPermanentDelete == false
+	is_policy_valid(data_adapter.resource.extension.blobService.containerDeleteRetentionPolicy)
```
romulets, line 24: Don't this need a check `allowPermanentDelete == false` too? Or whatever the counterpart in code of `enable-container-delete-retention is 'true'` is

romulets (changes requested):

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
```diff
@@ -70,7 +76,54 @@ func (*storageAccountEnricher) addUsedForActivityLogsFlag(assets []inventory.Azu
...
+	}
+
+	perStorageAccountID := map[string]inventory.AzureAsset{}
+	for _, d := range dataProtection {
+		perStorageAccountID[strings.FromMap(d.Extension, "storageAccountId")] = d
+	}
+
+	var errAgg error
+	for i, a := range assets {
+		if a.Type != inventory.StorageAccountAssetType {
```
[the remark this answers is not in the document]
moukoublen: Because this is the part where we mutate (enrich) the initial assets slice.

resources/providers/azurelib/inventory/asset.go
```diff
@@ -47,6 +48,10 @@ const (
 	// Azure Resource Graph table groups
 	AssetGroupResources              = "resources"
 	AssetGroupAuthorizationResources = "authorizationresources"
+
+	// Extension keys
+	ExtensionUsedForActivityLogs = "usedForActivityLogs"
```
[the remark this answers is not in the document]
moukoublen: To avoid repeating the same constant strings throughout the code and the unit tests. 

I could add the network of the other enriched as well, wdyt?

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
```diff
@@ -70,7 +76,54 @@ func (*storageAccountEnricher) addUsedForActivityLogsFlag(assets []inventory.Azu
...
+			continue
+		}
+
+		dataProtectionSettings, exist := perStorageAccountID[a.Id]
+		if !exist {
+			continue
+		}
+		if a.Extension == nil {
```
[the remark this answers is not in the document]
moukoublen: Init it always might be challenging to ensure since we do not have any way to enforce a constructor for `AzureAsset.` 

We could embed this logic inside a getter. 

I am wondering how straightforward that API would be, though. (When do someone calls the getter when the Extension directly. Unles we make the extensions private?)

resources/utils/maps/maps.go
moukoublen: It may not be that simple.  Those Azure SDK Properties structs that are converted using this function have a custom JSON marshaller and also they have inner layers. 

e.g. [BlobServicePropertiesProperties](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage#BlobServicePropertiesProperties).

So we will have to both copy the field names and create all the inner `maps[string]any` as well.  It seemed a bit of dirty code that's why I followed this approach. 

Indeed is not performant there is a better solution to that, my first approach in this PRI was this:

```golang
func asMap(item any) (map[string]any, error) {
    reader, writer := io.Pipe()
    decoder := json.NewDecoder(reader)
    encoder := json.NewEncoder(writer)

    defer func() {
        _ = writer.Close()
        _ = reader.Close()
    }()

    go func() {
        _ = encoder.Encode(item)
    }()

    m := map[string]any{}
    err := decoder.Decode(&m)
    return m, err
}
```
That pipes the bytes from encoder to decoder and has less heap footprint. But then I thought it might be kind of complicated and changed it to marshal/unmarshal which Cloudbeat already uses elsewhere, with the intention to suggest the above as an improvement for the other parts as well.

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
moukoublen: Indeed, I followed the text on the Rationale part. I am not sure what would be the correct thing to do here.

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
moukoublen: Indeed but what I observed from the API calls was that the container delete retention has no `allowPermanentDelete` field. I am not sure why. 

I mostly got data like this:

```json
      "containerDeleteRetentionPolicy": {
        "days": 7,
        "enabled": true
      },
      "deleteRetentionPolicy": {
        "allowPermanentDelete": false,
        "days": 7,
        "enabled": true
      },
```

So I wasn't confident in putting the `allowPermanentDelete` check for container as well.

security-policies/RULES.md
moukoublen: I run both `dev/update_rule_status.py` and `dev/generate_rule_metadata.py` with no result (they also run through pre-commit). 

Am I missing something in the numbers?

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
moukoublen: Yes. If for some reason a storage account does not have the `blobService` extension it should not be evaluated for this rule (to avoid false negatives).

resources/providers/azurelib/inventory/asset.go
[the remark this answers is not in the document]
romulets: I can create a quick PR to add too. Whatever you prefer

resources/utils/maps/maps.go
romulets: Do you need to always get all the properties?

Fine to keep like this too, just wondering if there is a better way to do it.

security-policies/RULES.md
romulets: Ah, I'm missing it. I thought it would be higher. But this is indeed the right one.

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
romulets: Maybe quick discussion with the team? Not sure either. I would implement the steps in the Audit part, to avoid false positives. But also, I have been doing this for 1 month. Better to get a second opinion

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
romulets: Hm. Maybe good to dig a bit deeper to understand this, because the whole point is to see if there is soft delete or not, right?

Atm containers are not checked for soft delete

resources/providers/azurelib/inventory/provider.go
```diff
@@ -60,9 +66,21 @@ func NewProvider(log *logp.Logger, resourceGraphClient *armresourcegraph.Client,
...
+			return readPager(ctx, cl.NewListPager(resourceGroupName, storageAccountName, options))
+		},
 	}
 
-	return &provider{log: log, client: wrapper, diagnosticSettingsCache: cycle.NewCache[[]AzureAsset](log)}
+	return &provider{
+		log:                     log,
+		client:                  wrapper,
+		diagnosticSettingsCache: cycle.NewCache[[]AzureAsset](log),
+		// credentials:             credentials,
```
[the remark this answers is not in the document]
moukoublen: Done

resources/providers/azurelib/inventory/provider.go
```diff
@@ -20,37 +20,43 @@ package inventory
...
 	ListAllAssetTypesByName(ctx context.Context, assetsGroup string, assets []string) ([]AzureAsset, error)
 	ListDiagnosticSettingsAssetTypes(ctx context.Context, cycleMetadata cycle.Metadata, subscriptionIDs []string) ([]AzureAsset, error)
+	ListStorageAccountBlobServices(ctx context.Context, storageAccounts []AzureAsset) ([]AzureAsset, error)
 }
 
 type provider struct {
 	client                  *azureClientWrapper
 	log                     *logp.Logger
 	diagnosticSettingsCache *cycle.Cache[[]AzureAsset]
+	// credentials             azcore.TokenCredential
```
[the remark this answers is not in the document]
moukoublen: Done

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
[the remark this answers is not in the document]
moukoublen: Perhaps something like `AddExtension(string, any)` that adds the specific key to the extension map. Is that what you mean? 

I think this might be nice, I will do it and update the rest of the sets.

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
moukoublen: From what I saw by digging around the settings when the `soft delete` setting is not checked, the `containerDeleteRetentionPolicy` is missing entirely as a field. And when it is checked, then it is present.

It's the other way around. I think the  `allowPermanentDelete` check is wrong. 


Screenshot 2023-12-21 at 12 24 17 PM
It seems this is the checkbox `Enable permanent delete for soft deleted items` that says `Permanent delete enables you to permanently delete a soft-deleted snapshot or blob version before the retention period ends.` which is a different thing than having enabled the soft delete in general. 

That's why all these 3 checkboxes can be enabled simultaneously. 

I will remove the check on `allowPermanentDelete` from rego

resources/providers/azurelib/inventory/asset.go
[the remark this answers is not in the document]
moukoublen: No worries I will do it.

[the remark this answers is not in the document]
moukoublen: Done.

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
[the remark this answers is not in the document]
moukoublen: Done.

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
```diff
@@ -0,0 +1,30 @@
...
+	# set result
+	result := common.generate_result_without_expected(
+		common.calculate_result(ensure_soft_delete_is_enabled),
+		{"Resource": data_adapter.resource},
+	)
+}
+
+default ensure_soft_delete_is_enabled = false
+
+ensure_soft_delete_is_enabled if {
```
orestisfl, line 21:
```suggestion
soft_delete_is_enabled if {
```

nit: `ensure` implies taking an action, not just checking

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
orestisfl: Especially the less than 365 days part seems fishy

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
moukoublen: Done

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
moukoublen: So, you also suggest following the steps as described in the audit. Ok, I will update the rule check then.

moukoublen: Replaced it with `days > 0`

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
```diff
@@ -0,0 +1,29 @@
...
+default soft_delete_is_enabled = false
+
+soft_delete_is_enabled if {
+	is_policy_valid(data_adapter.resource.extension.blobService.properties.deleteRetentionPolicy)
+	is_policy_valid(data_adapter.resource.extension.blobService.properties.containerDeleteRetentionPolicy)
+}
+
+is_policy_valid(policy) if {
+	policy.enabled == true
+	policy.days > 0
```
romulets, line 28: Just double checking, are you 100% sure this will come as integer?  No possibility of days being a string?

romulets (approved):

security-policies/bundle/compliance/cis_azure/rules/cis_3_11/rule.rego
moukoublen: Yes, it's int in the response object in SDK. ([DeleteRetentionPolicy](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage@v1.5.0#DeleteRetentionPolicy))

PR #1674, opened 2023-12-18
Skip long tests when -short flag is set

### LINKED ISSUE #1140: Skip long tests when -short flag is set
**Motivation**
Go supports using the `-short` flag in `go test`.  It's possible to call `t.Skip()` if `t.Short()` returns `true`.

We can skip long-running tests (e.g. those that call `time.Sleep()` or expect timeouts or have more complicated dependencies) to make our unit tests run near-instantaneously when developing locally.

**Definition of done**
What needs to be completed at the end of this task
- [ ] If a test takes a long time, skip it when running `go test -short`
- [ ] All tests run when `-short` is not provided
- [ ] Add pre-commit hook

**Out of scope**
- Changing the CI in any way

orestisfl:
### Summary of your changes
Adds a pre-commit action for running the tests. By using the timeout flag, we ensure that all new tests that are long will be properly skipped when the short flag is used.

Fixes #1140

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -94,6 +94,18 @@ repos:
         language: system
         files: .*.go
         pass_filenames: false
+      - id: go-test
+        name: Run short unit tests
+        entry: bash -c 'go test -short -timeout 100ms ./...'
+        language: system
+        files: .*.go
+        pass_filenames: false
+      - id: go-mod-tidy
+        name: Run go mod tidy
+        entry: bash -c 'go mod tidy'
+        language: system
+        files: (go.mod|go.sum)
+        pass_filenames: false
 
   - repo: local
     hooks:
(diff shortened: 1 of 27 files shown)


romulets:
What is the criteria of long and short tests? When would execute only short and when all?

.pre-commit-config.yaml
```diff
@@ -94,6 +94,18 @@ repos:
         language: system
         files: .*.go
         pass_filenames: false
+      - id: go-test
```
romulets, line 97: Shouldn't we be running all tests in the pre-commit? I get your argument that if what you changed shouldn't impact longer tests, don't run. But also it seems that we are removing part of the benefit of the tests by not running them consistently. I'm not always fully aware of the impact of my changes

moukoublen (approved):

.pre-commit-config.yaml
orestisfl: The tests are still going to run completely in the CI, they weren't being run previously on pre-commit. Running long tests in pre-commit is going to be too cumbersome and time-consuming for every commit.

romulets: ok!

Issue #1661, opened 2023-12-14
Flaky unit test TestPublisher_HandleEvents/Publish_events_on_interval_reached_2_times


romulets:
**Motivation**
Because there are flakky unit tests, slowing down development.

**Definition of done**
What needs to be completed at the end of this task
- [ ] No flakky tests

**Broken Test**
```
=== Failed
=== FAIL: flavors TestPublisher_HandleEvents/Publish_events_on_interval_reached_2_times (0.05s)
    mock.go:334: 
        
        mock: Unexpected Method Call
        -----------------------------
        
        PublishAll([]beat.Event)
        		0: []beat.Event{beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}, beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}, beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}, beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}, beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}, beat.Event{Timestamp:time.Date(1, time.January, 1, 0, 0, 0, 0, time.UTC), Meta:null, Fields:null, Private:interface {}(nil), TimeSeries:false}}
        
        The closest call I have is: 
        
        PublishAll(mock.argumentMatcher)
        		0: mock.argumentMatcher{fn:reflect.Value{typ_:(*abi.Type)(0x72317c0), ptr:(unsafe.Pointer)(0xc001293510), flag:0x13}}
        
        
[...]

PR #1678, opened 2023-12-18
Add retry mechanism to TestPublisher_HandleEvents

romulets:
### Summary of your changes

TestPublisher_HandleEvents tests the events batching functionality of `flavors.Publisher`. This is a tricky test because it spawns multiple goroutines and want to verify that x events were published in Y period. With litlle external interference the tests work great. But if for any reason multiple go routines run 1ms or 2ms later, or 1 go routine delays 10ms, the test will fail. 

For example, in a 45ms period, it's expected:

```
1 event at 10ms
2 event at 20ms
3 event at 30ms
4 event at 40ms
------- Publish 10 events
5 event at 50ms
6 event at 60ms
7 event at 70ms
8 event at 80ms
------- Publish 26 events
9 event at 90ms
------- Publish 9 events
```

But what happens sometimes is:
```
1 event at 11ms
2 event at 23ms
3 event at 35ms
------- Publish 6 events
4 event at 47ms
5 event at 50ms
6 event at 60ms
7 event at 70ms
8 event at 80ms
------- Publish 26 events
9 event at 90ms
------- Publish 9 events
```

To try to time it correctly agnostically from external OS interferences is hard. We could also think on mocking everything, even the tick. But then, are we really testing the functionality?

I decided to add this Retry mechanism, with a backoff period in between retries. Locally, I have not seen it failed. But the chance still exists, although smaller.

If it happens again, I think we should rethink this test.

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/1661

### DIFF
--- a/flavors/publisher_test.go
+++ b/flavors/publisher_test.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/elastic/beats/v7/libbeat/beat"
+	"github.com/hashicorp/consul/sdk/testutil/retry"
 	"github.com/stretchr/testify/mock"
 	"go.uber.org/goleak"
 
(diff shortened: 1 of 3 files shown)


orestisfl (commented):
Thank you for the analysis @romulets

I am not sure about the approach, this seems like a race issue and we should fix it with native sync mechanisms, not retry logic. A flaky test is not really testing something concrete and fixing that by retrying is not really valuable IMO.

orestisfl:
Reading the analysis again, I see how this is tricky. Maybe we can increase the 10ms step between each write to 100ms so that the test is less likely to fail because of context switches and gc?

romulets:
Jah, I see your point. That is my approach and I also don't like the retry.

I had thought about playing with that, just didn't want to change all tests 😬 

I'll dig deeper and see what would be the impact

PR #1685, opened 2023-12-19
Add rerun workflow

gurevichdmitry:
### Summary of your changes

This PR improves Cloudbeat CI functionality by adding a workflow designed to address flaky job failures. In the event of a failure in the K8s-CI workflow, this new workflow will make one additional attempt to run the failed jobs. If flakiness persists, a notification will be sent to the #csp-qa-alerts channel via Slack.

### Screenshot/Data

Screenshot 2023-12-28 at 17 40 04

Screenshot 2023-12-28 at 17 56 08

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


.github/workflows/rerun-flaky-workflows.yml
```diff
@@ -0,0 +1,35 @@
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          MAX_RUNS=3
+          CURRENT_RUN=${{ github.event.workflow_run.run_attempt }}
+          echo "NOTIFICATION_SLACK=0" >> $GITHUB_ENV
+          if [ $CURRENT_RUN -lt $MAX_RUNS ]; then
+            gh run rerun ${{ github.event.workflow_run.id }} --repo ${{ github.repository }} --failed
+          else
+            echo "Max runs reached, not rerunning"
+            echo "NOTIFICATION_SLACK=1" >> $GITHUB_ENV
```
orestisfl, line 31: Instead of a environmental variable we can just fail the step, correct? And then run the slack notification step on failure.

.github/workflows/rerun-flaky-workflows.yml
```diff
@@ -0,0 +1,113 @@
+name: Rerun Flaky Workflows
+
+on:
+  workflow_run:
+    workflows: ["K8s-CI"]
```
orestisfl, line 5:
```suggestion
    workflows: ["K8s-CI", "CNVM-CI"]
```

See https://github.com/elastic/cloudbeat/pull/1706

.github/workflows/rerun-flaky-workflows.yml
```diff
@@ -0,0 +1,113 @@
...
+          # echo "Workflow Run ID: ${{ github.event.workflow_run.id }}"
+          # echo "Job ID: ${{ github.job }}"
+          # echo "Current run: ${{ github.run_attempt }}"
+          # echo "Author: ${{ github.event.workflow_run.head_commit.author.name }}"
+          # echo "Actor: ${{ github.actor }}"
+      - name: Rerun failed jobs in the current workflow
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          MAX_RUNS=2
```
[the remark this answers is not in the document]
gurevichdmitry: On one hand, when a specific job fails, it is better to rerun it several times. On the other hand, if there is a common problem causing all jobs to fail, rerunning them all multiple times could result in unnecessary redundancy. In the case of k8s, where 51 jobs run in parallel, they would be executed twice during a rerun. I think it's better to start with one rerun and then increase it if necessary.

PR #1686, opened 2023-12-19
updatecli: Update go mod dependencies

orestisfl:
### Summary of your changes
Since dependabot is often encountering issues updating our go.mod, we can work-around this by updating everything with updatecli.

Since not all modules can be upgraded without failures, the script updates modules one-by-one and reverts the changes if build & test fails.

### DIFF
--- a/.github/workflows/updatecli.yml
+++ b/.github/workflows/updatecli.yml
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pipeline-name: [ beats, golang, hermit ]
+        pipeline-name: [ beats, golang, hermit, mods ]
     steps:
       - uses: actions/checkout@v4
       - name: Init Hermit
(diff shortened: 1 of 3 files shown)


.github/workflows/updatecli.yml
```diff
@@ -48,7 +48,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pipeline-name: [ beats, golang, hermit ]
+        pipeline-name: [ beats, golang, hermit, mods ]
```
romulets, line 51: Should we also do it on the backport?

romulets (approved):

.github/workflows/updatecli.yml
orestisfl: I think we should avoid backporting dependency updates to keep released or to-be-released versions static to avoid potential issues.

PR #1697, opened 2023-12-20
update-mods.sh: Break after 5 hours

orestisfl:
### Summary of your changes
On the last run, the job timed-out after 6 hours because of the amount of work it had to do. We can do the updates in 5-hour batches in these cases.

For the $SECONDS variable see https://stackoverflow.com/a/31663949


romulets (approved):
The $SECONDS var is cool!

PR #1722, opened 2023-12-21
Skip go-test from pre-commit in gihub action

moukoublen:
### Summary of your changes

Skip go-test from pre-commit in gihub action since it runs in UintTest workflow.

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/.github/workflows/cloudbeat-ci.yml
+++ b/.github/workflows/cloudbeat-ci.yml
@@ -42,8 +42,9 @@ jobs:
 
       - name: Pre-commit Hooks
         env:
-          # Skipping golangci-lint as it's tested by golangci/golangci-lint-action
-          SKIP: golangci-lint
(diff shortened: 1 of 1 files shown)


orestisfl (commented):
I left that on purpose to make sure the timeout is not broken by someone that is not using the pre-commit actions.

.github/workflows/cloudbeat-ci.yml
```diff
@@ -42,8 +42,9 @@ jobs:
 
       - name: Pre-commit Hooks
         env:
-          # Skipping golangci-lint as it's tested by golangci/golangci-lint-action
-          SKIP: golangci-lint
+          # Skipping golangci-lint as it's tested by golangci/golangci-lint-action.
+          # Skipping go-test as it's tested by UnitTests workflow.
```
orestisfl, line 46: Maybe we can add a comment why we are not skipping `go-test` instead?

PR #1723, opened 2023-12-21
revive: Enable confusing-results and function-result-limit

orestisfl:
### Summary of your changes
confusing-results

Description: Function or methods that return multiple, no named, values of the same type could induce error.

function-result-limit

Description: Functions returning too many results can be hard to understand/use.

Configuration: (int) the maximum allowed return values

### Related Issues
Related: #1669

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -92,15 +92,11 @@ linters-settings:
         arguments: [30]
       - name: confusing-naming
         disabled: true
-      - name: confusing-results
-        disabled: true
       - name: cyclomatic
         arguments: [15]
       - name: exported
         disabled: true
       - name: function-length
-        disabled: true
-      - name: function-result-limit
         disabled: true
       - name: get-return
         disabled: true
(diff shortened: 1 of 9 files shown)


moukoublen:
I don't know if its configurable, but perhaps we can increase the function results limit to 4, since we already have a function that returns 4 values, and omit the disable linter instruction.

orestisfl:
It's configurable but I think 4 results are already a lot and should be an exception rather than the rule. 

```go
type benchInit interface {
	initialize(ctx context.Context, log *logp.Logger, cfg *config.Config, ch chan fetching.ResourceInfo) (registry.Registry, dataprovider.CommonDataProvider, dataprovider.IdProvider, error)
}
```

should probably be refactored to return less results

moukoublen:
Up to you. I tend to prefer less to no `nolint`. Meaning that if we need nolint then we could configure the linter and refactor later and then change the linter settings again. 

But I could totally see the exception using no lint in this case.

moukoublen (approved):

orestisfl:
I don't like nolint comments either but the benefit here is that future code is going to be checked by stricter settings.

PR #1724, opened 2023-12-21
revive: Split rules in sections and add comments

orestisfl:
### Summary of your changes
Explain rationale behind disabled rules

### Related Issues
Related: #1669

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -82,18 +82,19 @@ linters-settings:
   revive:
     enable-all-rules: true
     rules:
-      - name: add-constant
-        disabled: true
-      - name: bare-return
-        disabled: true
-      - name: blank-imports
-        disabled: true
+      # Configuration for enabled rules
       - name: cognitive-complexity
         arguments: [30]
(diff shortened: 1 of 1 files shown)


.golangci.yaml
```diff
@@ -82,18 +82,19 @@ linters-settings:
...
       - name: cognitive-complexity
         arguments: [30]
-      - name: confusing-naming
-        disabled: true
       - name: cyclomatic
         arguments: [15]
+      - name: unhandled-error
+        arguments:
+          - "bytes.Buffer.WriteString" # Always returns nil error
+      # TODO: work on these
```
romulets, line 93: Does this have a task already? Just so we don't forget. And surprise me a bit that the linter doesn't complain about todos hehe

romulets (approved):

.golangci.yaml
orestisfl: https://github.com/elastic/cloudbeat/issues/1669 tracks the progress on the rules

PR #1732, opened 2023-12-24
feat: update beats pkg

olegsu:
### Summary of your changes
Update beats pkg to 8.11.3 ([bfdd9fb0a3c4](https://github.com/elastic/beats/releases/tag/v8.11.3))
https://www.elastic.co/guide/en/beats/libbeat/8.11/release-notes-8.11.3.html

### DIFF
--- a/cloudbeat.reference.yml
+++ b/cloudbeat.reference.yml
@@ -44,8 +44,6 @@ cloudbeat:
 #timestamp.precision: millisecond
 
 # Internal queue configuration for buffering events to be published.
-# Queue settings may be overridden by performance presets in the
-# Elasticsearch output. To configure them manually use "preset: custom".
 #queue:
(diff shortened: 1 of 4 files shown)


amirbenun (approved):

olegsu:
beats has hardcoded v7 in the module name, therefore we cannot `go get ...@v8`. To workaround it, we use the commit `go get ...@bfdd9fb0a3c4`
I think that when the dependency scanner scan the changes in `go.mod` it does not see the patched version and raises an error. 

This version is the one with the fix, @oren-zohar wdyt we should do? force push?

orestisfl (changes requested):
beats is updated automatically by updatecli, in fact this is a downgrade in version. I propose to close this PR.

PR #1742, opened 2023-12-26
[Create Environment] Adding ec module

gurevichdmitry:
### Summary of your changes

This PR introduces a new internal EC module that deploys the basic ELK stack. Additionally, the ability to override the image has been implemented as part of this change.

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/.github/workflows/test-environment.yml
+++ b/.github/workflows/test-environment.yml
@@ -402,6 +402,6 @@ jobs:
           poetry run pytest -m "sanity" --alluredir=./allure/results/ --clean-alluredir --maxfail=4
 
       - name: Cleanup Environment
-        if: ${{ inputs.cleanup-env == true }}
+        if: ${{ inputs.cleanup-env == true || inputs.cleanup-env == 'true' }}
         run: |
           just delete-cloud-env ${{ env.DEPLOYMENT_NAME }} '' "false"
--- a/deploy/cloud/modules/ec/main.tf
+++ b/deploy/cloud/modules/ec/main.tf
@@ -1,0 +1,54 @@
+locals {
+  version = var.stack_version
+  region                           = var.region
+  name_prefix                      = var.deployment_name_prefix
+  deployment_template              = var.deployment_template
+  es_docker_image                  = lookup(var.docker_image, "elasticsearch", "")
+  es_docker_image_tag_override     = lookup(var.docker_image_tag_override, "elasticsearch", "")
+  kibana_docker_image              = lookup(var.docker_image, "kibana", "")
+  kibana_docker_image_tag_override = lookup(var.docker_image_tag_override, "kibana", "")
+  apm_docker_image                 = lookup(var.docker_image, "apm", "")
+  apm_docker_image_tag_override    = lookup(var.docker_image_tag_override, "apm", "")
+}
+
+data "ec_stack" "deployment_version" {
+  version_regex = local.version
+  region        = local.region
+}
+
+resource "ec_deployment" "deployment" {
+  name                   = "${local.name_prefix}-${data.ec_stack.deployment_version.version}"
+  version                = data.ec_stack.deployment_version.version
+  region                 = local.region
+  deployment_template_id = local.deployment_template
+  tags                   = var.tags
+
+  elasticsearch = {
+    autoscale = var.elasticsearch_autoscale
+    strategy  = "rolling_all"
+    config = {
+      docker_image = local.es_docker_image_tag_override != "" ? "${local.es_docker_image}:${local.es_docker_image_tag_override}" : null
+    }
+
+    hot = {
+      autoscaling = {
+        max_size = "128g"
+      }
+      size       = var.elasticsearch_size
+      zone_count = var.elasticsearch_zone_count
+    }
+  }
+
+  kibana = {
+    config = {
+      docker_image = local.kibana_docker_image_tag_override != "" ? "${local.kibana_docker_image}:${local.kibana_docker_image_tag_override}" : null
+    }
+  }
+
+  integrations_server = {
+    config = {
+      docker_image = local.apm_docker_image_tag_override != "" ? "${local.apm_docker_image}:${local.apm_docker_image_tag_override}" : null
+    }
+  }
+
+}
--- a/deploy/cloud/modules/ec/output.tf
+++ b/deploy/cloud/modules/ec/output.tf
@@ -1,0 +1,26 @@
+output "kibana_url" {
+  value       = ec_deployment.deployment.kibana.https_endpoint
+  description = "The secure Kibana URL"
+}
+
+output "elasticsearch_url" {
+  value       = ec_deployment.deployment.elasticsearch.https_endpoint
+  description = "The secure Elasticsearch URL"
+}
+
+output "elasticsearch_username" {
+  value       = ec_deployment.deployment.elasticsearch_username
+  sensitive   = true
+  description = "The Elasticsearch username"
+}
+
+output "elasticsearch_password" {
+  value       = ec_deployment.deployment.elasticsearch_password
+  sensitive   = true
+  description = "The Elasticsearch password"
+}
+
+output "stack_version" {
+  value       = data.ec_stack.deployment_version.version
+  description = "The matching stack pack version from the provided stack_version"
+}
--- a/deploy/cloud/modules/ec/variables.tf
+++ b/deploy/cloud/modules/ec/variables.tf
@@ -1,0 +1,76 @@
+variable "ec_api_key" {
+  type = string
+}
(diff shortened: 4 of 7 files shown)


.github/workflows/test-environment.yml
```diff
@@ -152,6 +152,18 @@ jobs:
...
+      - name: Process BC version
+        id: remove-commit-hash
+        run: |
+          # Extract the stack version
+          stack_version="${{ inputs.elk-stack-version }}"
+
+          # Check if the version contains a commit hash, remove it
+          if [[ $stack_version =~ -[a-f0-9]+ ]]; then
+            cleaned_version=$(echo $stack_version | awk -F"-" '{print $1}')
+            echo "STACK_VERSION=$cleaned_version" >> $GITHUB_ENV
```
amirbenun, line 164: Remove it's declaration on line 115

deploy/cloud/modules/ec/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  version                          = var.stack_version
+  region                           = var.region
+  name_prefix                      = var.deployment_name_prefix
+  deployment_template              = var.deployment_template
+  es_docker_image                  = lookup(var.docker_image, "elasticsearch", "")
+  es_docker_image_tag_override     = lookup(var.docker_image_tag_override, "elasticsearch", "")
+  kibana_docker_image              = lookup(var.docker_image, "kibana", "")
+  kibana_docker_image_tag_override = lookup(var.docker_image_tag_override, "kibana", "")
+  apm_docker_image                 = lookup(var.docker_image, "apm", "")
+  apm_docker_image_tag_override    = lookup(var.docker_image_tag_override, "apm", "")
```
amirbenun, line 11: Do we really have a usecase of overriding these?

deploy/cloud/modules/ec/output.tf
```diff
@@ -0,0 +1,26 @@
...
+  description = "The Elasticsearch username"
+}
+
+output "elasticsearch_password" {
+  value       = ec_deployment.deployment.elasticsearch_password
+  sensitive   = true
+  description = "The Elasticsearch password"
+}
+
+output "stack_version" {
```
amirbenun, line 23: Nice, we should start using it instead of the "requested" version.

deploy/cloud/modules/ec/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+  description = "Optional autoscale the Elasticsearch cluster"
+}
+
+# Docker image overrides
+
+variable "docker_image_tag_override" {
+  default = {
+    "elasticsearch" = "",
+    "kibana"        = "",
+    "apm"           = "",
```
amirbenun, line 62: Do we need those?

deploy/test-environments/main.tf
```diff
@@ -54,25 +56,26 @@ provider "restapi" {
...
 module "ec_deployment" {
-  count  = var.serverless_mode ? 0 : 1
-  source = "github.com/elastic/apm-server/testing/infra/terraform/modules/ec_deployment"
+  count = var.serverless_mode ? 0 : 1
 
+  source        = "../cloud/modules/ec"
+  ec_api_key    = var.ec_api_key
   region        = var.ess_region
-  stack_version = var.stack_version
+  stack_version = local.cleaned_version
```
amirbenun, line 64: So we don't even deploy the requested version? Where do we use the commit sha?

.github/workflows/test-environment.yml
```diff
@@ -152,6 +152,18 @@ jobs:
...
           echo "TF_VAR_ec_api_key=$ec_api_key" >> $GITHUB_ENV
 
+      - name: Process BC version
+        id: remove-commit-hash
+        run: |
+          # Extract the stack version
+          stack_version="${{ inputs.elk-stack-version }}"
+
+          # Check if the version contains a commit hash, remove it
+          if [[ $stack_version =~ -[a-f0-9]+ ]]; then
```
amirbenun, line 162: I see that we also clean that on the terraform? Where do we use the full version?

amirbenun (commented):
Nice!

.github/workflows/test-environment.yml
gurevichdmitry: Updates the declaration to support the working of the entire flow retroactively. The version with the hash is only required for Terraform to pin the version of the stack to be deployed.

deploy/test-environments/main.tf
gurevichdmitry: In the case of a released or snapshot version, `stack_version` will not be modified and remains the same. However, in the case of a BC version, we need to remove the hash from `stack_version`. This is essential because the `ec` provider might fail with an unknown version if the hash is included. Additionally, we need to override `docker_image_tag`. Instead of creating complex logic to handle different scenarios, we always provide `docker_image_override` in all cases to ensure that our deployment deploys the pinned version.

.github/workflows/test-environment.yml
gurevichdmitry: For Terraform, we are utilizing the `TF_VAR_stack_version: ${{ inputs.elk-stack-version }} `environment variable, which is passed as is.

deploy/cloud/modules/ec/variables.tf
amirbenun: please add comment about the override logic and why do we must state an override for BC/

.github/workflows/test-environment.yml
```diff
@@ -152,6 +152,18 @@ jobs:
           echo "::add-mask::$ec_api_key"
           echo "TF_VAR_ec_api_key=$ec_api_key" >> $GITHUB_ENV
 
+      - name: Process BC version
```
amirbenun, line 155: I would consider getting the hash on a different optional parameter, `PIN_VERSION` for example.
This way we will avoid the cleaning code implemented both in bash and terraform, and the flow of using the pinned version will be more straight forward `PIN_VERSION != ""`.

deploy/cloud/modules/ec/variables.tf
```diff
@@ -0,0 +1,76 @@
...
+
+variable "elasticsearch_autoscale" {
+  default     = false
+  type        = bool
+  description = "Optional autoscale the Elasticsearch cluster"
+}
+
+# Docker image overrides
+
+variable "docker_image_tag_override" {
```
amirbenun, line 58: Please add comments on why those are needed especially when deploying BC versions.

amirbenun (approved):
A few more comments to your consideration

deploy/cloud/modules/ec/main.tf
oren-zohar: why do we even need the apm component?

.github/workflows/upgrade-environment.yml
```diff
@@ -52,7 +53,20 @@ jobs:
...
+        run: |
+          # Extract the stack version
+          stack_version="${{ inputs.target-elk-stack-version }}"
+
+          # Check if the version contains a commit hash, remove it
+          if [[ $stack_version =~ -[a-f0-9]+ ]]; then
+            cleaned_version=$(echo $stack_version | awk -F"-" '{print $1}')
+            echo "AGENT_VERSION=$cleaned_version" >> $GITHUB_OUTPUT
+          else
+            echo "AGENT_VERSION=$stack_version" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
gurevichdmitry: The common part is just to retrieve `cleaned_version`. Afterwards, the handling is different. In this action, we are adding `AGENT_VERSION` to `GITHUB_OUTPUT` (different jobs), and in the `test-environments` step, we are updating env vars of the same job.

deploy/cloud/modules/ec/main.tf
gurevichdmitry: The default `manual` deployment on the cloud includes the `Integrations server` component.
We are installing the `Integrations server` component, which was [introduced](https://www.elastic.co/guide/en/cloud/current/ec-release-notes-2022-02-03.html) in version 8.x and [consists](https://www.elastic.co/guide/en/apm/guide/8.11/apm-quick-start.html) of APM and Fleet Server. Fleet is necessary to centrally manage all agents, integrations, and APM is part of Integrations server component.

.github/workflows/test-environment.yml
gurevichdmitry: @amirbenun I have refactored the code to support a separate `pin_version` Terraform variable. This eliminates the need for cleanup version in Terraform code. I've updated the step action responsible for initializing and updating `TF_vars` and provided comments for clarity. Additionally, I have improved the documentation to make it more clear.

Issue #1750, opened 2023-12-28
[CIS AWS] Using trail.Name when should be trail.TrailARN


romulets:
**Motivation**

Regarding the following errors:

> [example](https://qa-cycle-812-oleg-cfi-8-12-0.kb.us-west2.gcp.elastic-cloud.com:9243/app/discover#/doc/logs-*/.ds-logs-elastic_agent.cloudbeat-default-2023.12.24-000001?id=UBv8powByeskzjUx-v3U)
> 
> ```
> failed to get trail status arn:aws:cloudtrail:us-west-2:773216979769:trail/elastic-eng-org-cloudtrail operation error CloudTrail: GetTrailStatus, https response error StatusCode: 400, RequestID: 3febd1a2-5182-4c31-8174-5b3a237a9574, TrailNotFoundException: Unknown trail: arn:aws:cloudtrail:us-west-2:704479110758:trail/elastic-eng-org-cloudtrail for the user: 704479110758
> ```

> [example](https://qa-cycle-812-oleg-cfi-8-12-0.kb.us-west2.gcp.elastic-cloud.com:9243/app/discover#/doc/logs-*/.ds-logs-elastic_agent.cloudbeat-default-2023.12.24-000001?id=VRv8powByeskzjUx-v3U)
> 
> ```
> failed to get trail event selector arn:aws:cloudtrail:us-west-2:773216979769:trail/elastic-eng-org-cloudtrail operation error CloudTrail: GetEventSelectors, https response error StatusCode: 400, RequestID: 914eea74-8199-4782-8c86-a509e7c66d0c, TrailNotFoundException: Unknown trail: arn:aws:cloudtrail:us-west-2:704479110758:trail/elastic-eng-org-cloudtrail for the user: 704479110758
> ```

We are fetching trails with name when it should be ARN.

Example of failed call:

```bash
aws cloudtrail describe-trails --no-cli-pager --region us-east-1 | jq ".trailList.[].Name" | xargs -L1 aws cloudtrail --no-cli-pager get-trail-status --name
```

Example of call working with ARN
```bash
aws cloudtrail describe-trails --no-cli-pager --region us-east-1 | jq ".trailList.[].TrailARN" | | xargs -L1 aws cloudtrail --no-cli-pager get-trail-status --name
```

[...]

PR #1759, opened 2023-12-29
Use Cloud Trail ARN when get trail status and event selectors

romulets:
### Summary of your changes

Use Cloud Trail ARN when get trail status and event selectors instead of Cloud Trail name

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/1750

### DIFF
--- a/resources/providers/awslib/cloudtrail/provider.go
+++ b/resources/providers/awslib/cloudtrail/provider.go
@@ -75,7 +75,7 @@ func (p Provider) getTrailStatus(ctx context.Context, trail types.Trail) (*cloud
 		return nil, err
 	}
 
-	return client.GetTrailStatus(ctx, &cloudtrail.GetTrailStatusInput{Name: trail.Name})
+	return client.GetTrailStatus(ctx, &cloudtrail.GetTrailStatusInput{Name: trail.TrailARN})
 }
 
 func (p Provider) getEventSelectors(ctx context.Context, trail types.Trail) ([]types.EventSelector, error) {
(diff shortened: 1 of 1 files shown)


orestisfl (commented):
Is there any testing we can do? The fact that this does not need any changes in tests means that it might not be well-covered. That said, it's just an API change so perhaps not.

Additionally, have you tested this locally?

For reference, the documentation says:
```
	// Specifies the name or the CloudTrail ARN of the trail for which you are
	// requesting status. To get the status of a shadow trail (a replication of the
	// trail in another Region), you must specify its ARN. The following is the format
	// of a trail ARN. arn:aws:cloudtrail:us-east-2:123456789012:trail/MyTrail
	//
	// This member is required.
```

However, how do "shadow trails" reach this function, since it seems that we are not including them in the request here: https://github.com/elastic/cloudbeat/blob/e6872d79740840ff1f1a733d495cb730535d6e89/resources/providers/awslib/cloudtrail/provider.go#L42-L43

Not against this change as it doesn't seem to harm, just trying to understand.

romulets:
@orestisfl

The reason why I did this was because on this FF (8.12) we found the error that information could not be fetched for some Cloud Trails, and reading the docs I found (as you) that we should be using the ARN instead of the name. Once we specify a `name` instead of ARN, aws sdk converts the name to a ARN, using the configured region and user id. But Trails aren't always created by the same user, and that's why once we want to describe `arn:aws:cloudtrail:us-west-2:773216979769:trail/elastic-eng-org-cloudtrail` making a call with name `elastic-eng-org-cloudtrail` the actual call is made for `arn:aws:cloudtrail:us-west-2:704479110758:trail/elastic-eng-org-cloudtrail`, which is a not existent cloud trail.

I tested it locally and I did think about adding tests, but I'd need to add spies to a mock and catch the parameters - which is fine of course, but not sure if it would benefit.

> However, how do "shadow trails" reach this function, since it seems that we are not including them in the request here:

I don't understand your logic here, if we are not requesting, should be no problem, right?

Anyway, if I understand correct ([based on this](https://docs.aws.amazon.com/awscloudtrail/latest/APIReference/API_DescribeTrails.html#API_DescribeTrails_RequestSyntax)) shadow trails are trails created in another region, therefore the ARN should contain the correct region. I need to verify this, I could not find any docs on that nor we have any shadow trail in our aws account. Working on it!

orestisfl (approved):
Thanks for the explanation!

PR #1761, opened 2024-01-02
remove direct dependency go-git

orouz:
### Summary of your changes

we no longer use `go-git` directly


moukoublen (approved):

oren-zohar:
@orouz it still keeps it there as indirect

PR #1762, opened 2024-01-02
fix opa test coverage cmd

orouz:
[seems like](https://github.com/elastic/cloudbeat/actions/runs/7348683270/job/20007241210#step:4:20) calling `./opa` in `.github/workflows/test-opa-coverage.yml` fails. replaced with direct call to `opa`


moukoublen (approved):

.github/workflows/test-opa-coverage.yml
```diff
@@ -20,7 +20,7 @@ jobs:
 
       - name: OPA test coverage
         run: |
-          coverage=$( ./opa test ./bundle -c | jq '.coverage' )
+          coverage=$(opa test ./bundle -c | jq '.coverage' )
```
oren-zohar, line 23: Can we also trim the precision?
image

Issue #1417, opened 2023-10-16
[CIS Azure] Database Services rules


jeniawhite:
We would like to implement the following rules:
- [x] 4.1.1 Ensure that 'Auditing' is set to 'On'
- [x] 4.1.3 Ensure SQL server's Transparent Data Encryption (TDE) protector is encrypted with Customer-managed key
- [x] 4.1.5 Ensure that 'Data encryption' is set to 'On' on a SQL Database
- [x] 4.1.6 Ensure that 'Auditing' Retention is 'greater than 90 days'
- [x] 4.2.1 Ensure that Microsoft Defender for SQL is set to 'On' for critical SQL Servers
- [ ] 4.2.2 Ensure that Vulnerability Assessment (VA) is enabled on a SQL server by setting a Storage Account
- [ ] 4.2.3 Ensure that Vulnerability Assessment (VA) setting 'Periodic recurring scans' is set to 'on' for each SQL server
- [ ] 4.2.4 Ensure that Vulnerability Assessment (VA) setting 'Send scan reports to' is configured for a SQL server
- [ ] 4.2.5 Ensure that Vulnerability Assessment (VA) setting 'Also send email notifications to admins and subscription owners' is set for each SQL Server
- [x] 4.3.2 Ensure Server Parameter 'log_checkpoints' is set to 'ON' for PostgreSQL Database Server
- [x] 4.3.3 Ensure server parameter 'log_connections' is set to 'ON' for PostgreSQL Database Server
- [x] 4.3.4 Ensure server parameter 'log_disconnections' is set to 'ON' for PostgreSQL Database Server
- [x] 4.3.5 Ensure server parameter 'connection_throttling' is set to 'ON' for PostgreSQL Database Server
- [x] 4.3.6 Ensure Server Parameter 'log_retention_days' is greater than 3 days for PostgreSQL Database Server
- [x] 4.3.7 Ensure 'Allow access to Azure services' for PostgreSQL Database Server is disabled
- [x] 4.3.8 Ensure 'Infrastructure double encryption' for PostgreSQL Database Server is 'Enabled'
- [x] 4.4.2 Ensure 'TLS Version' is set to 'TLSV1.2' for MySQL flexible Database Server

romulets:
**[Regarding rules 4.2.2, 4.2.3, 4.2.4, 4.2.5]**

[...]

PR #1768, opened 2024-01-03
CIS 4.1.3

romulets:
### Summary of your changes

Implement CIS 4.1.3.

The implementation follows the from AzureCLI approach:
image

The implemented approach is strict, were a encryption protection must exist and all returned encryption protection should comply to the rule. I think it's valid because it makes no sense (nor is possible to configure it) multiple transparent data encryptions. But, the API returns a slice of encryption protectors, so from a reading perspective it seems like it's possible to have multiple. In practice I believe we will always have only encryption protector returned.

Example of return from the API:

```json
{
  "value": [
    {
      "kind": "azurekeyvault",
      "properties": {
        "serverKeyName": "transparent-encryption_transparent-encryption-pass_235c66b5a8e1462fa021f9c4276af36f",
        "serverKeyType": "AzureKeyVault",
        "uri": "https://transparent-encryption.vault.azure.net/keys/transparent-encryption-pass/235c66b5a8e1462fa021f9c4276af36f"
      },
      "id": "/subscriptions/ef111ee2-6c89-4b09-92c6-5c2321f888df/resourceGroups/evgb_group/providers/Microsoft.Sql/servers/pass-412/encryptionProtector/current",
      "name": "current",
      "type": "Microsoft.Sql/servers/encryptionProtector"
    }
  ]
}
```

### Testing

[SQL Server pass-412](https://portal.azure.com/#@elastic365.onmicrosoft.com/resource/subscriptions/ef111ee2-6c89-4b09-92c6-5c2321f888df/resourceGroups/evgb_group/providers/Microsoft.Sql/servers/pass-412/overview) has been configured with proper transparent encryption and passes

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/1417

### Checklist

#### Introducing a new rule?

- [x] Add relevant unit tests


romulets:
Beaware: there is a infosec rule about keyvault changes. If you change a key, you might get pinged by infosec. You just need to tell them it was on purpose :)

resources/providers/azurelib/inventory/sql_server.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return assets, nil
+}
+
+func convertEncryptionProtector(ep *armsql.EncryptionProtector, resourceGroup string, subID string) AzureAsset {
+	return AzureAsset{
+		Id:       deref(ep.ID),
+		Name:     deref(ep.Name),
+		Location: deref(ep.Location),
+		Properties: map[string]any{
```
moukoublen, line 61: [Note]
Azure sdk models have custom marshaling containing names and populating optional values. 

Even though double conversion  `struct` -> `json` -> `map[any]` is not great performance-wise, it has the advantage of not having to hardcode the json name in cloudbeat code (and also takes care of omitting optional if this is something we need). 


https://github.com/Azure/azure-sdk-for-go/blob/b7a02cc26d3822ae665c896c758b6f94f5ca94eb/sdk/resourcemanager/sql/armsql/models_serde.go#L4869-L4878

moukoublen (approved):

resources/providers/azurelib/inventory/sql_server.go
romulets: I agree. It's good to make a decision and keep it consistent

resources/providers/azurelib/inventory/sql_server_test.go
```diff
@@ -0,0 +1,210 @@
...
+			} else {
+				require.NoError(t, err)
+			}
+
+			require.ElementsMatch(t, tc.expectedAssets, got)
+		})
+	}
+}
+
+func ref[T any](v T) *T {
```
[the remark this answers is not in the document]
romulets: Will do in the other PR!

Issue #1775, opened 2024-01-05
[CIS Azure] Make Storage Account Enricher best-effort


moukoublen:
**Is your feature request related to a problem? Please describe.**
Based on this comment, we can change the storage account enricher implementation to best-effort in all of its internal functions (rather than the main one) to enrich the assets with whatever possible and not completely abort each sub-enrichment when the first error occurs.

PR #1776, opened 2024-01-05
Make Storage Account Enricher best-effort

moukoublen:
### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/1775

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


resources/fetching/fetchers/azure/assets_enricher_storage_account.go
```diff
@@ -41,28 +41,37 @@ func (e storageAccountEnricher) Enrich(ctx context.Context, cycleMetadata cycle.
...
+		errs = append(errs, fmt.Errorf("storageAccountEnricher: error while getting data protection settings: %w", err))
 	}
 
 	if err := e.addStorageAccountServicesDiagnosticSettings(ctx, storageAccounts, assets); err != nil {
-		return fmt.Errorf("storageAccountEnricher: error while getting services diagnostic settings: %w", err)
+		errs = append(errs, fmt.Errorf("storageAccountEnricher: error while getting services diagnostic settings: %w", err))
+	}
+
+	if len(errs) > 0 {
+		return errors.Join(errs...)
```
romulets, line 63: _[Mostly wondering]_

Nothing wrong with this approach, I just wonder why do this way and not directly `errs = errors.Join(errs, err)` on each error?

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
```diff
@@ -41,28 +41,37 @@ func (e storageAccountEnricher) Enrich(ctx context.Context, cycleMetadata cycle.
...
 	}
 
 	return nil
 }
 
-func (storageAccountEnricher) addUsedForActivityLogsFlag(assets []inventory.AzureAsset, diagSettings []inventory.AzureAsset) {
+func (e storageAccountEnricher) addUsedForActivityLogsFlag(ctx context.Context, cycleMetadata cycle.Metadata, assets []inventory.AzureAsset, subscriptionIDs []string) error {
+	diagSettings, err := e.provider.ListDiagnosticSettingsAssetTypes(ctx, cycleMetadata, subscriptionIDs)
```
romulets, line 70: Why fetch diagnostic settings instead to this level of receiving as part of this PR? You can have a "best effort" on both levels, right?

romulets (approved):
No blocker comment, mostly questions

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
moukoublen: Actually, `errors.Join` does not append to an existing `joinError` if this is the first argument or something. 
It creates a new `joinError` and filters out the `nil`s. 

So basically, the [implementation](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/errors/join.go;l=15) logic of `errors.Join` is to put multiple operations under the function, and it will filter out the nils and return an error **if at least one** error appeared.

e.g. 
```golang
err := errors.Join(
    op1(),
    op2(),
    op3(),
)

// handle if err != nil
```

by doing something like `errAgg = errors.Join(errAgg, err)` what will be created is something like this:

```golang
&joinError{
	errs: []error{
		&joinError{
			errs: []error{
				&joinError{
					errs: []error{
						//error a
					},
				},
				//error b
			},
		},
		//error c
	},
	//error d
}
```
Which will work "as expected" (in terms of `.Error()` and `errors.{Is,As}()` handling (because of recursive logic) but seemed "cleaner" to me to create just a single slice.

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
moukoublen: Because diagnostic settings (global subscription id based ones) are only needed in this enrichment the other ones are different. So it seemed appropriate to remove the call from the root one and enclose it here.

If, later on, more than one sub-enrichment needs the same call, we could extract to the root level and pass it as an argument.

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
romulets: It seems `.Is`,`.As` and `.Error` would all work because it's recursive in case of join [.Is impl](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/errors/wrap.go;l=63;bpv=0), [.As impl](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/errors/wrap.go;l=125;bpv=0)
and joinError [.Error impl](https://cs.opensource.google/go/go/+/refs/tags/go1.21.5:src/errors/join.go;l=46;bpv=0)

romulets: Ah I misread. I do think the code is cleaner the other way. This way the inner structure is cleaner 🤷 

No deal breaker, but I'm still in favor of the  errAgg = errors.Join(errAgg, err) :P

PR #1797, opened 2024-01-10
CIS Azure 4.3.8

romulets:
### Summary of your changes

Implement CIS Azure 4.3.8 Ensure 'Infrastructure double encryption' for PostgreSQL Database Server is 'Enabled'

I could not test this rule passing because it's available only for postgre single servers, and that's not [possible to be created anymore](https://learn.microsoft.com/en-us/azure/postgresql/single-server/whats-happening-to-postgresql-single-server). In 2025 it will be completely phased out.

But is possible to see it failing on the already created servers with Disabled infrastructure double encryption

image

### Related Issues
- Related https://github.com/elastic/cloudbeat/issues/1417

### Checklist

#### Introducing a new rule?

- [x] Add relevant unit tests

### DIFF
--- a/security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/rule.rego
+++ b/security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/rule.rego
@@ -1,0 +1,22 @@
+package compliance.cis_azure.rules.cis_4_3_8
+
+import data.compliance.lib.common
+import data.compliance.policy.azure.data_adapter
+import future.keywords.if
+
+finding = result if {
+	# filter
+	data_adapter.is_postgresql_single_server_db
+
+	# set result
+	result := common.generate_result_without_expected(
+		common.calculate_result(infrastructure_encryption_enabled),
+		{"Resource": data_adapter.resource},
+	)
+}
+
+default infrastructure_encryption_enabled = false
+
+infrastructure_encryption_enabled if {
+	data_adapter.properties.infrastructureEncryption == "Enabled"
+}
--- a/security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/rule.rego
+++ b/security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/rule.rego
@@ -18,5 +18,5 @@ finding = result if {
 default infrastructure_encryption_enabled = false
 
 infrastructure_encryption_enabled if {
-	data_adapter.properties.infrastructureEncryption == "Enabled"
+	lower(data_adapter.properties.infrastructureEncryption) == "enabled"
 }
--- a/security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/test.rego
+++ b/security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/test.rego
@@ -6,13 +6,14 @@ import data.lib.test
 import future.keywords.if
 
 test_violation if {
-	# fail if ssl_enforcement is disabled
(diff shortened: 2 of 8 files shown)


security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/test.rego
```diff
@@ -0,0 +1,32 @@
+package compliance.cis_azure.rules.cis_4_3_8
+
+import data.cis_azure.test_data
+import data.compliance.policy.azure.data_adapter
+import data.lib.test
+import future.keywords.if
+
+test_violation if {
+	# fail if ssl_enforcement is disabled
+	eval_fail with input as test_data.generate_postgresql_server_with_infrastructure_encryption("Disabled")
```
oren-zohar, line 10: is there an option for empty value? what would be the result in that case?

security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/rule.rego
```diff
@@ -0,0 +1,22 @@
...
+	result := common.generate_result_without_expected(
+		common.calculate_result(infrastructure_encryption_enabled),
+		{"Resource": data_adapter.resource},
+	)
+}
+
+default infrastructure_encryption_enabled = false
+
+infrastructure_encryption_enabled if {
+	data_adapter.properties.infrastructureEncryption == "Enabled"
```
oren-zohar, line 21: I see that we couldn't test it since we can't create postgre single servers, but is it possible to verify the `Enabled` will be capitalized? I there any chance it will be lower case in some cases?

security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/test.rego
romulets: An empty value is different than "Enabled", so it would be considered disabled, which is fine

security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/rule.rego
romulets: The capital letter is the pattern from Azure Assets Resource Graph. [You can see a implementation with the properties from ARG here](https://github.com/elastic/cloudbeat/blob/main/security-policies/bundle/compliance/cis_azure/rules/cis_4_3_1/rule.rego).

So it _should_ be with capital. With that said, I agree that it's sensible and if azure changes its pattern we would have false positives. 

Just for safety I will lower case it and check with `enabled` instead of `Enabled`. Sounds good?

security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/test.rego
romulets: Added a test confirming what I explained https://github.com/elastic/cloudbeat/pull/1797/files#diff-98815038831f8bde93f4e3d0ccf4f4d1a57f715feb0273b9f43e1f989628213dR10

security-policies/bundle/compliance/cis_azure/rules/cis_4_3_8/rule.rego
romulets: https://github.com/elastic/cloudbeat/pull/1797/files#diff-b2e6d792dfee3d7ee031452d839f317586707fa0c7c9479ef69189b4340106e9R21 lower cased it and added tests on it

oren-zohar (approved):

PR #1800, opened 2024-01-10
Move strings.Dereference to ptrs.Deref

romulets:
### Summary of your changes

In a previous PR the `ptrs.Deref` was introduced, which is a generic version of the string specific `strings.Dereference`. This PR refactors the usage of strings to the generic ptrs

### Checklist

### DIFF
--- a/resources/providers/awslib/account_provider.go
+++ b/resources/providers/awslib/account_provider.go
@@ -27,7 +27,7 @@ import (
 	"github.com/elastic/elastic-agent-libs/logp"
 
 	"github.com/elastic/cloudbeat/dataprovider/providers/cloud"
-	"github.com/elastic/cloudbeat/resources/utils/ptrs"
+	"github.com/elastic/cloudbeat/resources/utils/pointers"
 )
 
 type AccountProviderAPI interface {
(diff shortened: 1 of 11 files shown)


moukoublen:
Perhaps the package can be named `pointers` (like `strings`, `slices` or `maps`) instead of `ptrs`. 

(Totally optional though)

romulets:
Fair point, done @moukoublen

orestisfl:
Unfortunately, this now loses the typing information as it's not clear immediately that the argument is a string

romulets:
What do you mean with typing information? @orestisfl

moukoublen:
@orestisfl I think an alternative (since its generic) could be to write it like this if you want to:

```golang
s := pointers.Deref([string](ptr)
```

orestisfl:
Maybe just name it `pointers.ToString()`?

orestisfl (approved):

romulets:
@orestisfl and I chatted and there was a miscommunication regarding the behaviour of `pointers.Deref`. He thought it was for string only, while it is for any type. We agreed the code is good enough as it's

Issue #1757, opened 2023-12-28
[CIS Azure] Logging and Monitoring rules


jeniawhite:
We would like to implement the following rules:
- [x] 5.1.4 Ensure the storage account containing the container with activity logs is encrypted with Customer Managed Key
- [ ] 5.1.5 Ensure that logging for Azure Key Vault is 'Enabled'

orouz:
for rule `5.1.5`, this is the expected output from the diagnostic settings of a key vault:
```
{
"category": "AuditEvent",
"enabled": true,
"retentionPolicy": {
  "days": 180,
  "enabled": true
  }
}
```

1. `category` needs to be `AuditEvent` or `categoryGroup` needs to be `audit`
2. the entry must have `enabled` set to `true`
3. retention days is set to `180 days` or as appropriate (?) 

as for 3, besides having "as appropriate" quite ambiguous (does any number of days is deemed appropriate?)
setting the retention policy on the diagnostic setting itself is no longer possible (both from UI and CLI): 
![Screenshot 2024-01-28 at 14 07 15](https://github.com/elastic/cloudbeat/assets/20814186/5442a506-b2d3-4690-99ce-bfca6cbca7ca)

to address this, the new approach is to [define](https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/migrate-to-azure-storage-lifecycle-policy?tabs=cli#migration-procedures) a life cycle management policy on the storage account that gets the diagnostics. 

after [doing so](https://portal.azure.com/#@elastic365.onmicrosoft.com/resource/subscriptions/ef111ee2-6c89-4b09-92c6-5c2321f888df/resourcegroups/orz-rg/providers/Microsoft.Storage/storageAccounts/orzstroage/managementPolicies), i'm still not able to see my key vault diagnostic settings with a retention policy enabled. it does say to wait 48 hours for a policy to be completed, but that time has passed and yet still no retention policy is enabled. 

i'm wondering if we should merge the rule's PR without it or not. (it would be partial if we do) 

@oren-zohar wdyt ?

oren-zohar:
> as for 3, besides having "as appropriate" quite ambiguous (does any number of days is deemed appropriate?)
setting the retention policy on the diagnostic setting itself is no longer possible (both from UI and CLI):

Will this be addressed in the upcoming benchmark version?

> i'm wondering if we should merge the rule's PR without it or not. (it would be partial if we do)

[...]

Issue #1509, opened 2022-06-23
Improve rule's evidence collection


oren-zohar:
**Context**
Currently, every rule is split into three main parts, **_filter_** where we decide if the input resource is relevant for this rule, **_evaluate_** where we evaluate the resource against the rule logic and evidence collection.

**The Problem**
When collecting the evidence, we assume that the resource will contain the relevant evidence fields, but if it doesn't, the evidence will be undefined, and as a result, no finding will be sent. This is not a good outcome, and we should address it. 
Some options:

- Wrap the evidence and make sure it's not undefined, if so, set it to some default value e.g. "Evidence collection failed".
This will be easy to implement and maintain.
- Wrap or find an alternate way of extracting fields, with default values, such as in other programming languages. 

Thoughts? @ofiriro3 @uri-weisman @eyalkraft @jeniawhite

ofiriro3:
I agree 💯 percent.

We have to protect the evaluation result where there is no evidence.

Its a rare case but it can definitely happen.

orouz:
another aspect of evidence collection  is the value we provide to the user. (aside from failing silently when missing, which can be addressed separately, see elastic/csp-security-policies#274 ) i had short discussion with @oren-zohar about this and he suggested we make the `result.evidence` field a `json.path` which would be a link that would lead to the relevant resource data and highlight it. 

currently we sometimes send the actual relevant fields and sometimes we send the full resource, which is less helpful and also duplicated.

kubasobon:
After talking to stakeholders, it is clear that we need to consider two things: rules not producing findings for `undefined` evidence, and the best user-friendly evidence format.

## 1. Safe evidence collection

Regardless of the decision on **2.**, we need to prevent a missing field or a typo from completing rule evaluation and posting a finding. The solution I went with is to wrap Rego's built-in `object.get(object, path, default)` function. It never results in `undefined`, thus satisfying the condition of never interrupting rule evaluation. See the PR:
<https://github.com/elastic/cloudbeat/pull/1827>

It is important to note that there is inconsistency in the use of the `evidence` field across the bundle. Most often (~85% rules), we attach an entire resource. For ~15%, we extract and filter the input data to present reason more accurately. This brings us to part two...

## 2. Evidence presentation and UX

@tinnytintin10 & @smriti0321, I would appreciate your input and opinions on this.

There are two open questions, in my humble opinion:

**Q1**:  What would be the most useful Evidence content from an end-user perspective?

[...]

PR #1801, opened 2024-01-10
CIS Azure 5.1.4 

orouz:
### Summary of your changes

adds CIS Azure 5.1.4: 

> Ensure the storage account containing the container with activity logs is encrypted with Customer Managed Key (Automated)

specifically - 
> ensure `keySource` is set to `Microsoft.Keyvault` and `keyVaultProperties` is not set to `null`

### Screenshot/Data
| passed 	| failed 	|
|-------------------	|------------------------------	|
|  not available yet |Screenshot 2024-01-15 at 15 01 01|

### Related Issues
 - partially closes https://github.com/elastic/cloudbeat/issues/1757

### Checklist

#### Introducing a new rule?

- [x] Add relevant unit tests


resources/providers/azurelib/inventory/storage_provider.go
```diff
@@ -77,9 +86,52 @@ func NewStorageAccountProvider(log *logp.Logger, diagnosticSettingsClient *armmo
...
+		storageAccountsAssets, err := transformStorageAccounts(res, saID)
+		if err != nil {
+			return nil, fmt.Errorf("error while transforming storage accounts: %w", err)
+		}
+		assets = append(assets, storageAccountsAssets...)
+	}
+	return assets, nil
+}
 
+func transformStorageAccounts(accountPages []armstorage.AccountsClientListResponse, storageAccountId string) (_ []AzureAsset, errs error) {
```
romulets, line 105: I see you used name returns as a mean to declare errs with value zero (`nil`).

This saves one line, but not sure if it helps on readability. I would favor by manually init the errs variable manually tbh. I don't remember seeing usage of name returns for the purpose you used here.

security-policies/bundle/compliance/cis_azure/rules/cis_5_1_4/rule.rego
```diff
@@ -0,0 +1,32 @@
...
+	# set result
+	result := common.generate_result_without_expected(
+		common.calculate_result(is_customer_managed_key_encrypted),
+		evidence,
+	)
+}
+
+is_customer_managed_key_encrypted if {
+	data_adapter.resource.extension.storageAccount.properties.encryption.keySource == "Microsoft.KeyVault"
+	common.contains_key(data_adapter.resource.extension.storageAccount.properties.encryption, "keyVaultProperties")
```
romulets, line 21: I think this has the same effect. Good to test though.

```suggestion
	data_adapter.resource.extension.storageAccount.properties.encryption.keyVaultProperties
```

security-policies/bundle/compliance/cis_azure/rules/cis_5_1_4/rule.rego
```diff
@@ -0,0 +1,32 @@
...
+	data_adapter.resource.extension.storageAccount.properties.encryption.keyVaultProperties != null
+} else = false
+
+evidence = {
```
romulets, line 25: Why isn't the evidence the whole storage account? That's how I've commonly saw it being implemented

romulets (commented):
Small comments. Curious to see the test of a storage account passing too :)

security-policies/bundle/compliance/cis_azure/rules/cis_5_1_4/rule.rego
orouz: figured sending the relevant fields for 
- tracking the resource
- reasoning why it failed 

is the better way to go, as the whole resource is already being sent anyway and selecting just those fields makes it easier to tell why the rule failed

i'm open to change this if you think it'd be better

flavors/benchmark/azure.go
```diff
@@ -49,6 +50,7 @@ func (a *Azure) NewBenchmark(ctx context.Context, log *logp.Logger, cfg *config.
 
 	return builder.New(
 		builder.WithBenchmarkDataProvider(bdp),
+		builder.WithManagerTimeout(20*time.Minute),
```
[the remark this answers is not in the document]
orouz: no, this is required and unless i'm missing something, it won't be overwritten. needed to avoid timeouts.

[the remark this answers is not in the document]
romulets: Hmmmmmm maybe it's good to discuss with the rest of the team? I agree we should increase. But this looks like something we should be able to configure. For some accounts we might need more than 20min. For some less

security-policies/bundle/compliance/cis_azure/rules/cis_5_1_4/rule.rego
romulets: Hmmmm just had not seen the evidence being manipulated. Usually I see just the resource being sent. But if you are confident this is the way to go, sounds good to me

kubasobon: Working on #1509, I found that we manipulate the `evidence` field in approximately 15% of cases. IMHO as part of #1509 we will need to agree on a common `evidence` strategy.

resources/fetching/fetchers/azure/assets_enricher_storage_account.go
```diff
@@ -59,13 +59,35 @@ func (e storageAccountEnricher) Enrich(ctx context.Context, cycleMetadata cycle.
 		errs = append(errs, fmt.Errorf("storageAccountEnricher: error while getting services diagnostic settings: %w", err))
 	}
 
+	storageAccountsSubscriptionsIds := lo.Uniq(lo.Map(storageAccounts, func(item inventory.AzureAsset, i int) string {
+		return item.SubscriptionId
+	}))
```
moukoublen, line 64: [Optional small improvement]
we could put something like this 
```golang
subIDs := lo.Keys(subscriptions)
```
at, e.g., line 45 and use the `subIDs` slice both in line 46 and in `e.addStorageAccounts`

orouz: i don't think i can use this as `lo.Keys(subscriptions)` is not the same as the set of storage account subscription ids, which is smaller, as it only consists of the storage account type, not all subscriptions. 

for example, on our dev account the first has a length of 139 and the 2nd has 13

moukoublen (approved):

PR #1807, opened 2024-01-11
CIS Azure 4.3.7

romulets:
### Summary of your changes

Implement CIS Azure 4.3.7 Ensure 'Allow access to Azure services' for PostgreSQL Database Server is disabled

#### Test results with flexible

✅  When the allow access to azure service is _not enabled_ it **passes**
image

✅  When the allow access to azure service is _enabled_ it **fails**
image

On the flexible UI you enabled through: 
image

#### Test results with single

✅  When the allow access to azure service is _not enabled_ it **passes**
image

✅  When the allow access to azure service is _enabled_ it **fails**
image

On the single UI you enable through
image

### Related Issues
- Related https://github.com/elastic/cloudbeat/issues/1417

### Checklist

#### Introducing a new rule?

- [x] Add relevant unit tests

### DIFF
--- a/resources/providers/azurelib/inventory/postgresql_provider.go
+++ b/resources/providers/azurelib/inventory/postgresql_provider.go
@@ -32,13 +32,17 @@ import (
 )
 
 type psqlAzureClientWrapper struct {
-	AssetConfigurations         func(ctx context.Context, subID, resourceGroup, serverName string, clientOptions *arm.ClientOptions, options *armpostgresql.ConfigurationsClientListByServerOptions) ([]armpostgresql.ConfigurationsClientListByServerResponse, error)
-	AssetFlexibleConfigurations func(ctx context.Context, subID, resourceGroup, serverName string, clientOptions *arm.ClientOptions, options *armpostgresqlflexibleservers.ConfigurationsClientListByServerOptions) ([]armpostgresqlflexibleservers.ConfigurationsClientListByServerResponse, error)
+	AssetSingleServerConfigurations   func(ctx context.Context, subID, resourceGroup, serverName string, clientOptions *arm.ClientOptions, options *armpostgresql.ConfigurationsClientListByServerOptions) ([]armpostgresql.ConfigurationsClientListByServerResponse, error)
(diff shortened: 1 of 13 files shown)


moukoublen (approved):

resources/providers/azurelib/inventory/postgresql_provider.go
```diff
@@ -144,3 +164,73 @@ func (p *psqlProvider) ListFlexiblePostgresConfigurations(ctx context.Context, s
...
+			continue
+		}
+
+		assets = append(assets, AzureAsset{
+			Id:       pointers.Deref(fr.ID),
+			Name:     pointers.Deref(fr.Name),
+			Location: assetLocationGlobal,
+			Properties: map[string]any{
+				"name":           pointers.Deref(fr.Name),
+				"startIPAddress": pointers.Deref(fr.Properties.StartIPAddress),
```
moukoublen, line 190: My only concern is that we have a generic firewall rules provider at this level, but we filter only the data with the end goal in mind (the enricher and, thus, the rules this will be used for).

Of course, we could always refactor later, but would it be better if the provider provided the whole asset? 

(not necessary to change just a brainstorming point)

romulets: What data is being filtered?

moukoublen: Yes, sorry, it wasn't clear; by data, I mean the complete information/fields included inside each Azure struct, for example, `FirewallRule`.

It does not transform, for example, all `.Properties` into `AzureAsset` but instead only the ones it will use later in the enricher and rule side.  

So we have a provider that is called Postgre SQL Provider and, for example, a function that is called `ListSinglePostgresFirewallRules`, but actually, it does not provide the complete transformed object of `armpostgresql.FirewallRule` but only a subset of the properties/fields included in it. 

As I said, it's not necessarily something to change here, but perhaps we can reevaluate later - regarding providers and if they should provide the complete set of fields or a sub-set.

moukoublen: Sorry, apparently, I mixed up the structs 😞  (flexible server `ConfigurationProperties` with the code introduced in this PR) and thought that a subset of fields was used. 

False alarm. Let's discuss, away from this PR, the possibility of adding possible missing fields in other structs as well.

PR #1823, opened 2024-01-16
CIS Azure 8.1, 8.2, 8.3, 8.4

moukoublen:
### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


resources/fetching/fetchers/azure/assets_enricher.go
```diff
@@ -36,6 +36,7 @@ func initEnrichers(provider azurelib.ProviderAPI) []AssetsEnricher {
 	enrichers = append(enrichers, vmNetworkSecurityGroupEnricher{})
 	enrichers = append(enrichers, sqlServerEnricher{provider: provider})
 	enrichers = append(enrichers, postgresqlEnricher{provider: provider})
+	enrichers = append(enrichers, keyVaultEnricher{provider: provider})
```
romulets, line 39: Circling back to the discussion we had a few weeks ago, we said that in case we had too many asset enrichers, we would consider creating a package for it, just for directory structure purposes.

What do you think? I'm pro creating a new package.

(If you agree, let's do it in another PR soon)

resources/providers/azurelib/inventory/keyvault_provider.go
```diff
@@ -0,0 +1,190 @@
...
+		Name:           pointers.Deref(key.Name),
+		DisplayName:    "",
+		Location:       pointers.Deref(key.Location),
+		ResourceGroup:  vault.ResourceGroup,
+		SubscriptionId: vault.SubscriptionId,
+		TenantId:       vault.TenantId,
+		Type:           pointers.Deref(key.Type),
+		Properties:     properties,
+		Extension: map[string]any{
+			ExtensionVaultID: vault.Id,
```
romulets, line 120: I find it interesting that this Asset, which is an extension already, has an extension. Why is that?

romulets: And I could not find a usage of the extension key `vaultId`

romulets (approved):
Looks good overall!

resources/providers/azurelib/inventory/keyvault_provider.go
moukoublen: Good catch; this was left there by mistake (the first approach was to group all resources together and then, using this extension field, merge with each key vault, like in the storage account). 
But then changed to provider call per asset approach so this is unnecessary. 
I will remove it.

resources/fetching/fetchers/azure/assets_enricher.go
moukoublen: My concern, in general (when thinking about creating a new package), is whether we need an isolated package or not (no matter the number of files that are in it): functionality-wise, or/and if we need to hide some implementation or/and if the new package offers some kind of specialization (like for example `encoding` and `encoding/json` where the second specializes and depends on the first)

We could create a package `azure/enrichers`, but if the only reason for that is the number of files, I would be a bit skeptical (not against though) . 

Let's decide async for a next possible PR, agreed.

PR #1825, opened 2024-01-17
Activate Azure CI

jeniawhite:
### Summary of your changes
Activating Azure CI workflow job

Fixes:
- https://github.com/elastic/cloudbeat/issues/1829

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


.github/workflows/azure-ci.yml
```diff
@@ -62,13 +64,16 @@ jobs:
         run: |
           ./cloudbeat -c deploy/azure/cloudbeat-azure.yml -d '*' &
 
+      - name: Wait for cloudbeat to send events
+        run: sleep 264
```
kubasobon, line 68: This jumps out at me as a very specific numer. Out of curiosity, is that an off-the-wall value, or does it correspond to some specific timeout?

jeniawhite: Just off-the-wall.
I wanted to give some time for `cloudbeat` to send findings.
We can change it, I will add more tests and see what is the minimal time needed to get findings.

kubasobon: Thank you for clarifying.

.github/workflows/azure-ci.yml
```diff
@@ -49,6 +47,10 @@ jobs:
           stack-version: ${{ env.ELK_VERSION }}
           security-enabled: false
 
+      - name: Elasticsearch disable dynamic mapping for findings
+        run: |
+          curl -s -XPUT http://localhost:9200/_index_template/common_template -H 'Content-Type: application/json' -d '{ "index_patterns": ["logs-cloud_security_posture.findings*"], "priority": 500, "template": { "mappings": { "dynamic": false, "properties": { "resource": { "dynamic": false, "properties": { "type": { "type": "keyword" } } }, "result": { "dynamic": false, "properties": { "evaluation": { "type": "keyword" } } }, "rule": { "dynamic": false, "properties": { "benchmark": { "dynamic": false, "properties": { "id": { "type": "keyword" } } }, "tags": { "type": "keyword" } } }, "agent": { "dynamic": false, "properties": { "id": { "type": "keyword" }, "version": { "type": "keyword" } } }, "@timestamp": { "type": "date" } } } } }'
+
```
gurevichdmitry, line 53: @jeniawhite Why is it necessary to disable dynamic mapping for findings?

gurevichdmitry (approved):
@jeniawhite LGTM, Can we include this workflow in the [list of rerun workflows](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/rerun-flaky-workflows.yml#L5)?

.github/workflows/azure-ci.yml
jeniawhite: If we do not disable dynamic mapping this creates a scenario when we drop events with more than 1000 fields.
In our case since the raw resource data is part of the document, we essentially do not send evaluations for large resources.
This creates flakiness in the test cases.

PR #1827, opened 2024-01-17
Add a wrapper allowing safe evidence collection

kubasobon:
### Summary of your changes

Add a new common function `collect_evidence()` which allows for safe evidence collection, even in cases when some values are `undefined`.

### Screenshot/Data

image

### Related Issues

Related: https://github.com/elastic/cloudbeat/issues/1509

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


security-policies/bundle/compliance/lib/common/common.rego
```diff
@@ -33,6 +34,13 @@ calculate_result(evaluation) = "passed" if {
 	evaluation
 } else = "failed"
 
+# Safely evaluate evidence. In case a key is undefined, it will be defaulted.
+# keypaths is an object defined as {str: array}
+collect_evidence(resource, key_paths) := {key: evidence |
+	some key, path in key_paths
+	evidence := object.get(resource, path, "<undefined>")
```
oren-zohar, line 41: Are we going to replace the `<undefined>` with full resource instead?

orouz: [seems like](https://github.com/elastic/cloudbeat/issues/1509#issuecomment-1909686173) the FE will display a specific message to indicate the evidence collection failed, in which case sending `<undefined>` would be one way to identify this case. if we send the resource (which is already sent at the top level) the FE wouldn't be able to differentiate when evidence was collected or not.

orouz (approved):
LGTM

security-policies/bundle/compliance/lib/common/common.rego
oren-zohar: @orouz Let's follow [product's call](https://github.com/elastic/cloudbeat/issues/1509#issuecomment-1906789385) on this for now

oren-zohar (changes requested):
Based on [product's request](https://github.com/elastic/cloudbeat/issues/1509#issuecomment-1906789385) let's send the full resource as a fallback instead of `"<undefined>"` for now

oren-zohar (approved):

PR #1830, opened 2024-01-18
Refactor extensions to use full asset instead of properties only

romulets:
### Summary of your changes

@moukoublen @original-brownbear @kubasobon and I agreed to add the whole asset to the extensions field whenever it's possible, instead of adding the properties only as it was being done.

The reason to do so is to have a richer evidence field.

Each extension was a commit, might be easier to review the changes by commit.

### Checklist


kubasobon (approved):
🚀 Looks good to me

PR #1831, opened 2024-01-18
k8s fix process name matching

moukoublen:
### Summary of your changes
**The issue**
KSPM produces false negative results when [calico](https://www.tigera.io/project-calico/) CNI is used in the cluster.  

The failing rules examine specific running processes in Kubernetes nodes.

**Note: how cloudbeat fetches processes information about running processes:**
Cloudbeat fetches all running process IDs, and for each one of those, it retrieves the name from [proc status file](https://man7.org/linux/man-pages/man5/proc.5.html) (_check section `/proc/pid/status`_). 
Based on the **name** retrieved from the status file, Cloudbeat filters the processes that evaluates.  The names cloudbeat filters are included in the code, for example, for self-managed k8s clusters are in the file `resources/fetching/preset/k8s_preset.go` in `vanillaRequiredProcesses` map. 

**Root Cause**
When the Calico Kubernetes controller is installed, it runs in the control plane node and executes this process `/usr/bin/kube-controllers` inside a pod. 

But [proc status file](https://man7.org/linux/man-pages/man5/proc.5.html) has a limitation regarding process names, as is mentioned in the section `/proc/pid/status`:

```plain
Name   Command run by this process.  Strings longer than
    TASK_COMM_LEN (16) characters (including the
    terminating null byte) are silently truncated.
```

This means that both calico `kube-controllers` and kubernetes `kube-controller-manager` are silently truncated to `kube-controller` which is one of the names that Cloudbeat filters with.

And as a result, Cloudbeat identifies both processes as `kube-controller` and runs the rules that targets `kube-controller` on both of them.
Screenshot 2024-01-18 at 10 49 19 PM

As we can see in the screenshot (column `process.command_line`), rules `1.3.3` and `1.3.3` evaluated both `kube-controller-manager` (the originally intended) and `/usr/bin/kube-controllers` which is the callico controller (not intended).

**Solution**
To fix this issue the name handling based on which cloudbeat filters out processes changed and `/proc/<pid>/cmdline` is used to identify the full name of the running process. The map `vanillaRequiredProcesses` got updated with the full name as well. 

### Screenshot/Data
Screenshot 2024-01-19 at 12 41 47 AM
Screenshot 2024-01-19 at 12 43 24 AM
Screenshot 2024-01-19 at 12 43 37 AM

### Related Issues

Fixes: https://github.com/elastic/security-team/issues/8351

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/resources/fetching/fetchers/k8s/process_fetcher.go
+++ b/resources/fetching/fetchers/k8s/process_fetcher.go
@@ -142,14 +142,14 @@ func (f *ProcessesFetcher) Fetch(_ context.Context, cycleMetadata cycle.Metadata
 	for _, p := range pids {
 		stat, err := proc.ReadStatFS(f.Fs, p)
 		if err != nil {
-			f.log.Error(err)
+			f.log.Errorf("error while reading /proc/<pid>/stat: %s", err.Error())
 			continue
 		}
 
 		// Get the full command line name and not the /proc/pid/status one which might be silently truncated.
 		cmd, err := proc.ReadCmdLineFS(f.Fs, p)
 		if err != nil {
-			f.log.Error(err)
+			f.log.Error("error while reading /proc/<pid>/cmdline: %s", err.Error())
 			continue
 		}
 		name := extractCommandName(cmd)
--- a/resources/fetching/fetchers/k8s/process_fetcher.go
+++ b/resources/fetching/fetchers/k8s/process_fetcher.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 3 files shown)


resources/fetching/fetchers/k8s/process_fetcher.go
```diff
@@ -141,33 +142,35 @@ func (f *ProcessesFetcher) Fetch(_ context.Context, cycleMetadata cycle.Metadata
 	for _, p := range pids {
 		stat, err := proc.ReadStatFS(f.Fs, p)
 		if err != nil {
-			return err
-		}
-		processConfig, isProcessRequired := f.processes[stat.Name]
-		if !isProcessRequired {
+			f.log.Error(err)
```
orestisfl, line 145: Maybe give a more descriptive log error message?

romulets (approved):

Issue #1473, opened 2023-10-25
[CIS Azure] Fix 6.6 Ensure that Network Watcher is 'Enabled'


orestisfl:
**Describe the bug**
Write a detailed description of the observed behavior using the actions you performed.

**Preconditions**
Environment state and/or platform-specific configurations.

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1.

**Expected behavior**

PR #1844, opened 2024-01-23
CIS 6.6 Fail case location has no Network Watcher

romulets:
### Summary of the changes

This PR implements a new resource batch fetcher that, per subscription, fetch all the available locations and network watchers grouping network watchers per location. 

In order to achieve such a grouping a resource Id non existent in Azure was created, which is held as place holder for the network watchers grouped by location. To see the format, check the attached image in the tests section.

Then, in Rego level, verify if that location has at least one watcher with provisioning state `Succeeded`.

### Reasoning

Before this PR the evaluation of the rule CIS Azure 6.6 was like the following (pseudo code):

```python
watchers = fetch_network_watchers()
for watcher in watchers:
  if watcher.provisioningState == "Succeeded":
    return "Pass"
  else:
    return "Fail"
```

However, the CIS Azure 6.6 requires that **every** location enabled in your account to contain at least one network watcher enabled. And evaluating only the watchers returned by Azure means that we are not checking every available location, only the existent watchers.

image

### Tests
There are 166 findings because I have two subscriptions and for each subscription the command `az account list-locations` is return 83 locations
image

Integration tests:
image

### Related Issues
- Fixes https://github.com/elastic/cloudbeat/issues/1473

### Checklist


resources/fetching/fetcher.go
```diff
@@ -62,25 +62,25 @@ const (
...
+	AzureNetworkWatchersFlowLogType       = "azure-network-watchers-flow-log"
+	AzureNetworkSecurityGroupType         = "azure-network-group"
+	AzurePostgreSQLDBType                 = "azure-postgresql-server-db"
+	AzureFlexiblePostgreSQLDBType         = "azure-flexible-postgresql-server-db"
+	AzureSQLServerType                    = "azure-sql-server"
+	AzureStorageAccountType               = "azure-storage-account"
+	AzureVMType                           = "azure-vm"
+	AzureVaultType                        = "azure-vault"
+	AzureWebSiteType                      = "azure-web-site"
+	AzureBatchedNetworkWatchersByLocation = "azure-batched-network-watchers-by-location"
```
orestisfl, line 83: nit

```suggestion
	AzureActivityLogAlertType             = "azure-activity-log-alert"
	AzureBastionType                      = "azure-bastion"
	AzureBatchedNetworkWatchersByLocation = "azure-batched-network-watchers-by-location"
	AzureClassicStorageAccountType        = "azure-classic-storage-account"
	AzureDiagnosticSettingsType           = "azure-diagnostic-settings"
	AzureDiskType                         = "azure-disk"
	AzureDocumentDBDatabaseAccountType    = "azure-document-db-database-account"
	AzureFlexibleMySQLDBType              = "azure-flexible-mysql-server-db"
	AzureFlexiblePostgreSQLDBType         = "azure-flexible-postgresql-server-db"
	AzureInsightsComponentType            = "azure-insights-component"
	AzureMySQLDBType                      = "azure-mysql-server-db"
	AzureNetworkSecurityGroupType         = "azure-network-group"
	AzureNetworkWatchersFlowLogType       = "azure-network-watchers-flow-log"
	AzurePostgreSQLDBType                 = "azure-postgresql-server-db"
	AzureSQLServerType                    = "azure-sql-server"
	AzureStorageAccountType               = "azure-storage-account"
	AzureVMType                           = "azure-vm"
	AzureVaultType                        = "azure-vault"
	AzureWebSiteType                      = "azure-web-site"
```

resources/providers/azurelib/inventory/susbcriptions_provider.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/elastic/cloudbeat/resources/utils/pointers"
+)
+
+type subscriptionAzureClientWrapper struct {
+	AssetLocations func(ctx context.Context, subID string, clientOptions *arm.ClientOptions, options *armsubscriptions.ClientListLocationsOptions) ([]armsubscriptions.ClientListLocationsResponse, error)
+}
+
+type SubscriptionProviderAPI interface {
+	ListLocations(ctx context.Context, subID string) ([]AzureAsset, error)
+}
```
orestisfl, line 38: Regarding the name: Isn't this a "location provider"? it doesn't provide any subscriptions. The subscription part is only relevant because the implementation (azure) returns a list of locations under the subscription API for some reason.

security-policies/bundle/compliance/policy/azure/data_adapter.rego
```diff
@@ -105,8 +105,8 @@ is_network_watchers_flow_log if {
 	input.subType == "azure-network-watchers-flow-log"
 }
 
-is_network_watcher if {
-	input.subType == "azure-network-watcher"
+is_batched_network_watcher_by_location if {
+	input.subType == "azure-batched-network-watchers-by-location"
 }
```
orestisfl, line 110: I suggest to revert this as we typically don't follow this naming scheme for sub-types AFAICT. E.g. `azure-bastion` is a batched type but it doesn't have such a long name. We could make a decision to rename those but I think we should be consistent for now.

romulets: But isn't just a network watcher either. I find the `azure-bastion` without any hint that it's a special type of asset misleading

resources/providers/azurelib/inventory/susbcriptions_provider.go
romulets: Hmmmmm, I followed the pattern I was following on the other providers. I named a subscription provider because it uses the subscription resource of azure. 

The same way mysql only return configurations but is called mysql provider. I could rename to locations, but it would narrow the scope of this provider quite a lot and differ from the others

resources/fetching/fetcher.go
romulets: I had not realised it was in alphabetical order hehe

security-policies/bundle/compliance/policy/azure/data_adapter.rego
romulets: I slept over it and I'll change to `azure-network-watcher`.

I still think it would be a good idea to have a way to hint case the resource is not a traditional resource

orestisfl (approved):
:+1:

PR #1855, opened 2024-01-24
Add flag to skip SSL verification and docs on how to run agains local ES

romulets:
### Summary of your changes

Connecting to a local ES instance brought up with elastic-package was failing due to SSL verification. This PR adds a flag to skip the verification when needed.


kubasobon (approved):
🚀

PR #1861, opened 2024-01-25
Add Integration Tests Coverage Metric

romulets:
### Summary of your changes

Based on `./tests/product/tests/data` modules, generate Integration Tests Coverage metadata and update in `./security-policies/RULES.md`

It works only for CSPM rules. There are no automated tests for GCP, therefore only AWS and Azure have coverage higher than 0%.

The coverage percentage is calculated as `(IMPLEMENTED_TESTS / ALL_RULES * 2)`

- ALL_RULES is all the rules in the benchmark (not only the ones we implemented)
- ALL_RULES is multiplied by 2 because we expect two tests per rule, a passing and a failing one
- IMPLEMENTED_TESTS are found based on the exported `test_cases` of each test file inside of `./tests/product/tests/data/provider`

### Screenshot/Data
image

### Related Issues
- Fixes https://github.com/elastic/security-team/issues/6420

### Checklist


gurevichdmitry (approved):
@romulets Great work, LGTM

PR #1868, opened 2024-01-28
Add azure-ci to post-merge

orouz:
running this workflow on `pull_request_target` means that PRs that introduce new azure integration tests won't run until they are merged, which could lead to CI failing in PRs with unrelated, previously introduced tests. 

this PR changes the workflow to run post-merge, which makes it so that every new azure test that is added is checked once it's merged


.github/workflows/azure-ci.yml
```diff
@@ -4,11 +4,10 @@ on:
   # TODO: Remove after testing
   workflow_dispatch:
     inputs:
```
gurevichdmitry, line 6: @orouz, I believe the workflow_dispatch was initially used to test the Azure-CI workflow, and now that the workflow is functioning correctly, can we remove this trigger as well?

orestisfl (approved):

oren-zohar (changes requested):
It should be both post and pre merge, no?

oren-zohar (approved):

Issue #1412, opened 2023-10-15
[CIS Azure] Microsoft Defender rules


jeniawhite:
We would like to implement the following rules:
- [x] 2.1.18 Ensure That 'All users with the following roles' is set to 'Owner'
- [x] 2.1.20 Ensure That 'Notify about alerts with the following severity' is Set to 'High'
- [x] 2.1.15 Ensure that Auto provisioning of 'Log Analytics agent for Azure VMs' is Set to 'On'
- [x] 2.1.19 Ensure 'Additional email addresses' is Configured with a Security Contact Email

romulets:
QA:

✅  2.1.15 Failed with Log Analytics agent for Azure VMs set to `off`
✅  2.1.15 Passed with Log Analytics agent for Azure VMs set to `on`
![image](https://github.com/elastic/cloudbeat/assets/5350001/a94cc483-bf1b-4210-9a4f-b6e373d63149)

✅  2.1.18 Failed with role contributor
✅  2.1.18 Passed with role Owner
![image](https://github.com/elastic/cloudbeat/assets/5350001/d5992f39-993a-4f27-88ba-1bcf00ffa897)


✅  2.1.19 Failed without email
✅  2.1.19 Passed with email
![image](https://github.com/elastic/cloudbeat/assets/5350001/b74f8f31-4d45-48dc-ac35-0b870bc2b7d4)

✅  2.1.20 Failed without notifications
✅  2.1.20 Passed with low level
✅  2.1.20 Passed with high level
![image](https://github.com/elastic/cloudbeat/assets/5350001/937d403f-3d3c-4694-b4e1-4adebd28576b)

PR #1883, opened 2024-01-30
[CIS Azure] 2.1.18

moukoublen:
### Summary of your changes

2.1.18 Ensure That 'All users with the following roles' is set to 'Owner'
### Screenshot/Data

Screenshot 2024-01-30 at 3 06 11 AM

### Related Issues

Related: https://github.com/elastic/cloudbeat/issues/1412

### Checklist

#### Introducing a new rule?

- [x] Add relevant unit tests


kubasobon (approved):

resources/providers/azurelib/inventory/asset.go
```diff
@@ -38,6 +38,7 @@ const (
 	NetworkSecurityGroup               = "microsoft.network/networksecuritygroups"
 	PostgreSQLDBAssetType              = "microsoft.dbforpostgresql/servers"
 	FlexiblePostgreSQLDBAssetType      = "microsoft.dbforpostgresql/flexibleservers"
+	SecurityContacts                   = "microsoft.security/securitycontacts"
```
romulets, line 41: Maybe better stick to the `AssetType` sufix pattern? 

If you change this code, could you also fix the NetworkSecurityGroup one? 

(my mistake 😬 )

resources/fetching/fetchers/azure/security_fetcher.go
```diff
@@ -0,0 +1,91 @@
...
+	var errs []error
+
+	for _, sub := range subscriptions {
+		securityContacts, err := f.provider.ListSecurityContacts(ctx, sub.ShortID)
+		if err != nil {
+			f.log.Errorf("AzureSecurityAssetFetcher.Fetch failed to fetch contacts for subscription %s: %s", sub.ShortID, err.Error())
+			errs = append(errs, err)
+			continue
+		}
+		if securityContacts == nil {
```
romulets, line 64: why check if it's null? As far as I understand the behaviour of empty slice and nil is the same (or very close to each other)

resources/providers/azurelib/inventory/security_provider.go
```diff
@@ -0,0 +1,153 @@
...
+
+	urlPath := fmt.Sprintf(listSecurityContactsURI, url.PathEscape(subID))
+
+	req, err := runtime.NewRequest(ctx, http.MethodGet, runtime.JoinPaths(endpoint, urlPath))
+	if err != nil {
+		return nil, err
+	}
+
+	reqQP := req.Raw().URL.Query()
+	reqQP.Set("api-version", "2020-01-01-preview")
```
romulets, line 84: The whole azure thing with api versions 🙄 

Maybe a good idea to create a constant for it just for easier findability? (totally optional ofc)

resources/providers/azurelib/provider.go
```diff
@@ -61,16 +63,22 @@ func (p *ProviderInitializer) Init(log *logp.Logger, azureConfig auth.AzureFacto
 		return nil, fmt.Errorf("failed to init monitor client: %w", err)
 	}
 
+	genericARMClient, err := arm.NewClient("armsecurity-custom", "v0.0.1", azureConfig.Credentials, nil)
```
romulets, line 66: What does this version stands for?

security-policies/bundle/compliance/cis_azure/rules/cis_2_1_18/rule.rego
```diff
@@ -0,0 +1,28 @@
...
+
+default owner_enabled = false
+
+owner_enabled if {
+	# Ensure at least one Security Contact Settings exists and owner is selected.
+	some security_contact in data_adapter.resource
+
+	security_contact.name == "default"
+	"Owner" in security_contact.properties.notificationsByRole.roles
+	security_contact.properties.notificationsByRole.state == "On"
```
romulets, line 27: A bit of defensive comment:

I've seen azure API give back `on`, `On` and `ON`. Just to prevent silly breaking changes, maybe good if we do case insensitive check?

resources/providers/azurelib/inventory/asset.go
moukoublen: sure

resources/fetching/fetchers/azure/security_fetcher.go
moukoublen: I just wasn't sure what is better to store in elastic, empty array `[]` vs `null`, and chose the empty slice/array.

resources/providers/azurelib/inventory/security_provider.go
moukoublen: This "low level" client should be temporary (til the issue https://github.com/Azure/azure-sdk-for-go/issues/19740 is fixed).  I am not sure we will use these values anywhere else. 

If we will we could extract them, sure.

resources/providers/azurelib/provider.go
moukoublen: Nothing in our case. The generic arm client asks for a module name and a version string. In our case, I put those just to not put empty strings.

security-policies/bundle/compliance/cis_azure/rules/cis_2_1_18/rule.rego
moukoublen: Yes, good thought.  WIll convert both to lower.

PR #1894, opened 2024-02-04
Publishing log with event.action

amirbenun:
### Summary of your changes
Add `event.action` and `event.count` context on publishing logs to allow aggregation on eladsticsearch.

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


flavors/publisher.go
```diff
@@ -88,7 +95,8 @@ func (p *Publisher) publish(events *[]beat.Event) {
 		return
 	}
 
-	p.log.Infof("Publishing %d events to elasticsearch", len(*events))
+	msg := fmt.Sprintf("Publishing %d events to elasticsearch", len(*events))
+	p.log.Infow(msg, ecsEventActionField, ecsEventActionValue, ecsEventCountField, len(*events))
```
orestisfl, line 99: Kinda prefer this syntax, not sure what's considered best:
```suggestion
	p.log.With(ecsEventActionField, ecsEventActionValue, ecsEventCountField, len(*events)).Infof("Publishing %d events to elasticsearch", len(*events))
```

The benefit of this is we avoid a `fmt.Sprintf` call which will run even when the configured log level is higher than the one of this specific line.

internal/flavors/publisher.go
```diff
@@ -19,12 +19,19 @@ package flavors
...
 	"time"
 
 	"github.com/elastic/beats/v7/libbeat/beat"
 	"github.com/elastic/elastic-agent-libs/logp"
 )
 
+const (
+	ecsEventActionField = "event.action"
+	ecsEventActionValue = "publish-events"
+	ecsEventCountField  = "event.count"
```
kubasobon, line 32: Where can I find the ECS Event type definition? I've looked at https://www.elastic.co/guide/en/ecs/current/ecs-event.html, but found no mention of `event.count`.

orestisfl: Good catch, I think as a practice we don't use ECS fields that do not (yet) exist. A trick we did in the past is to capitalize them to avoid future conflicts, for example `event.Count`.

kubasobon: Ack, thanks.

amirbenun: Right, `event.count` is not an ECS field.
The reason we used to capitalise non ECS fields is to avoid committing to a schema before it has been officially declared.
In that case, since I didn't find any "counter" field on ECS, I figured I can use `event.count` without capitalizing, because other uses of the same fields (if any) are most likely treat it as a numeric field as well.

orestisfl (approved):

Issue #664, opened 2023-01-18
Refactor Cloudbeat folder structure


amirbenun:
**Motivation**
Keeping Cloudbeat folder structure more clear and aligned will make it ease our lives as developers. It will prevent disagreements as there is a single standard and hopefully it will translate into the speed of our deliveries.

**Definition of done**

- [ ] Refactor `resources/providers` in a way every specific provider is under a dedicated folder
- [ ] Refactor `resources/providers` to group folders under the same topic (`awslib`, `kubernetes`)
- [ ] Refactor `resources/fetchers` in a way every specific fetcher and factory are under a dedicated folder
- [ ] Refactor `resources/fetchers` to group folders under the same topic (`awslib`, `kubernetes`)
- [ ] Cloudbeat-specific go folders should all reside under a single `internal` folder
  - (beater, config, dataprovider, evaluator, include, launcher, pipeline, processor, resources, transformer, uniqueness)
- [ ] Rename files and packages accordingly to follow [golang best practices
](https://go.dev/blog/package-names)
- [ ] Ensure no link was broken in our docs

**Out of scope**
- Do not try to merge related providers for example `ec2_metadata_provider` and `ec2`

**Related tasks/epics**

orouz:
> * Refactor `resources/providers` in a way every specific provider is under a dedicated folder
> * Refactor `resources/providers` to group folders under the same topic (`awslib`, `kubernetes`)

already done, this is the current structure:
```
resources/providers
├── aws_cis
├── awslib
├── azurelib
└── gcplib
```
> * Refactor `resources/fetchers` in a way every specific fetcher and factory are under a dedicated folder
> * Refactor `resources/fetchers` to group folders under the same topic (`awslib`, `kubernetes`)

already done, this is the current structure:

```
resources/fetching/fetchers
├── aws
├── azure
├── gcp
└── k8s
```

> * [ ]  Cloudbeat-specific go folders should all reside under a single `internal` folder

https://github.com/elastic/cloudbeat/pull/1893

> * [ ]  Rename files and packages accordingly to follow [golang best practices](https://go.dev/blog/package-names)

WIP

> * [ ]  Ensure no link was broken in our docs

WIP

PR #1896, opened 2024-02-04
removed unused files

orouz:
### Summary of your changes

removes unused files/folders/lines

closes https://github.com/elastic/cloudbeat/issues/664


orouz:
we could probably also remove our [docs](https://github.com/elastic/cloudbeat/tree/main/docs) folder and remove the `updatecli` go version update. _doesn't seem_ like we need those files.

Issue #672, opened 2023-01-24
resources with 0 findings


oren-zohar:
### Summary

While setting the log level to `debug` we saw some resources with no findings. We need to investigate why this is happening and address it.

**DOD**
- [ ] Reproduce the issue
- [ ] Identify the cause
- [ ] Suggest solution

uri-weisman:
I've noticed it's happening repeatedly with the `ClusterRole` sub type.

```
[elastic_agent.cloudbeat][debug] {"decision_id":"ba39b3c1-b22b-4abe-9912-9c360fafaf7c","input":{"type":"k8s_object","subType":"ClusterRole","resource":{"kind":"ClusterRole","apiVersion":"rbac.authorization.k8s.io/v1","metadata":{"name":"system:controller:generic-garbage-collector","uid":"3c4a3e18-69a5-49fa-b926-04519d5bb75c","resourceVersion":"123","creationTimestamp":"2023-06-21T08:05:16Z","labels":{"kubernetes.io/bootstrapping":"rbac-defaults"},"annotations":{"rbac.authorization.kubernetes.io/autoupdate":"true"}},"rules":[{"verbs":["delete","get","list","patch","update","watch"],"apiGroups":["*"],"resources":["*"]},{"verbs":["create","patch","update"],"apiGroups":["","events.k8s.io"],"resources":["events"]}]},"benchmark":"cis_eks"},"message":"Decision Log","result":{"findings":[],"metadata":{"opa_version":"0.45.0","policy_version":"1.0.0"},"resource":{"apiVersion":"rbac.authorization.k8s.io/v1","kind":"ClusterRole","metadata":{"annotations":{"rbac.authorization.kubernetes.io/autoupdate":"true"},"creationTimestamp":"2023-06-21T08:05:16Z","labels":{"kubernetes.io/bootstrapping":"rbac-defaults"},"name":"system:controller:generic-garbage-collector","resourceVersion":"123","uid":"3c4a3e18-69a5-49fa-b926-04519d5bb75c"},"rules":[{"apiGroups":["*"],"resources":["*"],"verbs":["delete","get","list","patch","update","watch"]},{"apiGroups":["","events.k8s.io"],"resources":["events"],"verbs":["create","patch","update"]}]}}}
```

orouz:
got another one locally working on #1262:

```
{
  "log.level": "debug",
  "@timestamp": "2023-08-29T15:00:58.845+0300",
  "log.logger": "posture",
[...]

PR #1898, opened 2024-02-04
add logs for non evaluated resources

orouz:
### Summary of your changes

logs whenever OPA evaluates a resource and returns no findings

### Related Issues
 - closes https://github.com/elastic/cloudbeat/issues/672


internal/evaluator/opa.go
```diff
@@ -133,7 +133,12 @@ func (o *OpaEvaluator) Eval(ctx context.Context, resourceInfo fetching.ResourceI
 		return EventData{}, fmt.Errorf("error decoding findings: %v", err)
 	}
 
-	o.log.Debugf("Created %d findings for input: %v", len(ruleResults.Findings), fetcherResult)
+	if len(ruleResults.Findings) == 0 {
```
orestisfl, line 136: What's the benefit of adding an `if` here? Isn't the `Created %d findings for input` always going to be logged?

orouz: this is a semantic change to avoid logging `Created 0 findings for input` when we don't create any findings. no actual benefit other than that tbh.

orouz: just to clarify (unless i'm missing smth), there `Created %d findings...` is inside an `else` block, so it won't always be logged

orestisfl: I guess I don't understand the benefit of having a different log line for zero findings? You can always search for `Created 0 findings for input` and that would be the equivalent of `Evaluated resource has no findings`. Without the change, it's more obvious that `Created 7 findings for input` and `Created 0 findings for input` come from the same place.

Issue #1508, opened 2022-08-24
Re-evaluate EKS rule 5.4.5 - Encrypt traffic to HTTPS load balancers with TLS certificates (Manual)


ofiriro3:
**Motivation**
We have recently received feedback from @jamesspi, that he suspects that the evaluation of EKS rule `5.4.5` is not compatible with the rule description. 
We would like to reevaluate this rule and see if the rule is correlating with the rule description or not.

**In-depth**  
Taken from the EKS official benchmark page:
> Rule Encrypt traffic to HTTPS load balancers with TLS certificates (Manual)

As of today, when evaluating this rule as the following suggests:
For each of the ELB listeners do:
a. Check if the SSL certificate id has a value - if it doesn't exist, exit and set evaluation to `Fail`
b. Check that the `InstanceProtocol` is `HTTPS`, if it isn't, exit and set the evaluation to `fail`
c. Set the evaluation to pass.

We would like to reconsider what needs to be checked: the `InstanceProtocol` or the `Protocol`.

**Useful links**
- The Rego rule - https://github.com/elastic/csp-security-policies/blob/main/bundle/compliance/policy/aws_elb/ensure_certificates.rego
- [Load balancer API](https://docs.aws.amazon.com/cli/latest/reference/elb/describe-load-balancers.html)

ofiriro3:
Attaching the raw resource that results in a `fail`.

<details>
<summary>Click here to expand 📄</summary>

```json
{
  "agent": {
    "name": "ip-192-168-38-189.us-west-2.compute.internal",
    "id": "2167088d-46a7-48c4-8d22-833ed8bda89e",
    "ephemeral_id": "02e9186a-ae60-4f14-b039-b5439ed368db",
    "type": "cloudbeat",
    "version": "8.4.0"
  },
  "cycle_id": "3b70983e-b253-43aa-96a2-a4e62794d6c9",
  "resource": {
    "sub_type": "aws-elb",
    "name": "abc25cad240dd44219a16e600ab28ac5",
    "raw": {
      "SecurityGroups": [
        "sg-0f9ed8b53694a3945"
      ],
      "Policies": {
        "LBCookieStickinessPolicies": [],
        "OtherPolicies": [
          "ELBSecurityPolicy-2016-08"
        ],
        "AppCookieStickinessPolicies": []
      },
      "Scheme": "internet-facing",
      "CreatedTime": "2022-08-15T19:52:54.59Z",
[...]

PR #1911, opened 2024-02-07
Fix cis eks 5.4.5

moukoublen:
### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/1508

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


oren-zohar (commented):
Can we add an integration test here?

moukoublen:
@oren-zohar this rule has some extra requirements on infra to be tested:
  * An elb with SSL certificate and https `Listen` rules (for pass scenario) and/or plain HTTP for fail scenarions.
  * Something deployed in eks cluster that uses ingress using this elb (e.g. an nginx or something)
  * Deploy elastic agent (cloudbeat) with the integration configuration also having AWS credentials (to get ELB listeners) using Direct access keys or temporary. keys (I used the later) or perhaps shared credentials.

So perhaps, at least at the moment, it is not worth the effort; what do you think?

oren-zohar (approved):

Issue #1506, opened 2023-04-11
compliance.policy.aws_cloudtrail.trail.at_least_one_trail_satisfied: Compare metric filter patterns semantically


orestisfl:
### Is your feature request related to a problem? Please describe

Currently, when the `at_least_one_trail_satisfied` rule is evaluated for Cloudtrails in CIS 4.x section benchmarks, the `metric_filter_patterns` must match exactly one of the trail's patterns. This means that differences that are caused by insignificant whitespace will still fail the evaluation.

### Describe the solution you'd like

Ideally, we need to investigate ways to compare the patterns semantically. That would involve parsing them and comparing the resulting trees.

OPA provides the ability for extensions: https://www.openpolicyagent.org/docs/latest/extensions/

### Describe alternatives you've considered

Completely ignoring whitespace is an option but it can lead to false-negatives when the patters have differences in significant whitespace.
Additionally, other types of differences in patterns that are semantically equivalent would still cause failures, for example `a || b` vs `b || a`.

### Additional context

By applying this diff (just adding space around `||` in the 4.4 rule test):

```diff
diff --git a/bundle/compliance/cis_aws/rules/cis_4_4/test.rego b/bundle/compliance/cis_aws/rules/cis_4_4/test.rego
index 9532e45..a2efce6 100644
--- a/bundle/compliance/cis_aws/rules/cis_4_4/test.rego
+++ b/bundle/compliance/cis_aws/rules/cis_4_4/test.rego
@@ -26,7 +26,7 @@ test_pass {
 			"Status": {"IsLogging": true},
 			"EventSelectors": [{"IncludeManagementEvents": true, "ReadWriteType": "All"}],
 		},
-		"MetricFilters": [{"FilterName": "filter_1", "FilterPattern": "{($.eventName=DeleteGroupPolicy)||($.eventName=DeleteRolePolicy)||($.eventName=DeleteUserPolicy)||($.eventName=PutGroupPolicy)||($.eventName=PutRolePolicy)||($.eventName=PutUserPolicy)||($.eventName=CreatePolicy)||($.eventName=DeletePolicy)||($.eventName=CreatePolicyVersion)||($.eventName=DeletePolicyVersion)||($.eventName=AttachRolePolicy)||($.eventName=DetachRolePolicy)||($.eventName=AttachUserPolicy)||($.eventName=DetachUserPolicy)||($.eventName=AttachGroupPolicy)||($.eventName=DetachGroupPolicy)}"}],
[...]

PR #1914, opened 2024-02-07
Cloud Watch Comparison Algorithm

romulets:
### Summary of your changes

Implement an advanced cloudwatch filter expressions comparison, were the order of the expressions doesn't matter.

First, once filters are fetched, the pattern is parsed to a structured format easy to compare on OPA layer. Then, in the OPA layer, the comparison between all the parsed filters against a desired pattern (manually written in the new format) happens.

For example the expressions are all the same, but in different formats: 

`{ ($.eventName = CreateRoute) || ($.eventName = CreateRouteTable) }`
`{($.eventName=CreateRoute) || ($.eventName=CreateRouteTable)}`
`{ $.eventName = CreateRoute || $.eventName = CreateRouteTable }`
`{ ($.eventName = CreateRouteTable)|| ($.eventName = CreateRoute) }`
`{ (CreateRouteTable = $.eventName)|| ($.eventName = CreateRoute) }`

There were 3 considered approaches:

- ❌ Full parsing and comparison implemented in go, add a go built in function to OPA and use in the OPA layer
  - Implemented, but it turned out to be too disruptive to the workflow. To actually add a new function to OPA Runtime you need to build a new binary which means update CI (easiest) and all the team adapt their local development workflow to use custom binary.
- ❌  Full parsing and comparison implemented in OPA layer
  - Implemented. It go unreadable and too slow.
- ✅ **Parsing in go and comparison in OPA**
  - it's a good middle ground when compared with the other solutions. Performs well, reads fine. The comprise is that OPA layer doesn't know how to parse expression, so instead of comparing strings, now you need to build the expression on the expect parsed format.

Working integration tests:

image

### Related Issues
- Related https://github.com/elastic/cloudbeat/issues/1506

### Checklist

### DIFF
--- a/internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
+++ b/internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
@@ -1,0 +1,275 @@
+// Licensed to Elasticsearch B.V. under one or more contributor
+// license agreements. See the NOTICE file distributed with
+// this work for additional information regarding copyright
+// ownership. Elasticsearch B.V. licenses this file to you under
+// the Apache License, Version 2.0 (the "License"); you may
+// not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package monitoring
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+const maxDepth = 5
+
+type logicalOperator string
+type comparisonOperator string
+
+const (
+	loAnd logicalOperator = "&&"
+	loOr  logicalOperator = "||"
+
+	coEqual     comparisonOperator = "="
+	coNotEqual  comparisonOperator = "!="
+	coNotExists comparisonOperator = "NOT EXISTS"
+)
+
+func listLogicalOperators() []logicalOperator {
+	return []logicalOperator{loAnd, loOr}
+}
+
+func listComparisonOperator() []comparisonOperator {
+	// This order must be kept because we need to check first different and then equals
+	return []comparisonOperator{coNotExists, coNotEqual, coEqual}
+}
+
+// MetricFilterPattern is a union between simpleExpression and complexExpression
+//
+//	because it can be either one or the other
+type MetricFilterPattern struct {
+	simpleExpression  //nolint:unused
+	complexExpression //nolint:unused
+}
+
+type simpleExpression struct {
+	Simple   bool
+	Left     string
+	Operator comparisonOperator
+	Right    string
+}
+
+func newSimpleExpression(left string, op comparisonOperator, right string) MetricFilterPattern {
+	return MetricFilterPattern{
+		simpleExpression: simpleExpression{
+			Simple:   true,
+			Left:     left,
+			Operator: op,
+			Right:    right,
+		},
+	}
+}
+
+type complexExpression struct {
+	Complex     bool
+	Operator    logicalOperator
+	Expressions []MetricFilterPattern
+}
+
+func newComplexExpression(op logicalOperator, exps ...MetricFilterPattern) MetricFilterPattern {
+	return MetricFilterPattern{
+		complexExpression: complexExpression{
+			Complex:     true,
+			Operator:    op,
+			Expressions: exps,
+		},
+	}
+}
+
+// parseFilterPattern receives a string with CloudWatch FilterPattern syntax and parses to an expressions
+// Known limitations:
+//   - It parses only 5 levels deep expressions. You can increase with maxDepth const. But we don't have that use case
+//   - It doesn't properly handle reserved characters (=, !=, NOT EXISTS, &&, ||) inside strings. We don't have that use case yet
+func parseFilterPattern(s string) (MetricFilterPattern, error) {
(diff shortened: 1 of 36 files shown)


internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
```diff
@@ -0,0 +1,275 @@
...
+		}
+
+		tmpString := buf.String()
+		if contains, op := hasSuffixLogicalOp(tmpString); contains { // && or || marks the end of a Simple MetricFilterPattern
+			if logicalOp == "" { // set the Operator of the MetricFilterPattern without overriding it
+				logicalOp = op
+			}
+
+			if logicalOp != op {
+				return MetricFilterPattern{}, errors.New("not supported comparison with alternating logical operators")
```
kubasobon, line 144: Do I understand correctly, that the parser only supports a chain of same boolean operators, so both `x AND y AND z` and `x OR y OR z` are supported, but not any combination like `x AND y OR z`?

kubasobon (approved):
🚀 The unit tests made this PR more understandable. Nice.

internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+}
+
+// parseFilterPattern receives a string with CloudWatch FilterPattern syntax and parses to an expressions
+// Known limitations:
+//   - It parses only 5 levels deep expressions. You can increase with maxDepth const. But we don't have that use case
+//   - It doesn't properly handle reserved characters (=, !=, NOT EXISTS, &&, ||) inside strings. We don't have that use case yet
+func parseFilterPattern(s string) (MetricFilterPattern, error) {
+	// remove trailing spaces and { }
+	cleanS := strings.TrimSpace(strings.TrimRight(strings.TrimLeft(strings.TrimSpace(s), "{"), "}"))
```
oren-zohar, line 97: calling `TrimSpace` twice?

internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
```diff
@@ -0,0 +1,275 @@
...
+	coEqual     comparisonOperator = "="
+	coNotEqual  comparisonOperator = "!="
+	coNotExists comparisonOperator = "NOT EXISTS"
+)
+
+func listLogicalOperators() []logicalOperator {
+	return []logicalOperator{loAnd, loOr}
+}
+
+func listComparisonOperator() []comparisonOperator {
```
oren-zohar, line 44: name is a bit misleading IMO

internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
```diff
@@ -0,0 +1,275 @@
...
+	// This order must be kept because we need to check first different and then equals
+	return []comparisonOperator{coNotExists, coNotEqual, coEqual}
+}
+
+// MetricFilterPattern is a union between simpleExpression and complexExpression
+//
+//	because it can be either one or the other
+type MetricFilterPattern struct {
+	simpleExpression  //nolint:unused
+	complexExpression //nolint:unused
```
oren-zohar, line 54: can you explain why did you disable linting here?

internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
```diff
@@ -0,0 +1,275 @@
...
+
+	subS := s[i+1 : closingParenthesisPos+i] // take what is inside matching parenthesis
+	exp, err := safeParse(subS, depth+1)     // safe sub MetricFilterPattern
+	if err != nil {
+		return MetricFilterPattern{}, -1, err
+	}
+	return exp, closingParenthesisPos, nil
+}
+
+func matchingParenthesisPos(s string) int {
```
oren-zohar, line 200: Please consider adding a comment explaining the logic here

internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
```diff
@@ -0,0 +1,275 @@
...
+func parseSimpleStatement(s string) (MetricFilterPattern, error) {
+	buf := strings.Builder{}
+	buf.Grow(len(s))
+
+	var left string
+	var operator comparisonOperator
+	foundOp := false
+
+	for i, r := range s {
+		if buf.Len() == 0 && (r == ' ' || r == '(') { // ignore trailing spaces and (
```
oren-zohar, line 228: wouldn't an opening parentheses would be unexpected here? I would expect handling it would have been done at a higher level (e.g., in safeParse or resolveParenthesis).

security-policies/bundle/compliance/policy/aws_cloudtrail/pattern.rego
oren-zohar, line 1: please add more comments here to maintain clarity

oren-zohar (changes requested):

internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
romulets: Where is it misleading you too? What did you expect to happen?

internal/resources/providers/aws_cis/monitoring/filter_pattern_parser.go
romulets: I'll update the comment to better reflect.

But this `MetricFilterPattern` struct is never actually accessed inside go layer, only in opa layer.

Ideally we would have an interface that both `simpleExpression` and `complexExpression` implements, but because there is actually no need for a method, just a "type holder" I can't do it in go. The simplest way I found to do it was by using embedded fields, acting as a union of both types.

oren-zohar (approved):
Great job here!

Issue #1175, opened 2023-08-01
CI cache constantly above limit


orestisfl:
**Motivation**
The [CI cache](https://github.com/elastic/cloudbeat/actions/caches) is constantly hitting the GitHub 10GB limit.
![image](https://github.com/elastic/cloudbeat/assets/5778622/9c206d80-dbbc-412d-9b6a-f838616ec8f7)
As far as I am aware, it's not possible to increase this limit.

This likely impacts the performance of our pipelines as caches are not maintained between runs. We should investigate the impact further.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Investigate performance impact of caches becoming unavailable
- [ ] List all caches by size, which ones are the bigger offenders?
- [ ] Investigate if some of the caches do not provide benefits, e.g. they are written but never read or if their impact is minimal
- [ ] Investigate alternative caching locations, e.g. AWS
- [ ] Investigate if some caches can be merged together, e.g. the go cache with the golangci cache
- [ ] Change our pipelines to fix the problem **or** branch of to a different ticket for the actual solution

moukoublen:
Before ci workflows refactor the github cache was used both tools (golang, go mods etc) and even for docker intermediate build cache. 

By multiplying this to every open pr the cache limit was constantly above the limit. 



After the ci workflows refactor, an attempt was made to minimize the cache usage as much as possible, using a single hermit build -> load to cache step for every job while also using the artifacts upload/download functionality for uploading and downloading docker images (as we can in file https://github.com/elastic/cloudbeat/blob/main/.github/actions/docker-images/action.yml )

It is still possible for the cache usage to exceed the limit, but currently, it's not that easy.
<img width="1378" alt="Screenshot 2024-06-05 at 5 26 32 PM" src="https://github.com/elastic/cloudbeat/assets/1380039/cc95ff98-8aad-400b-9d8e-b9bcbd10cea0">


I think we can conclude this task and mark it as closed.

PR #1926, opened 2024-02-13
attempt to fix ga hermit cache

moukoublen:
### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/1175

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/.github/actions/hermit/action.yml
+++ b/.github/actions/hermit/action.yml
@@ -1,0 +1,38 @@
+name: 'Hermit Environment'
+description: 'Activate hermit environment'
+inputs:
+  cache:
+    description: 'Cache hermit files'
+    required: false
+    default: 'true'
+runs:
+  using: composite
+  steps:
+    - if: ${{ inputs.cache == 'true' }}
+      id: hermit-hash
+      shell: bash
+      run: |
+        bin_hashes="${{ hashFiles('./bin/*') }}"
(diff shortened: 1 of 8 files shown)


orestisfl (commented):
What about https://github.com/cashapp/activate-hermit that has a cache option?

moukoublen:
@orestisfl that was my first thought but we need some extra directories in cache that are not included in `https://github.com/cashapp/activate-hermit`. 

We need (so far):
```
          ~/.cache/hermit/pkg
          ./.hermit
          ~/.cache/go-build
          ~/go/pkg/
          ~/.cache/pre-commit
```

This is still draft. I am stiil thinking about the optimal way to separate the caches in order not to get overwritten (because cloudbeat has a lot workflows that run in parallel using the same -or not- cache).

orestisfl:
Great, thanks. How are we going to handle the case where different items are cached depending on the pipeline? For example, opa pipelines will just initialize `opa` while go tests will initialize ~/go as well.

moukoublen:
This is what I meant when I wrote that I am still thinking about the optimal way to separate the caches in order not to get overwritten. Different pipelines run in parallel and might initialize different parts of "cache". (Hermit install at the beginning is the easy part and already implemented. The difficult part are python envs, go pks and pre-commit cache). 

This is still wip

moukoublen:
closed due to this: https://github.com/elastic/cloudbeat/pull/2000

PR #1927, opened 2024-02-13
Fetch AWS Management Account resources only when explicitly selected

kubasobon:
🚧 WIP 🚧

- TODO: replace TODO with description of how role assumption works

### Summary of your changes

- do not attach SecurityAudit policy to `cloudbeat-root`
- force cloudbeat to always assume `cloudbeat-securityaudit` role
- add `iam:List*` capabilities to `cloudbeat-root` policy

### Related Issues

Fixes https://github.com/elastic/security-team/issues/8672

### Checklist
- ~~I have added tests that prove my fix is effective or that my feature works~~

#### Introducing a new rule?

No.

### DIFF
--- a/deploy/cloudformation/elastic-agent-ec2-cspm-organization.yml
+++ b/deploy/cloudformation/elastic-agent-ec2-cspm-organization.yml
@@ -93,8 +93,6 @@ Resources:
                 Action:
                   - sts:AssumeRole
                 Resource: '*'
-      ManagedPolicyArns:
-        - arn:aws:iam::aws:policy/SecurityAudit
 
   # Instance profile to attach to EC2 instance
   ElasticAgentInstanceProfile:
--- a/deploy/cloudformation/elastic-agent-ec2-cspm-organization.yml
+++ b/deploy/cloudformation/elastic-agent-ec2-cspm-organization.yml
@@ -86,6 +86,13 @@ Resources:
             Statement:
               - Effect: Allow
                 Action:
+                  - iam:ListAccountAliases
+                  - iam:ListGroup
+                  - iam:ListRoles
+                  - iam:ListUsers
+                Resource: '*'
+              - Effect: Allow
+                Action:
                   - organizations:List*
                   - organizations:Describe*
                 Resource: '*'
--- a/internal/flavors/benchmark/aws_org.go
+++ b/internal/flavors/benchmark/aws_org.go
@@ -114,17 +114,17 @@ func (a *AWSOrg) getAwsAccounts(ctx context.Context, log *logp.Logger, initialCf
 
 	accounts := make([]preset.AwsAccount, 0, len(accountIdentities))
 	for _, identity := range accountIdentities {
-		var memberCfg awssdk.Config
-		if identity.Account == rootIdentity.Account {
-			memberCfg = rootCfg
-		} else {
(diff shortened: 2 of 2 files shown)


deploy/cloudformation/elastic-agent-ec2-cspm-organization.yml
```diff
@@ -84,6 +84,13 @@ Resources:
           PolicyDocument:
             Version: "2012-10-17"
             Statement:
+              - Effect: Allow
+                Action:
+                  - iam:ListAccountAliases
+                  - iam:ListGroup
+                  - iam:ListRoles
+                  - iam:ListUsers
+                Resource: '*'
```
orestisfl, line 93: There is a significant consideration when updating ClodFormation files:

For existing installations, the Elastic Agent will be updated but the CloudFormation will not. Without a workaround or fallback path, existing multi-account aws installations that are updated from 8.12 to 8.13 will stop working.

internal/flavors/benchmark/aws_org.go
```diff
@@ -114,17 +114,17 @@ func (a *AWSOrg) getAwsAccounts(ctx context.Context, log *logp.Logger, initialCf
...
-				stsClient,
-				rootCfg,
-				fmtIAMRole(identity.Account, memberRole),
-			)
-		}
-
+		// TODO(kuba): Add a comment explaining role assumption:
+		// - we're logged in into the main account, we use "cloudbeat-root" role
+		// - "cloudbeat-root" has limited permissions - account listing and role assumption
+		// - CF StackSets will install "cloudbeat-securityaudit" role where applicable
```
orestisfl, line 120: Have you verified this part? Does it install the `cloudbeat-securityaudit` role in the management account if that's included in the organization unit the user has selected?

kubasobon:
@orestisfl 
> Have you verified this part? Does it install the `cloudbeat-securityaudit` role in the management account if that's included in the organization unit the user has selected?

Good point, the StackSets don't get created in the Management Account, even if it's a part of OU.

> For existing installations, the Elastic Agent will be updated but the CloudFormation will not.

I wasn't aware of this, but it makes sense, since Elastic Agent is deployed manually. This throws a wrench in my plans, since changing a Cloud Formation template won't propagate to older installations.

---

My other idea was adding a toogle to the CloudFormation template, which would allow user to select/deselect the Management Account, but it was based on the same premise - using AWS permissions and changing the role `cloudbeat` uses to fetch data.

All those options are off the table because of what you mentioned. I am left with little room to improve this, since `cloudbeat` is unaware of selected OUs.

What do you think about using @maxcold's idea? Let me rephrase what he suggested:
- Add a UI text input for user to add comma-separated OUs, e.g. in the **Setup Access** section
- Propagate this list to the Cloud Formation template and Cloudbeat agent config

CSPM 'Configure integration' screen

_CSPM "Configure integration" screen_

maxcold:
> What do you think about using @maxcold's idea? 

I'm happy to be involved here in the discussion but I don't remember when I brought up this idea :slightly_smiling_face: . I also wasn't involved in the discussions in the related ticket https://github.com/elastic/security-team/issues/8672 afair. Can you give me a hint on why do you think I have some additional context about it? maybe then I can be useful here

kubasobon:
Will be replaced by another PR with different approach.

Issue #1362, opened 2023-09-21
[KSPM EKS] EKS role is not authorized to perform the iam:ListAccountAliases action


gurevichdmitry:
**Describe the bug**
When deploying EKS KSPM integration with EKS role permissions, Cloudbeat encounters an issue that results in unexpected restarts. In the Cloudbeat log, the following error is recorded:

```
[elastic_agent.cloudbeat][error] Exiting: launcher could not run Beater: could not create beater: failed to initialize AWS config: failed to get AWS identity: failed to get aliases: operation error IAM: ListAccountAliases, https response error StatusCode: 403, RequestID: 9240fae6-dba4-4509-81c3-5e6855fbc069, api error AccessDenied: User: arn:aws:sts::704479110758:assumed-role/cloudbeat-tf-5jA/aws-go-sdk-1695130876926194821 is not authorized to perform: iam:ListAccountAliases on resource: * because no identity-based policy allows the iam:ListAccountAliases action.
```

**Workaround**
The `live-track-eks-policy` policy has been added to the `cloudbeat-tf-5jA` IAM role to grant the necessary permissions for Cloudbeat to function correctly.

**To Reproduce**

This issue has been consistently reproduced in the live track environment. To reproduce the problem, follow these steps:

1. Access the AWS IAM role named `cloudbeat-tf-5jA`.
2. Remove the `live-track-eks-policy` policy from the role.

**Expected behavior**
2. The Cloudbeat is functioning without encountering the IAM error, and there are no restarts performed.

**Desktop (please complete the following information):**
 - Kibana Version: 8.10.1
 - Endpoint Version: 8.10.1

moukoublen:
Verified:
<img width="1349" alt="Screenshot 2024-02-22 at 1 39 11 PM" src="https://github.com/elastic/cloudbeat/assets/1380039/4717d6e5-af8a-43e8-8e93-1eb40a16bd6f">
<img width="1348" alt="Screenshot 2024-02-22 at 1 38 53 PM" src="https://github.com/elastic/cloudbeat/assets/1380039/9dc11a08-6282-429c-bd86-ba4eee7d64a1">
<img width="3322" alt="Screenshot 2024-02-22 at 1 38 42 PM" src="https://github.com/elastic/cloudbeat/assets/1380039/f95dd51c-557b-42ec-bda9-2933bceb2ec3">

PR #1931, opened 2024-02-14
Make AWS Alias Account Optional

romulets:
### Summary of your changes

Currently once `cis_eks` is started cloudbeat tries to retrieve account alias to enrich the events. If the alias can't be retrieved the whole start up of cloudbeat is halted due to the error.

Without granting the policy `iam:ListAccountAliases` the following error is seen:

```
Exiting: launcher could not run Beater: could not create beater: failed to initialize AWS config: failed to get AWS identity: failed to get aliases: operation error IAM: ListAccountAliases, https response error StatusCode: 403, RequestID: 9d9d4844-33e0-4b3e-8583-028151dfa76d, api error AccessDenied: User: arn:aws:sts::704479110758:assumed-role/romulo-eks-test/aws-go-sdk-1707923720522731118 is not authorized to perform: iam:ListAccountAliases on resource: * because no identity-based policy allows the iam:ListAccountAliases action
```

However, the alias is not essential to run the implemented cis eks benchmark rules, it's used only to enrich events metadata. After a discussion with @gurevichdmitry we agreed that a customer should not have to grant `iam:ListAccountAliases` to have cis_eks running.  

This PR makes aws account alias optional, and if it fails to retrieve, don't halt cloudbeat execution, but only log an error.

Import to notice that AccountId is still mandatory (retrieved via `AssumeRoleWithWebIdentity`). Only AccountAlias has been made optional.

Resources where test has happened:
- [EKS Cluster](https://eu-west-1.console.aws.amazon.com/eks/home?region=eu-west-1#/clusters/romulo-eks-test?selectedTab=cluster-overview-tab)
- [IAM Role without `iam:ListAccountAliases` permission](https://us-east-1.console.aws.amazon.com/iam/home?region=eu-west-1#/roles/details/romulo-eks-test?section=permissions)
- [Custom built image with development branch](https://eu-west-1.console.aws.amazon.com/ecr/repositories/public/704479110758/romulo?region=eu-west-1)
- [Kibana link to logs](https://romulo-eks-test-ugb-8-13-0-snapshot.kb.us-west2.gcp.elastic-cloud.com:9243/app/r/s/W3vUQ)

### Screenshot/Data
Clean logs once solution has been applied
image

### Related Issues
- Fixes https://github.com/elastic/cloudbeat/issues/1362


internal/resources/providers/awslib/identity_provider.go
```diff
@@ -44,13 +47,18 @@ func (p IdentityProvider) GetIdentity(ctx context.Context, cfg aws.Config) (*clo
...
 
 	alias, err := p.getAccountAlias(ctx, cfg)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get aliases: %w", err)
+	var aliasPtr *string
+
+	if err == nil {
+		aliasPtr = &alias
+	} else {
+		p.Logger.Errorf("failed to get aliases: %w", err)
```
oren-zohar, line 55: maybe a warning instead? This does not seem severe enough to be classified as an error IMO, especially since we will begin tracking error rates in serverless.

internal/dataprovider/providers/cloud/data_provider.go
```diff
@@ -39,7 +39,7 @@ const (
 type Identity struct {
 	Provider         string
 	Account          string
-	AccountAlias     string
+	AccountAlias     *string
```
orestisfl, line 42: Would there be any downside if we avoided the pointer here? Instead of `nil` just use `""`.

romulets: None that I can think of tbh. Maybe that's my java minded approach here. Absence of a value, to me, is nil. But no harm on keeping it empty, I believe.

internal/resources/providers/awslib/identity_provider.go
romulets: Will do!

orestisfl (approved):

PR #1982, opened 2024-02-22
[Cloud Security] Create Serverless QA Environment 

Omolola-Akinleye:
### Summary of your changes

This PR creates a Serverless QA Environment. In this PR we updated `Mask Sensitive Data` to update the `TF_VAR_ec_api_key`  with  the secrets QA API KEY when `input.ec-url`  is  `https://console.qa.cld.elstc.co`.

Also, we added options 
### Related Issues
https://github.com/elastic/kibana/issues/176748


.github/workflows/test-environment.yml
```diff
@@ -24,15 +32,18 @@ on:
         required: true
         description: "Elastic Cloud deployment region"
         default: "gcp-us-west2"
+        options:
+          - gcp-us-west2
+          - aws-eu-west-1
         type: string
```
gurevichdmitry, line 38: Since the field is of `type: string`, using `options` on this field is obsolete (you still need to fill the text box). To enable the use of selection, it is necessary to update it to `type: choice`. However, I'm not sure which is better. On one hand, since we are using specific options for the region, it might be preferable to provide selection. On the other hand, if we need to extend to other environments, this will require more options. Providing `type: string` at least allows deploying the environment to known regions without updating the workflow.

.github/workflows/test-environment.yml
```diff
@@ -138,17 +157,37 @@ jobs:
...
+
+          qa_ec_url="${{ env.QA_EC_URL }}"
+          ec_url="${{ env.EC_URL }}"
+
+          # When user stores input ec api key then update Elastic Cloud env API KEY
+          if [[ $ec_api_key != '' ]]; then
+            echo "::add-mask::$ec_api_key"
+            echo "TF_VAR_ec_api_key=$ec_api_key" >> $GITHUB_ENV
+          exit 0
+          fi
```
gurevichdmitry, line 183:
```suggestion
          if [[ $ec_api_key != '' ]]; then
            echo "::add-mask::$ec_api_key"
            echo "TF_VAR_ec_api_key=$ec_api_key" >> $GITHUB_ENV
            exit 0
          fi
```

.github/workflows/test-environment.yml
```diff
@@ -229,6 +268,7 @@ jobs:
           terraform apply --auto-approve \
             -var="deployment_name=${{ env.DEPLOYMENT_NAME }}" \
             -var="region=${{ env.AWS_REGION }}" \
+            -var="ec_url=${{ env.TF_VAR_ec_url }}" \
```
gurevichdmitry, line 271: All variables following the `TF_VAR_name` pattern are [environment variables](https://developer.hashicorp.com/terraform/cli/config/environment-variables#tf_var_name) and should not be provided in the command; this one can be removed.

deploy/test-environments/main.tf
```diff
@@ -10,7 +10,7 @@ locals {
     project  = "${var.project}"
     owner    = "${var.owner}"
   }
-  ec_url = "https://cloud.elastic.co"
+  ec_url = "${var.ec_url}"
```
gurevichdmitry, line 13: `var.ec_url` is used in the `restapi` provider and in the `ec_project` module. In both cases, this variable is used directly. So, there is no need to create` local.ec_url`, and it can be removed.

gurevichdmitry (changes requested):
@Omolola-Akinleye and @opauloh great work! I've added some comments that require attention. Additionally, it would be good to update the documentation. [This one](https://github.com/elastic/cloudbeat/blob/main/deploy/test-environments/README.md) is specifically for working with Terraform. Please add information about ec_url and note the supported regions, maybe with a small table per cloud and region. Also, update [this one](https://github.com/elastic/cloudbeat/blob/main/dev-docs/Cloud-Env-Testing.md), which is related to how to run the workflow. There, we also need to update information about the introduced ec_url, including new/updated images.

.github/workflows/test-environment.yml
Omolola-Akinleye: @gurevichdmitry When we were testing the apply terraform deployment we had to update ec_url with the qa cloud url `terraform apply  -target "module.ec_project" -var='ec_url=https://console.qa.cld.elstc.co' -var='ess_region=aws-eu-west-1' -var='serverless_mode=true'`. If we remove the `ec_url` variable, then how can we create deployments in QA cloud environments?

gurevichdmitry: We have multiple options to assign variables to [root module variables](https://developer.hashicorp.com/terraform/language/values/variables#assigning-values-to-root-module-variables). The `-var` `ec_url` (command-line variable) is the same as `TF_VAR_ec_url` (environment variable), which is defined [here](https://github.com/elastic/cloudbeat/pull/1982/files#diff-e92acdebda21024508b591a7dae2cd202f1e96f7741c94ca643a88fbf9905666R133). By assigning beforehand, Terraform will know to use the environment variable for ec_url.

Omolola-Akinleye: I'll remove in that case.

gurevichdmitry:
@Omolola-Akinleye, The last workflow's successful [run](https://github.com/elastic/cloudbeat/actions/runs/8510115446) didn't involve destroying the environment. Can you please ensure that the destruction process is also functioning correctly?

Omolola-Akinleye:
> @Omolola-Akinleye, The last workflow's successful [run](https://github.com/elastic/cloudbeat/actions/runs/8510115446) didn't involve destroying the environment. Can you please ensure that the destruction process is also functioning correctly?

@gurevichdmitry  I [created another workflow](url) and also [deleted the workflow](url).

deploy/test-environments/delete_env.sh
```diff
@@ -17,6 +17,8 @@ AWS_REGION="eu-west-1" # Add your desired default AWS region here
 DELETED_ENVS=()
 FAILED_ENVS=()
 
+: "${TF_VAR_qa_ec_api_key:?Please set TF_VAR_qa_ec_api_key with an Elastic QA Cloud API Key}"
```
gurevichdmitry, line 20: @Omolola-Akinleye I'm unsure if this is the appropriate location to request the `TF_VAR_qa_ec_api_key`. Since I intend to delete a production environment, the `TF_VAR_qa_ec_api_key` isn't necessary. If I define it here, both keys would be required.

gurevichdmitry (changes requested):
@Omolola-Akinleye I've updated the branch and it appears that the CI is failing due to linting issues. Could you please fix them?

Omolola-Akinleye:
Closing PR as result of QA environment being created

Issue #1976, opened 2024-02-21
Refactor and centralize cloudbeat testing workflows


gurevichdmitry:
**Motivation**
Currently, Cloudbeat has several testing workflows, such as GCP-CI, AZURE-CI, K8S-CI, AWS-CI, and CNMV-CI. However, only the K8S-CI and EKS-CI workflows publish results, while others lack this functionality. Additionally, there is redundancy in the cache mechanism used across these workflows. This task aims to refactor and centralize all testing workflows into a single "main" workflow. The primary goals include consolidating testing logic, enabling result reporting for all workflows, eliminating redundancy in caching mechanisms, and leveraging the Allure reporter to combine and publish the results.

**Definition of done**
- [ ] Combine existing testing workflows (GCP-CI, AZURE-CI, K8S-CI, AWS-CI, CNMV-CI) into a single "main" workflow.
- [ ] Ensure that all testing logic, including test execution and result reporting, is centralized within the "main" workflow.
- [ ] Improve caching mechanisms used across different workflows.
- [ ] All testing workflows should report results using the Allure reporter, and the aggregated results should be published and displayed as a comment in the respective pull request.
- [ ] Update documentation to reflect the changes in the testing workflows.
- [ ] Validate that the refactored workflow successfully builds Cloudbeat, executes tests, and reports results for all supported integrations.

**Out of scope**

**Related tasks/epics**
Reference related issues and epics

PR #2000, opened 2024-03-05
GA Workflows to Actions

moukoublen:
### Screenshot/Data

Untitled drawing (1)

### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/1976

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


.github/actions/aws-ci/action.yml
```diff
@@ -0,0 +1,74 @@
...
+
+    - name: Run elasticsearch
+      uses: elastic/elastic-github-actions/elasticsearch@master
+      with:
+        stack-version: ${{ inputs.elk-version }}
+        security-enabled: false
+
+    - name: Elasticsearch disable dynamic mapping for findings
+      shell: bash
+      run: ./.ci/scripts/disable_dynamic_mapping.sh
```
gurevichdmitry, line 32: @moukoublen These actions are repeated in all testing CIs. Perhaps we should consider creating a common action for this as well?

.github/actions/aws-ci/action.yml
```diff
@@ -0,0 +1,74 @@
...
+        AWS_ACCESS_KEY_ID: ${{ inputs.aws-access-key-id }}
+        AWS_SECRET_ACCESS_KEY: ${{ inputs.aws-secret-access-key }}
+        AWS_ACCOUNT_TYPE: ${{ inputs.aws-account-type }}
+      shell: bash
+      run: |
+        ./cloudbeat -c deploy/aws/cloudbeat-aws.yml -d '*' &
+
+    - name: Wait for cloudbeat to send some events
+      shell: bash
+      run: sleep 20
```
romulets, line 48: 20 second seems not a lot (?). Hard to say, right. But I think 20s can easily add to flakyness. 

I also see other actions (e.g. azure-ci) wait a different amount of seconds. Maybe good to have a const or something where we can easily bump all the wait periods? Or at least have all the pipelines with the same await period

.github/actions/aws-ci/action.yml
```diff
@@ -0,0 +1,74 @@
...
+    #   if: always()
+    #   shell: bash
+    #   run: |
+    #     cat logs/cloudbeat*
+    # - name: Upload cloudbeat logs
+    #   if: always()
+    #   uses: actions/upload-artifact@v4
+    #   with:
+    #     name: cloubeat-logs-ci-azure
+    #     path: logs/
```
gurevichdmitry, line 74: Remove? Alternatively, we could add an input debug flag in case we need to debug logs.

.github/actions/aws-ci/action.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: poetry run pytest -k "aws" --alluredir=./allure/results/ --clean-alluredir
+
+    - name: Upload test results
+      if: ${{ success() || failure() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: allure-results-ci-aws
+        path: tests/allure/results/
+
+    # - name: Print cloudbeat logs
```
romulets, line 64: Remove debug comments?

.github/actions/docker-images/action.yml
```diff
@@ -0,0 +1,112 @@
...
+    required: false
+    default: ${{ github.run_id }}
+  elastic-agent-docker-image:
+    description: 'Elastic-Agent docker image'
+    required: false
+    default: "704479110758.dkr.ecr.eu-west-2.amazonaws.com/elastic-agent"
+  elastic-agent-docker-image-tag:
+    description: 'Elastic-Agent docker image tag'
+    required: false
+    default: "8.14.0-SNAPSHOT"
```
romulets, line 20: Do we want this default? We will pass 8.14 in one or two months and forget this

.github/actions/gcp-ci/action.yml
```diff
@@ -0,0 +1,83 @@
...
+    - name: Run cloudbeat in background
+      env:
+        ES_HOST: http://localhost:9200
+        ES_USERNAME: elastic
+        ES_PASSWORD: changeme
+        GCP_PROJECT_ID: ${{ inputs.gcp-project-id }}
+        GCP_ACCOUNT_TYPE: ${{ inputs.gcp-account-type }}
+      shell: bash
+      run: ./cloudbeat -c deploy/gcp/cloudbeat-gcp.yml -d '*' &
+
```
romulets, line 52: Why all the other cis have a wait period and gcp doesn't have?

.github/actions/aws-ci/action.yml
moukoublen: Indeed, I will add a `debug` flag just to be able to get the logs if needed.

.github/actions/aws-ci/action.yml
romulets: I won't comment on the other file's debug comments, but same applies to them

.github/actions/aws-ci/action.yml
moukoublen: Sure, I will create something like `run-elastisearch`

.github/actions/hermit/action.yml
romulets, line 1: How does other actions have hermit enabled? Can't we re-use somehow what other actions already do?

.github/actions/azure-ci/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    #   if: always()
+    #   shell: bash
+    #   run: |
+    #     cat logs/cloudbeat*
+    # - name: Upload cloudbeat logs
+    #   if: always()
+    #   uses: actions/upload-artifact@v4
+    #   with:
+    #     name: cloubeat-logs-ci-azure
+    #     path: logs/
```
gurevichdmitry, line 79: remove?

.github/actions/cnvm-ci/action.yml
```diff
@@ -0,0 +1,81 @@
...
+    #   if: always()
+    #   shell: bash
+    #   run: |
+    #     cat logs/cloudbeat*
+    # - name: Upload cloudbeat logs
+    #   if: always()
+    #   uses: actions/upload-artifact@v4
+    #   with:
+    #     name: cloubeat-logs-ci-cnvm
+    #     path: logs/
```
gurevichdmitry, line 81: remove?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,378 @@
+name: CI
```
romulets, line 1: I find the name "CI" not explicit enough 🤔 

Maybe `Build and Test` or something like that?

.github/actions/docker-images/action.yml
```diff
@@ -0,0 +1,112 @@
...
+        cache-to: type=local,mode=max,dest=${{ inputs.docker-build-cache-folder }}-new/cloudbeat
+        outputs: type=docker,dest=${{ inputs.docker-images-folder }}/cloudbeat.tar
+        no-cache: false
+
+    - if: ${{ inputs.build-docker-images == 'true' }}
+      name: Build elastic-agent
+      uses: docker/build-push-action@v5
+      env:
+        GOOS: linux
+        GOARCH: amd64
```
gurevichdmitry, line 81: It might be problematic to hardcode `GOOS` and `GOARCH` because this composite action may be executed on different GitHub runners. Currently, we are using Ubuntu, which is fine. Perhaps we should set these as input variables with defaults to `linux` and `amd64`?

.github/actions/gcp-ci/action.yml
```diff
@@ -0,0 +1,83 @@
...
+    #   if: always()
+    #   shell: bash
+    #   run: |
+    #     cat logs/cloudbeat*
+    # - name: Upload cloudbeat logs
+    #   if: always()
+    #   uses: actions/upload-artifact@v4
+    #   with:
+    #     name: cloubeat-logs-ci-cnvm
+    #     path: logs/
```
gurevichdmitry, line 83: remove?

.github/actions/k8s-ci/action.yml
```diff
@@ -0,0 +1,70 @@
...
+
+  docker-images-folder:
+    description: 'Docker build cache folder'
+    required: false
+    default: '/tmp/.docker-images'
+runs:
+  using: composite
+  steps:
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v3
```
gurevichdmitry, line 27: Why is this action required?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,378 @@
...
+          kind-config: ${{ matrix.kind-config }}
+          test-target: ${{ matrix.test-target }}
+          values-file: ${{ matrix.values-file }}
+          docker-images-folder: '/tmp/.docker-images'
+
+      # - name: Setup tmate session
+      #   uses: mxschmitt/action-tmate@v3
+      #   if: failure()
+      #   with:
+      #     limit-access-to-actor: true
```
gurevichdmitry, line 293: remove?

gurevichdmitry (commented):
@moukoublen impressive work! This should significantly enhance our CI experience. One more question regarding the `eks-ci`: I know that this is not related to `pre-merge` CI, and we are executing it `post-merge`. However, the workflow itself includes `build` and `test_matrix` jobs, which are essentially the same (except for uploading images to AWS Artifactory) as what we're using there. Do you think it's possible to refactor this workflow in this PR, or would it be better to create a new task to do that?

.github/actions/aws-ci/action.yml
moukoublen: https://github.com/elastic/cloudbeat/blob/6db9c916e2d91b16cf7fc65554284c395e2adceb/.github/workflows/aws-ci.yml#L66

It's `264` in the current workflow, and I tried to reduce it while ensuring that the pytest finds rules results (same goes to other runs as well)

.github/actions/docker-images/action.yml
moukoublen: I am not sure I understood if you suggest using a different tag or something entirely different. 

This is how it works currently: We need a base image in which we build the agent and patched cloudbeat to test. 

We could explore updating those values automatically in a separate job, perhaps, but I believe this is out of the scope of this PR, right?

.github/actions/k8s-ci/action.yml
moukoublen: It's not needed; left it there while exporting the docker image build elsewhere. I will clean it up.

.github/actions/azure-ci/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Check for findings
+      working-directory: ./tests
+      env:
+        USE_K8S: "false"
+      shell: bash
+      run: poetry run pytest -k "azure" --alluredir=./allure/results/ --clean-alluredir
+
+    - name: Stop cloudbeat
+      shell: bash
+      run: pkill -15 cloudbeat
```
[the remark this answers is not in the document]
moukoublen: I don't think it is needed. I remember seeing it somewhere, but perhaps I am mistaken. I will remove it.

.github/actions/docker-images/action.yml
romulets: Can be outside of the scope.

I'm just not a fan of a default with a tag that will be outdated very soon

.github/actions/hermit/action.yml
moukoublen: Feel free to look at the screenshot in the description, I just uploaded; I tried to create a high-level picture of how and what.

moukoublen:
@gurevichdmitry Yes eks could be added as well and separated it by `if: ${{ github.event_name != 'pull_request' }}` to run only after merge. 

But it might be better to add it as a separate task to keep the scope of this PR as is if its ok by you also.

.github/workflows/ci.yml
moukoublen: My concern was not to make the root title verbose because is is combined with the title of each sub-job in PR UI. 
Screenshot 2024-03-27 at 11 04 13 AM


But sure why not.

gurevichdmitry:
> @gurevichdmitry Yes eks could be added as well and separated it by `if: ${{ github.event_name != 'pull_request' }}` to run only after merge.
> 
> But it might be better to add it as a separate task to keep the scope of this PR as is if its ok by you also.

@moukoublen, I agree. It would be better to handle the EKS flow as a separate task. Could you please open one for that?

.github/actions/docker-images/action.yml
oren-zohar: > I am not sure I understood if you suggest using a different tag or something entirely different.
> 
> This is how it works currently: We need a base image in which we build the agent and patched cloudbeat to test.
> 
> We could explore updating those values automatically in a separate job, perhaps, but I believe this is out of the scope of this PR, wdyt?

I think APM implemented an action for something like that

PR #2010, opened 2024-03-08
Asset Inventory POC structrue & EC2 Implementation

romulets:
### Summary of your changes

As part of testing some technical concepts for the Asset Inventory project a new cloudbeat flavour is being added to hold the Asset Inventory Proof of Concept.

This PR adds a new flavour and implements EC2 fetching

to remove the host data configure cloudbeat.yml with:

```yaml
processors:
  - drop_fields:
      fields: ["host.name"]
#  - add_host_metadata: ~
#  - add_cloud_metadata : ~
#  - add_docker_metadata: ~
```

### Screenshot/Data
image

### Related Issues
- Related: https://github.com/elastic/security-team/issues/8710

### Checklist


internal/inventory/inventory.go
```diff
@@ -0,0 +1,126 @@
...
+	PublishAll([]beat.Event)
+}
+
+func NewAssetInventory(logger *logp.Logger, fetchers []AssetFetcher, publisher AssetPublisher, now func() time.Time) AssetInventory {
+	logger.Info("Initializing Asset Inventory POC")
+	return AssetInventory{
+		logger:    logger,
+		fetchers:  fetchers,
+		publisher: publisher,
+		// move to a configuration parameter
```
kubasobon, line 56: This looks like a leftover TODO

internal/inventory/asset.go
```diff
@@ -0,0 +1,169 @@
...
+	CategoryInfrastructure assetCategory = "infrastructure"
+
+	SubCategoryCompute assetSubCategory = "compute"
+
+	TypeVirtualMachine assetType = "virtual-machine"
+
+	SubTypeEC2 assetSubType = "ec2"
+
+	AwsCloudProvider assetCloudProvider = "aws"
+)
```
kubasobon, line 42: I know this is just a stub, but it would be neat to break down the `const` block. There will be hundreds of Categories, SubTypes and Types by the end of implementation. How about an `enum`? We can split those by cloud provider, so its easier to know `SubCategoryCompute` goes with `AwsCloudProvider` and not the others. 
Just something to consider.

internal/resources/utils/pointers/pointers.go
```diff
@@ -25,3 +25,7 @@ func Deref[T any](v *T) T {
 
 	return *v
 }
+
+func Ref[T any](v T) *T {
```
kubasobon, line 29: 👌

internal/inventory/aws/fetcher_ec2.go
```diff
@@ -0,0 +1,116 @@
...
+	return &Ec2Fetcher{
+		logger:   logger,
+		provider: provider,
+	}
+}
+
+func (e *Ec2Fetcher) Fetch(ctx context.Context, assetChannel chan<- inventory.AssetEvent) {
+	instances, err := e.provider.DescribeInstances(ctx)
+	if err != nil {
+		e.logger.Errorf("Could not list ec2 instances (%v)", err)
```
kubasobon, line 60: Nitpicking: 
```suggestion
		e.logger.Errorf("Could not list ec2 instances: %v", err)
```

internal/inventory/inventory.go
```diff
@@ -0,0 +1,126 @@
...
+				"iam":     e.IAM,
+			},
+		}
+	})
+
+	a.publisher.PublishAll(events)
+}
+
+func generateIndex(a Asset) string {
+	return fmt.Sprintf("asset_inventory_%s_%s_%s_%s", a.Category, a.SubCategory, a.Type, a.SubStype)
```
kubasobon, line 121: Since the Categories, SubCategories, Types, and SubTypes are used to generate index name, are there any symbols we'd like to avoid in their definitions? It would be good to mention this in comments in `internal/inventory/asset.go`.

If I were assigned to create a new fetcher for a new type of resource, I would appreciate it if there was a comment that warns me about how the names will be used when I define another entry in the 'const' or 'enum' blocks.

kubasobon (commented):
I like the design, left some questions and minor comments.

internal/inventory/inventory.go
romulets: Jah, it's indeed. For the poc I decided to keep it here, I don't see a reason to make it an external configuration.

Do you disagree?

internal/inventory/asset.go
romulets: I agree.

For now I'll just split. Once we have more cloud providers we can think of a more elaborated way to organise stuff

internal/inventory/inventory.go
kubasobon: Makes sense to me. Just making sure you haven't missed it by accident :)

kubasobon (approved):
Looks good to me, my concerns were addressed

PR #2068, opened 2024-03-28
fix ci results comment

moukoublen:
### Summary of your changes

Fix the missing comment in PRs with allure report.

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -298,6 +298,7 @@ jobs:
     timeout-minutes: 60
     permissions:
       pull-requests: write
+      checks: write
     if: ${{ success() || failure() }}
     env:
       AWS_REGION: "eu-west-1"
@@ -318,8 +319,8 @@ jobs:
         if: ${{ success() || failure() }}
         shell: bash
         run: |
-          ls -lahR || true
-          rm -f tests/allure/results/cover.out || true
+          ls -lahR tests/allure/results/ || true
+          echo "event_name: ${{ github.event_name }}"
 
       - name: Publish allure report
         if: ${{ success() || failure() }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -318,8 +318,8 @@ jobs:
         if: ${{ success() || failure() }}
         shell: bash
         run: |
-          ls -lahR || true
-          rm -f tests/allure/results/cover.out || true
+          ls -lahR tests/allure/results/ || true
+          echo "event_name: ${{ github.event_name }}"
```
gurevichdmitry, line 322: @moukoublen, to prevent script injection, it's better to store `github.event_name` as an environment variable.
```
shell: bash
env:
    EVENT_NAME: ${{ github.event_name }}
run: |
    ls -lahR tests/allure/results/ || true
    echo "event_name: ${EVENT_NAME}"
```

.github/workflows/ci.yml
```diff
@@ -298,6 +298,7 @@ jobs:
     timeout-minutes: 60
     permissions:
       pull-requests: write
+      checks: write
```
gurevichdmitry, line 301: Why do we need this permission?

moukoublen: To be honest, I am not sure if we actually need it. We shouldn't.

The issue is that summary entries and comments do not appear in forks-originated PRs; they do appear in PRs from this repo. 

So my initial assumption is that: 

* Either `github.event_name` in case of `pull_request_target` is not `'pull_request'` (so I changed the `if` to `if: ${{ success() && github.event_name != 'push' }}`) 
* Or this permission is needed (because it is also present in this workflow https://github.com/elastic/cloudbeat/blob/2451b25449ded5bef249d134eede8148bce16efe/.github/workflows/publish-test-results.yml#L24)

Do you think we should first merge the first only, and if it does not work, then we should try the `check: write` permission?

gurevichdmitry: Let's attempt to merge without it (checks: write), and if necessary, we can add this permission later.

PR #2080, opened 2024-04-02
Move opa ci to new ci workflow

moukoublen:
### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


gurevichdmitry (approved):

.github/actions/opa/action.yml
```diff
@@ -0,0 +1,39 @@
+name: 'OPA CI'
```
oren-zohar, line 1: Why change workflow name?
```suggestion
Test OPA Policies
```

moukoublen: This is mainly because it's not just testing; it involves fmt check, bundle building, test, opa check, and lint with Regal. So, I thought `CI` better describes the steps.

But I could change it to `Test OPA Policies`.

.github/actions/aws-ci/action.yml
```diff
@@ -51,15 +51,19 @@ runs:
...
 
-    - if: ${{ (success() || failure()) && inputs.debug == 'true' }}
+    - if: ${{ failure() || cancelled() || inputs.debug == 'true' }}
       name: Upload cloudbeat logs
       uses: actions/upload-artifact@v4
       with:
         name: cloubeat-logs-ci-aws
         path: logs/
+        if-no-files-found: warn
+        retention-days: 5
```
gurevichdmitry, line 68: @moukoublen I'm not sure if we need to store debug logs for more than a day. Enabling debug mode in the workflow and generating logs is usually a proactive action, and typically, you'll use those logs on the same day.

.github/actions/aws-ci/action.yml
```diff
@@ -51,15 +51,19 @@ runs:
       run: poetry run pytest -k "aws" --alluredir=./allure/results/ --clean-alluredir
 
     - name: Upload test results
-      if: ${{ success() || failure() }}
+      if: ${{ always() }}
```
gurevichdmitry, line 54: Several consecutive commits will cancel the workflow run, and in such cases, we don't need to upload test results. So, we can use `!cancelled()` instead of `always()`.

.github/workflows/ci.yml
```diff
@@ -306,7 +328,7 @@ jobs:
     timeout-minutes: 60
     permissions:
       pull-requests: write
-    if: ${{ success() || failure() }}
+    if: ${{ always() }}
```
gurevichdmitry, line 331: The same applies here; I believe using `!cancelled()` is better than `always()`

gurevichdmitry (commented):
Added some comments

.github/actions/aws-ci/action.yml
moukoublen: It used to be like that (`success() || failure()` == `!cancelled()` ), and I specifically changed it because when we have missing results, then pytest retries, and it gets canceled, so the job is marked as canceled. 

In this case, we want the logs. That's why I changed it.

gurevichdmitry: I'm not entirely clear on the scenario. Are you referring to situations where pytest is running and the results are incomplete when the job gets canceled? If the job is interrupted or canceled, I'm not convinced we need to retain the incomplete results.

moukoublen: Lets say that because of some bug in code or testcase, or cloud resource configuration, we end up with some expected results to be missing.

Pytest retries trying to find the expected rule with the expected status. 

In that case, the job's timeout might be reached before the pytest ends with an exit code that indicates an error. 

If this happens, we must upload the test results to identify which test case didn't pass. 

I don't remember the specific CI that this happened I could search. 


Perhaps we could better sync pytest timeout vs job timeout?

.github/actions/aws-ci/action.yml
moukoublen: CHanged it to `1`.

.github/actions/aws-ci/action.yml
moukoublen: Yes, unfortunately, it introduces a bit of noise and a small delay on consecutive commits. 😞 

Let's see if it becomes a bummer we can revert back to `not canceled`.

Issue #2079, opened 2024-04-01
Replace `EcsGcp` with `CloudAccountMetadata`


orouz:
**Motivation**

these two structs are basically the same, we should use just `CloudAccountMetadata` which is already used in aws/azure.

**Definition of done**
- `EcsGcp` is removed and its usage replaced with `CloudAccountMetadata`

PR #2085, opened 2024-04-03
use CloudAccountMetadata in GCP

orouz:
### Summary of your changes

AWS and Azure both use `fetching.CloudAccountMetadata` to describe cloud metadata and `resource.GetMetadata` to populate it (via `EnrichEvent` of cloud data provider). 

GCP uses `EcsGcp` to describe the same cloud metadata and uses `resource.GetElasticCommonData` to populate it. 

this PR makes it so that GCP describes and populates cloud data the same way AWS and Azure do it. 

### Screenshot/Data

Screenshot 2024-04-03 at 6 55 38

### Related Issues
- closes https://github.com/elastic/cloudbeat/issues/2079

### DIFF
--- a/internal/resources/providers/gcplib/inventory/provider.go
+++ b/internal/resources/providers/gcplib/inventory/provider.go
@@ -42,7 +42,7 @@ type Provider struct {
 	config    auth.GcpFactoryConfig
 	inventory *AssetsInventoryWrapper
 	crm       *ResourceManagerWrapper
-	crmCache  map[string]*fetching.EcsGcp
+	crmCache  map[string]*fetching.CloudAccountMetadata
 }
 
 type AssetsInventoryWrapper struct {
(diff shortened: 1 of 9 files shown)


internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -390,16 +387,16 @@ func getAssetsByProject[T any](assets []*ExtendedGcpAsset, log *logp.Logger, f T
 			continue
 		}
 
-		ecs := projectAssets[0].Ecs
+		c := projectAssets[0].CloudAccount
```
uri-weisman, line 390: I would use a more informative variable name

uri-weisman (approved):
I suggest testing it with a GCP Org as well, just to verify that nothing breaks.

orouz:
> I suggest testing it with a GCP Org as well, just to verify that nothing breaks.

✅  tested on a GCP org (cspelastic)

Issue #2005, opened 2024-03-05
[GCP] Automated credentials creation flow for agentless


orouz:
in the manual deployment of GCP integration the user is required to provide credentials. 

in https://github.com/elastic/kibana/issues/175977 there is an example of how to deploy `access_key_id` and `secret_access_key` so the user can copy+paste them easily 

this task is for doing the same but for GCP which uses `cerdentials json`

PR #2097, opened 2024-04-07
add GCP credentials deployment

orouz:
### Summary of your changes

adds a deployment manager script that will:
1. enable all required APIs for CSPM GCP 
2. create a service account with bindings and a key. 
3. print the key after creation so user can copy it to the CSPM GCP integration

this script is meant to be accompanied by a kibana PR to introduce a button to "Deploy Credentials" in agentless mode, so a user installing CSPM GCP can paste created key into the the `Credentials JSON` input field. 

the button itself should go to: 
```
https://shell.cloud.google.com/cloudshell/?ephemeral=true&cloudshell_git_repo=https%3A%2F%2Fgithub.com%2Felastic%2Fcloudbeat&cloudshell_git_branch=<latest_stack_version_minor>&cloudshell_workspace=deploy%2Fdeployment-manager&show=terminal
```
where `latest_stack_version_minor` is something like `8.13` 

cc @maxcold 

### Screenshot/Data

Screenshot 2024-04-07 at 14 26 04

verified the created key allows calling `ListAssets` by running:
```
gcloud asset list --project="<project>" --content-type=iam-policy --asset-types=cloudresourcemanager.googleapis.com/Project --format json --impersonate-service-account=elastic-agent-cspm-user-sa@<project>.iam.gserviceaccount.com
```

### Related Issues
 - closes https://github.com/elastic/cloudbeat/issues/2005

### Checklist


oren-zohar:
@orouz Do we need to update the docs after this?

orouz:
> Do we need to update the docs after this?

i guess so, but i think that will be done by the UX writes. first there needs to be a follow-up PR in kibana to use this. i'm not sure about the UX though, having the user copy-paste the credentials, maybe we would like to send the data back to elastic and have it automatically filled in the integration installation form. i think that's the end goal in https://github.com/elastic/security-team/issues/6053. will discuss this with @tinnytintin10 to see how to move forward.

deploy/deployment-manager/deploy_compute.sh
orouz, line 1: renamed this script because we now have 2 deployments - `compute` and `service_account`

orouz: updated the path in the kibana deployment instructions - https://github.com/elastic/kibana/pull/180369

orouz:
closing this in favor of https://github.com/elastic/cloudbeat/pull/2118 as it will include a CI test as well

PR #2114, opened 2024-04-10
Revert "[CIS GCP] Add owner label to a deployment manager deployments"

uri-weisman:
- Reverts elastic/cloudbeat#2088
- Fix wrong working directory paths in the dm-test workflow 

We want to avoid interfering with the customer's label policy.


oren-zohar:
@uri-weisman any idea why the DM test is failing?

orouz:
> @uri-weisman any idea why the DM test is failing?

i _think_ it needs to run on upstream (cloudbeat) and not a fork

PR #2115, opened 2024-04-10
Update DM test workflow

uri-weisman:
### Summary of your changes
The DM workflow runs only when we change DM-related files, meaning it didn’t run for a while.
It seems that some test file locations were changed due to a refactor that was done recently.

### DIFF
--- a/.github/workflows/test-gcp-dm.yml
+++ b/.github/workflows/test-gcp-dm.yml
@@ -7,6 +7,7 @@ on:
       - "[0-9]+.[0-9]+"
     types: [opened, synchronize, reopened]
     paths:
+      - ".github/workflows/test-gcp-dm.yml"
       - "deploy/deployment-manager/compute_engine.py"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-gcp-dm.yml
```diff
@@ -80,7 +82,7 @@ jobs:
 
       - name: Install CSPM GCP integration
         id: cspm-gcp-integration
-        working-directory: deploy/test-environments/fleet_api
+        working-directory: tests/integrations_setup
```
orouz, line 85: i didn't see there is a PR for this already so opened one too https://github.com/elastic/cloudbeat/pull/2125/files
let me know if you want to use this one or the other one
in anyway -
1. i think `poetry install` will fail as the `toml` file is not in `tests/integrations_setup`
2. i think `poetry run python src/install_cspm_gcp_integration.py` will fail as the working dir has no `src`

PR #2118, opened 2024-04-10
add GCP credentials deployment 

orouz:
- [x] adds a Deployment Manager template to deploy a new service account and key
- [x] adds a CI test to verify deployment of service account is useable

- this is how a successful deployment looks from a user POV after they ran the service account deployment script in a GCP shell:
Screenshot 2024-04-15 at 18 25 05

- tested we get findings using the generated key by providing it as a JSON blob credentials option on the manual deployment option:
Screenshot 2024-05-08 at 15 09 18

- closes https://github.com/elastic/cloudbeat/issues/2005

### DIFF
--- a/.github/workflows/test-gcp-dm.yml
+++ b/.github/workflows/test-gcp-dm.yml
@@ -12,6 +12,9 @@ on:
       - "deploy/deployment-manager/compute_engine.py.schema"
       - "deploy/deployment-manager/deploy.sh"
       - "deploy/deployment-manager/set_env.sh"
+      - "deploy/deployment-manager/deploy_service_account.sh"
+      - "deploy/deployment-manager/service_account.py"
+      - "deploy/deployment-manager/service_account.py.schema"
 
 env:
   TEST_ENVS_DIR: deploy/test-environments
--- a/.github/workflows/test-gcp-dm.yml
+++ b/.github/workflows/test-gcp-dm.yml
@@ -226,14 +226,12 @@ jobs:
           poetry run pytest -k "cspm_gcp" --alluredir=./allure/results/ --clean-alluredir --maxfail=4
 
       - name: Destory EC deployment
-        # if: steps.deploy_ec.outcome == 'success'
         if: always()
         working-directory: ${{ env.TEST_ENVS_DIR }}
         run: |
           terraform destroy --auto-approve -target="module.ec_deployment" -target="module.ec_project"
 
       - name: Delete GCP Deployments
-        # if: steps.gcp_deploy.outcome == 'success'
         if: always()
         working-directory: ${{ env.TEST_ENVS_DIR }}
         run: |
--- a/.github/workflows/test-gcp-dm.yml
+++ b/.github/workflows/test-gcp-dm.yml
@@ -14,14 +14,16 @@ on:
       - "deploy/deployment-manager/set_env.sh"
 
 env:
-  WORKING_DIR: deploy/test-environments
+  TEST_ENVS_DIR: deploy/test-environments
   INTEGRATIONS_SETUP_DIR: tests/integrations_setup
+  DEPLOYMENT_MANAGER_DIR: deploy/deployment-manager
   TF_VAR_ec_api_key: ${{ secrets.EC_API_KEY }}
   TF_VAR_ess_region: gcp-us-west2 # default region for testing deployments
 
 jobs:
-  Test-GCP-DM:
-    name: GCP Deployment Manager Test
+  # Test a GCP Deployment Manager deployment using Application Default Credentials
+  gcp_dm_adc:
+    name: CSPM GCP with ADC
     runs-on: ubuntu-22.04
     timeout-minutes: 60
     permissions:
@@ -65,7 +67,7 @@ jobs:
       - name: Provision Test Environment (EC)
         id: apply
         if: success()
-        working-directory: ${{ env.WORKING_DIR }}
+        working-directory: ${{ env.TEST_ENVS_DIR }}
         run: |
           terraform -v
           terraform init
@@ -108,7 +110,7 @@ jobs:
 
       - name: Destory EC deployment
         if: always()
-        working-directory: ${{ env.WORKING_DIR }}
+        working-directory: ${{ env.TEST_ENVS_DIR }}
         run: |
           terraform destroy --auto-approve -target="module.ec_deployment" -target="module.ec_project"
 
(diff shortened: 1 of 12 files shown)


orouz:
this is waiting for https://github.com/elastic/cloudbeat/pull/2115

.github/workflows/test-gcp-dm.yml
```diff
@@ -14,14 +14,16 @@ on:
...
+  DEPLOYMENT_MANAGER_DIR: deploy/deployment-manager
   TF_VAR_ec_api_key: ${{ secrets.EC_API_KEY }}
   TF_VAR_ess_region: gcp-us-west2 # default region for testing deployments
 
 jobs:
-  Test-GCP-DM:
-    name: GCP Deployment Manager Test
+  # Test a GCP Deployment Manager deployment using Application Default Credentials
+  gcp_dm_adc:
+    name: CSPM GCP with ADC
```
orouz, line 26: this job tests the code path that uses GCP's application default credentials 
https://github.com/elastic/cloudbeat/blob/e8309ef67d6bb49119a9464ff00ae0b9bf815a44/internal/resources/providers/gcplib/auth/credentials.go#L59

.github/workflows/test-gcp-dm.yml
```diff
@@ -118,9 +120,121 @@ jobs:
...
-        working-directory: ${{ env.WORKING_DIR }}
+        working-directory: ${{ env.TEST_ENVS_DIR }}
         run: |
           DEPLOYMENT=${{env.GCP_DEPLOYMENT_NAME}}
           PROJECT_NAME=$(gcloud config get-value core/project)
           PROJECT_NUMBER=$(gcloud projects list --filter="${PROJECT_NAME}" --format="value(PROJECT_NUMBER)")
           ./delete_gcp_env.sh $PROJECT_NAME $PROJECT_NUMBER $DEPLOYMENT
+
+  # Test a GCP Deployment Manager deployment using a Service Account
+  gcp_dm_sa:
```
orouz, line 131: this job tests the code path that uses a service account json 
https://github.com/elastic/cloudbeat/blob/e8309ef67d6bb49119a9464ff00ae0b9bf815a44/internal/resources/providers/gcplib/auth/credentials.go#L63

.github/workflows/test-gcp-dm.yml
orouz, line 1: there is some repetitiveness in this workflow as the two jobs do mostly the same thing. we could abstract some of the steps to an action (like `actions/init_env`, `actions/check_findings`, `actions/deploy_ec`)

.github/workflows/test-gcp-dm.yml
```diff
@@ -118,9 +120,121 @@ jobs:
...
+          # Installs CSPM GCP integration
+          cd ../../${{ env.INTEGRATIONS_SETUP_DIR }}
+          export SERVICE_ACCOUNT_JSON_PATH="KEY_FILE.json"
+          export DEPLOYMENT_NAME="${{ env.GCP_AGENT_DEPLOYMENT_NAME }}"
+          poetry install
+          poetry run python ./install_cspm_gcp_integration.py
+
+          # Deploys the agent using an existing service account (SERVICE_ACCOUNT_NAME)
+          cd ../../${{ env.DEPLOYMENT_MANAGER_DIR }}
+          . ./set_env.sh && ./deploy.sh
```
orouz, line 218: there's some trickery going on here:
1. we first deploy a service account using the new deployment manager template, which generates a `KEY_FILE.json` with the said service account
2. we then install a CSPM GCP integration that uses `gcp.credentials.json` variable to set the JSON content 
3. then we deploy the agent, but because we've defined a service account name, it won't deploy a service account and instead use the one we deployed earlier

this is all combined into a single step because otherwise we won't be able to share the JSON file

deploy/deployment-manager/compute_engine.py
```diff
@@ -52,7 +51,7 @@ def generate_config(context):
             ),
             "serviceAccounts": [
                 {
-                    "email": f"$(ref.{sa_name}.email)",
+                    "email": get_service_account_email(sa_name, project),
```
orouz, line 54: we can't use `ref` because we might be using an already-deployed service account

deploy/deployment-manager/deploy_service_account.sh
orouz, line 1: this is mostly copied from the original `./deploy.sh` script

deploy/deployment-manager/service_account.py
```diff
@@ -0,0 +1,65 @@
...
+def generate_config(context):
+    """Generates service account user"""
+    deployment_name = context.env["deployment"]
+    scope = context.properties["scope"]
+    parent_id = context.properties["parentId"]
+    sa_name = context.properties["serviceAccountName"]
+    roles = ["roles/cloudasset.viewer", "roles/browser"]
+
+    # pylint: disable=duplicate-code
+    service_account = {
```
orouz, line 13: it'd be nice to define a `./common.py` file and import it to both deployment templates. not sure why that didn't work tbh

deploy/deployment-manager/deploy_service_account.sh
uri-weisman: why not create a common.sh file?

deploy/deployment-manager/service_account.py
uri-weisman: I didn't manage to do it as well.
It's because we're using deployment manager templates and other py files can not be imported to the file.

deploy/deployment-manager/deploy.sh
```diff
@@ -115,7 +116,7 @@ fi
 # Apply the deployment manager templates
 run_command "gcloud deployment-manager deployments create --automatic-rollback-on-error ${DEPLOYMENT_NAME} --project ${PROJECT_NAME} \
     --template compute_engine.py \
-    --properties elasticAgentVersion:${STACK_VERSION},fleetUrl:${FLEET_URL},enrollmentToken:${ENROLLMENT_TOKEN},allowSSH:${ALLOW_SSH},zone:${ZONE},elasticArtifactServer:${ELASTIC_ARTIFACT_SERVER},scope:${SCOPE},parentId:${PARENT_ID}"
+    --properties elasticAgentVersion:${STACK_VERSION},fleetUrl:${FLEET_URL},enrollmentToken:${ENROLLMENT_TOKEN},allowSSH:${ALLOW_SSH},zone:${ZONE},elasticArtifactServer:${ELASTIC_ARTIFACT_SERVER},scope:${SCOPE},parentId:${PARENT_ID},serviceAccountName:${SERVICE_ACCOUNT_NAME}"
```
uri-weisman, line 120: Will the `SERVICE_ACCOUNT_NAME` param be propagated from the UI and if not we'll create one?
In case the user provides a wrong SA name, the deployment will not succeed and an error will be printed to the user.
Are we ok with it? do we want to add validation that the SA exists and if not to create one?

tests/integrations_setup/install_cspm_gcp_integration.py
```diff
@@ -48,6 +48,19 @@
...
+    try:
+        with open(file_path, "r") as json_file:
+            return json_file.read()
+    except FileNotFoundError:
+        logger.error(f"Error: File '{file_path}' not found.")
+        sys.exit(1)
+    except IOError as e:
+        logger.error(f"Error reading file '{file_path}': {e}")
+        sys.exit(1)
+
```
gurevichdmitry, line 62: @orouz, I think there's a bit of confusion here. The function `read_json_file()` and the variable `json_file` might suggest that they specifically handle JSON files, but they're actually generic and can read any type of file. For reading JSON files, we have a separate function defined [here](https://github.com/elastic/cloudbeat/blob/main/tests/fleet_api/utils.py#L26).

deploy/deployment-manager/deploy.sh
gurevichdmitry: I'm also unsure about the case where a service account is not provided at all. In that case, it will default to serviceAccountName:false. Is this case considered?

tests/integrations_setup/install_cspm_gcp_integration.py
orouz: @gurevichdmitry i would prefer using `read_json` but it returns an empty dict when it fails (also missing decode exception) and i don't want to check for an empty dict in my use case and then exit, as it's kinda dirty. i'd prefer updating `read_json` to remove the `try` and let consumers do it and act on errors. in my case, i would `sys.exit(1)` if reading the file failed. can we do the same in [`load_data()`](https://github.com/elastic/cloudbeat/blob/e8309ef67d6bb49119a9464ff00ae0b9bf815a44/tests/integrations_setup/install_d4c_integration.py#L47-L48) ?

deploy/deployment-manager/deploy.sh
orouz: > Will the `SERVICE_ACCOUNT_NAME` param be propagated from the UI and if not we'll create one? 

no, `SERVICE_ACCOUNT_NAME` is not user-facing. 

the addition of `SERVICE_ACCOUNT_NAME` to the `deploy.sh` script is done purely for testing purposes of deploying an agent with an existing service account. it simulates the new agentless flow, where a user will deploy a service account and we'd use its key in the integration credentials input. 

to test the deployment of a service account is working properly, i've used the existing agent deployment script to make it so that it can operate either by deploying the full resources: VM + SA, or just a VM, while using an existing SA, which is currently passed only during tests. 

> In case the user provides a wrong SA name, the deployment will not succeed and an error will be printed to the user. Are we ok with it? do we want to add validation that the SA exists and if not to create one?

as mentioned, `SERVICE_ACCOUNT_NAME` is only used in tests, so we don't want to deploy a new service account if `SERVICE_ACCOUNT_NAME` doesn't exists. if for some reason `SERVICE_ACCOUNT_NAME` is provided but doesn't exists, failing the deployment would be the right outcome.

uri-weisman (approved):
Approving deployment manager files

tests/integrations_setup/install_cspm_gcp_integration.py
gurevichdmitry: @orouz considering that `read_json` is only used in `load_data`, we can modify `read_json` to exit on error. This approach fulfills your requirement, and the reuse will be valid in your case.

gurevichdmitry (approved):
@orouz Workflow and tests are approved. One more question: did you test the environment destruction when the service account is used?

orouz:
> @orouz Workflow and tests are approved. One more question: did you test the environment destruction when the service account is used?

yes, deletion of GCP env works

PR #2126, opened 2024-04-11
Add correct host info for cloud provider virtual machines

kubasobon:
By default, beats are expected to fill the `host` section with data of a machine the beat/agent runs on. This PR changes the logic by disabling the default processor providing the data and enriching select resources with asset data fetched from cloud providers.

:warning: It is worth noting that adding the `add_host_metadata` processor cloudbeat config file (e.g. `data/cloudbeat.yml`) will result in the old behavior.

### Summary of your changes

Changes made:
- drop Host processor from the pipeline
- publish `host.*` data for Azure VMs
- publish `host.*` data for GCP CloudCompute Instances

Please note the change will affect findings for rules related to virtual machine resources:
- CIS Azure 7.1 - 7.4
- CIS GCP 4.1 - 4.9, 4.11

All other findings will have an empty `host` section.

### Screenshot/Data

<details>
<summary>Data before the change</summary>

image

</details>

<details>
<summary>Data after the change</summary>

image

</details>

### Related Issues

Related to https://github.com/elastic/cloudbeat/issues/2081

### Checklist
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added the necessary README/documentation (if appropriate)

#### Introducing a new rule?

No

PR #2151, opened 2024-04-22
Fix host info for GCP Compute Instances

kubasobon:
A quickfix for https://github.com/elastic/cloudbeat/pull/2126
Proof:
image

### DIFF
--- a/internal/resources/fetching/fetchers/gcp/assets_fetcher_test.go
+++ b/internal/resources/fetching/fetchers/gcp/assets_fetcher_test.go
@@ -135,9 +135,10 @@ func (s *GcpAssetsFetcherTestSuite) TestFetcher_ElasticCommonData() {
 
 		asset := &GcpAsset{
 			Type:    fetching.CloudCompute,
-			SubType: inventory.ComputeInstanceAssetType,
+			SubType: "gcp-compute-instance",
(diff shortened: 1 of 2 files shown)


internal/resources/fetching/fetchers/gcp/assets_fetcher_test.go
```diff
@@ -135,9 +135,10 @@ func (s *GcpAssetsFetcherTestSuite) TestFetcher_ElasticCommonData() {
 
 		asset := &GcpAsset{
 			Type:    fetching.CloudCompute,
-			SubType: inventory.ComputeInstanceAssetType,
+			SubType: "gcp-compute-instance",
```
romulets, line 138: Maybe a good idea to add a constant for it?

romulets (changes requested):

internal/resources/fetching/fetchers/gcp/assets_fetcher_test.go
kubasobon: This is used only for test purposes. In code we use `getGcpSubType()` which outputs this value.

moukoublen (approved):

romulets (approved):

Issue #2142, opened 2024-04-17
[CNVM] Trivy local cache file size increases indefinitely 


moukoublen:
**Describe the bug**
Trivy uses a local file (bbolt db) as a cache in the `/tmp` directory (`/tmp/trivy/fanal/fanal.db`) that always increases in size with each cycle. 

This results in the `tmpfs` file system holding the `/tmp` folder getting filled up _<sup>(1)</sup>_, and Cloudbeat can no longer download the new trivy db (which it does on each cycle). This leads to Cloudbeat's crash loop and not providing cnvm findings. It could also have implications for other applications hosted in the same instance that could use `/tmp` for any crucial operation.

**<sup>(2)</sup>** _`/tmp` (`tmpfs`) is a ram disk (placed in ram) with a maximum size, usually half of the host's total ram._ 

(Example screenshots of `fanal.db` size before and after some runs)
<img width="394" alt="Screenshot 2024-04-16 at 11 01 33 AM" src="https://github.com/elastic/cloudbeat/assets/1380039/872fa5ca-680c-4aa2-a942-651a1d439e55">
<img width="332" alt="Screenshot 2024-04-16 at 11 38 12 AM" src="https://github.com/elastic/cloudbeat/assets/1380039/1125ed7a-cacf-4956-a907-4cd8de72b92a">
<img width="294" alt="Screenshot 2024-04-17 at 9 03 19 AM" src="https://github.com/elastic/cloudbeat/assets/1380039/33df961b-7543-41d2-8c99-380a9d2b9707">
```plain
[ec2-user ~]$ sudo tree -h /tmp
/tmp
└── [   60]  trivy
    └── [   60]  fanal
        └── [ 1.9G]  fanal.db

12 directories, 1 file
[ec2-user ~]$ free -h
               total        used        free      shared  buff/cache   available
Mem:            15Gi       587Mi       8.4Gi       1.9Gi       6.3Gi        12Gi
Swap:             0B          0B          0B
```

**Preconditions**
Any cnvm deployment. 

**To Reproduce**
1. Create a cnvm deployment with agent version >= 8.12 (pending checking older releases as well).
2. Wait for many runs to pass (depending on the cloud assets, and host's ram size).

**Expected behavior**
Cloudbeat will be able to work indefinitely and produce events on each cycle. 

**Workaround till the fix**
Restarting the host machine will delete everything from `/tmp,` and thus the `fanal.db` so Cloudbeat can continue to work and produce findings.

[...]

PR #2168, opened 2024-05-01
Clean trivy cache before each cycle

moukoublen:
### Summary of your changes
**Clean trivy cache after each cycle** 

Trivy [artifact.NewRunner](https://github.com/aquasecurity/trivy/blob/main/pkg/commands/artifact/run.go#L116) calls [initCache](https://github.com/aquasecurity/trivy/blob/main/pkg/commands/artifact/run.go#L341). 

Depending on the provided configuration, the latter (initCache) could either return a runner or the error `SkipScan` (and nil runner).

In case one of the below flags is `true`, the `artifact.NewRunner` will perform the relevant clean/reset operation and return `SkipScan`. 

  * `CacheOptions.ClearCache`
  * `DBOptions.Reset`
  * `MisconfOptions.ResetPolicyBundle`

This means that:
  1. The clean/reset operations are embedded inside `initCache`, called inside `artifact.NewRunner`. We cannot execute them standalonely. 
  2. When at least one on those flags are true, the clean/reset operation takes place and returns `SkipScan` and `nil` runner, which means we cannot set the flags and get a valid runner with a single call (that's why I created a separate function `ClearCache` in my code changes). 
  3. `initCache` could perform only one of each clean/reset operation and then return `SkipScan`, which means we cannot perform a single `artifact.NewRunner` call and init all of those flags to true and expect all those clear/reset operations to run. We need to do one at a time (that's why I created the loop inside `ClearCache` in my code changes).
 
Also Note: `initCache` with reset/clean flags could run only when no other bbolt client "holds" open the db. So it must be run before we create the `artifact.NewRunner`.

### Screenshot/Data

With this PR the below happens at the beginning of the cycle:

Screenshot 2024-05-17 at 4 42 01 PM

### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/2142

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/internal/vulnerability/runner.go
+++ b/internal/vulnerability/runner.go
@@ -34,7 +34,13 @@ type VulnerabilityRunner struct {
 }
 
 func NewVulnerabilityRunner(log *logp.Logger) (VulnerabilityRunner, error) {
+	ctx := context.Background() // TODO: carry
 	log.Debug("NewVulnerabilityRunner: New")
+
+	if err := clearTrivyCache(ctx, log); err != nil {
+		log.Warnf("error during runner cache clearing %s", err.Error())
+	}
+
 	opts := flag.Options{
 		GlobalOptions: flag.GlobalOptions{
 			// TODO: Make configurable
@@ -56,7 +62,7 @@ func NewVulnerabilityRunner(log *logp.Logger) (VulnerabilityRunner, error) {
 		},
 	}
 
-	runner, err := artifact.NewRunner(context.Background(), opts)
+	runner, err := artifact.NewRunner(ctx, opts)
 	if err != nil {
 		log.Error("NewVulnerabilityRunner: NewRunner error: ", err)
 		return VulnerabilityRunner{}, err
@@ -72,7 +78,8 @@ func (f VulnerabilityRunner) GetRunner() artifact.Runner {
 	return f.runner
 }
 
-func (f VulnerabilityRunner) ClearCache(ctx context.Context) error {
+func clearTrivyCache(ctx context.Context, log *logp.Logger) error {
+	log.Info("Starting VulnerabilityRunner.ClearCache")
 	options := []flag.Options{
 		{CacheOptions: flag.CacheOptions{ClearCache: true}},
 		{DBOptions: flag.DBOptions{Reset: true}},
@@ -93,6 +100,8 @@ func (f VulnerabilityRunner) ClearCache(ctx context.Context) error {
 		// it should never go here (NewRunner should always return artifact.SkipScan), but just in case it goes, lets close the runner.
 		errs = append(errs, r.Close(ctx))
 	}
+
+	log.Infof("Ending VulnerabilityRunner.ClearCache %d", len(errs))
 
 	return errors.Join(errs...)
 }
--- a/internal/vulnerability/runner.go
+++ b/internal/vulnerability/runner.go
@@ -83,7 +83,6 @@ func clearTrivyCache(ctx context.Context, log *logp.Logger) error {
 	options := []flag.Options{
 		{CacheOptions: flag.CacheOptions{ClearCache: true}},
 		{DBOptions: flag.DBOptions{Reset: true}},
-		// {MisconfOptions: flag.MisconfOptions{ResetPolicyBundle: true}},
 	}
 
 	errs := make([]error, 0, len(options))
--- a/internal/vulnerability/runner.go
+++ b/internal/vulnerability/runner.go
@@ -19,6 +19,7 @@ package vulnerability
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/aquasecurity/trivy/pkg/commands/artifact"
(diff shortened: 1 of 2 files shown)


internal/vulnerability/runner.go
```diff
@@ -33,7 +34,13 @@ type VulnerabilityRunner struct {
 }
 
 func NewVulnerabilityRunner(log *logp.Logger) (VulnerabilityRunner, error) {
+	ctx := context.Background() // TODO: carry
```
romulets, line 37: Todo 😁

internal/vulnerability/runner.go
```diff
@@ -33,7 +34,13 @@ type VulnerabilityRunner struct {
...
 	log.Debug("NewVulnerabilityRunner: New")
+
+	if err := clearTrivyCache(ctx, log); err != nil {
+		log.Warnf("error during runner cache clearing %s", err.Error())
```
romulets, line 41: Why is this a warning and not an error? I mean, this can lead to out of memory, right?

internal/vulnerability/runner.go
```diff
@@ -70,3 +77,30 @@ func NewVulnerabilityRunner(log *logp.Logger) (VulnerabilityRunner, error) {
...
+
+	for _, opts := range options {
+		r, err := artifact.NewRunner(ctx, opts)
+		if err != nil {
+			if !errors.Is(err, artifact.SkipScan) {
+				errs = append(errs, err)
+			}
+		}
+
+		// it should never go here (NewRunner should always return artifact.SkipScan and nil runner), but just in case it goes, lets close the runner.
```
romulets, line 98: If skip scan is always expected, I would invert the flow to highlight as main flow.

So 

```
  if errors.Is(err, trivy.SkipScan) {
    continue
  }

  // handle error and close r here then 
```

romulets (commented):
Also, don't we have any tests covering this?

internal/vulnerability/runner.go
moukoublen: Yeah, context.Background was used in this constructor function before these changes and thus the todo, because is out of the scope of this pr.

internal/vulnerability/runner.go
moukoublen: It was like that at first https://github.com/elastic/cloudbeat/pull/2168/commits/6bd63491eaacfd324c54b6ee94769b94df7cdd99

But then I wanted to be 100% sure that if a runner is returned, we should close it.

internal/vulnerability/runner.go
moukoublen: Not necessarily; it depends on the error. Since it's not critical to cycle workflow, I used the warning level.

moukoublen:
@eyalkraft I thought the DB is being downloaded on each cycle either way. But I will check and perhaps remove the db reset option. 


@romulets I couldn't think of a way to test this, to be honest, other than the already existing test (constructor does not return an error). I couldn't think of a way to safely test that "ensure that the cache is cleared after init".

internal/vulnerability/runner.go
moukoublen: ok removed the comment and I will open a task

moukoublen:
@eyalkraft 

So apparently we of the two dbs that are being downloaded the java one is cached for 3 days so in that case yes we would loose that.

example log
```plain
Repeater cycle triggered
Need to update DB
DB Repository: ghcr.io/aquasecurity/trivy-db
Downloading DB...

Java DB Repository: ghcr.io/aquasecurity/trivy-java-db:1
Downloading the Java DB...
The Java DB is cached for 3 days. If you want to update the database more frequently, the '--reset' flag clears the DB cache.
```


So i removed the `{DBOptions: flag.DBOptions{Reset: true}` and I will test again to verify that `{CacheOptions: flag.CacheOptions{ClearCache: true}},` is enough (it should be).

internal/vulnerability/runner.go
```diff
@@ -70,3 +77,29 @@ func NewVulnerabilityRunner(log *logp.Logger) (VulnerabilityRunner, error) {
...
+	// 	{MisconfOptions: flag.MisconfOptions{ResetPolicyBundle: true}},
+	// In our case we will use only the ClearCache option.
+
+	errs := make([]error, 0, 2)
+	r, err := artifact.NewRunner(ctx, flag.Options{CacheOptions: flag.CacheOptions{ClearCache: true}})
+	if err != nil {
+		if !errors.Is(err, artifact.SkipScan) {
+			errs = append(errs, err)
+		}
+	}
```
romulets, line 97: nitpick: we don't need to wrap the with `if err != nil`, the `!errors.Is(err, artifcat.SkipScan)` already covers it (https://cs.opensource.google/go/go/+/master:src/errors/wrap.go;l=44)

```suggestion
	if !errors.Is(err, artifact.SkipScan) {
		errs = append(errs, err)
	}
```

romulets (approved):

PR #2188, opened 2024-05-08
Rollback to a previous beats version to verify when the problem started

romulets:
### Summary of your changes

Debugging the issues with buildkite build.

The issue has started with this commit https://github.com/elastic/cloudbeat/commit/27baac2d89cc5b7adc4402e8819d9f7ab2401645

Which introduced the beats version of commit https://github.com/elastic/beats/commit/81f63103678f (2024-04-26)

There was multiple added commits in this updated because the previous version we were using was https://github.com/elastic/beats/commit/f871b5c5f76b (2024-03-29)

Therefore I'm trying to narrow down when the issue started, after what commit - so we can better provide information to the beats team.

The chosen commit is this one, https://github.com/elastic/beats/commit/ca4adcecac031b8c7ba83a16d1803140b018f3a1 introduced on 2024-04-13, roughly mid of the road between 2024-04-26 and 2024-03-29.

Unfortunately I have to merge this to properly test. More PR's will follow up with further tests and a complete rollback to the latest version we are at right now.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/elastic/cloudbeat
 
-go 1.22
+go 1.22.0
 
 require (
 	cloud.google.com/go/asset v1.18.1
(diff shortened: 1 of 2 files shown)


romulets:
I'll close since this is not yet ready, I need to find a better way to downgrade beats version (and other needed dependecies)

romulets:
I'll close since this is not yet ready, I need to find a better way to downgrade beats version (and other needed dependecies)

PR #2219, opened 2024-05-20
Integration of basic functional test suite developed with FTR into the Create Environment workflow

gurevichdmitry:
### Summary of your changes

This PR adds the ability to run test suites written in the `Kibana` repository using the `Functional Test Runner` (FTR). The integration is achieved through the `Create Environment` workflow. When the UI sanity check option is selected, the `Create Environment` workflow will execute UI sanity checks from the `Kibana` repository. The `main` branch of `Kibana` will be used for checkout.

Running Kibana tests using FTR was implemented as a separate `GitHub` action `kibana-ftr`, which is reused in the `Create Environment` workflow. This action checks out the specified Kibana branch without history, then installs necessary tools and performs a bootstrap. After preparation, FTR is executed with the specified configuration.

### Screenshot/Data
This PR introduces two input parameters for the Create Environment workflow: `run-ui-sanity-tests` flag, which allows you to select whether to run UI sanity tests, and `ref_commit_sha`, which allows you to choose the appropriate Kibana branch, tag, or commit SHA.

Screenshot 2024-05-22 at 14 42 05

**New Step in the workflow**
Screenshot 2024-05-22 at 14 44 25

**Report Example**
Screenshot 2024-05-22 at 14 45 27

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


.github/workflows/test-environment.yml
```diff
@@ -442,6 +429,16 @@ jobs:
...
           poetry run pytest -m "sanity" --alluredir=./allure/results/ --clean-alluredir --maxfail=4
 
+      - name: Run UI Sanity checks (Kibana)
+        uses: ./.github/actions/kibana-ftr
+        if: ${{ success() && inputs.run-ui-sanity-tests == true }}
+        with:
+          test_kibana_url: ${{ env.TEST_KIBANA_URL }}
+          test_es_url: ${{ env.TEST_ES_URL }}
+          es_version: ${{ env.STACK_VERSION }}
+          ref_commit_sha: ${{ inputs.ref_commit_sha }}
```
gurevichdmitry, line 439: Version control for Kibana UI tests has become relevant because the UI controls implementation can differ between versions. Currently, I assume that the user executing the workflow will select the correct Kibana version and branch, for example, `8.15.0-SNAPSHOT` and `main`, or in the future `8.15.0` and `v8.15.0` in Kibana.
I am considering whether we should implement logic to prevent execution if versions do not match correctly. In the current create environment workflow, we allow installing different versions, so we need to decide if we should add additional logic to enforce version consistency, or if we will simply allow the UI tests to fail if an incorrect branch/tag is provided.

dev-docs/Cloud-Env-Testing.md
```diff
@@ -45,6 +45,10 @@ Follow these steps to run the workflow:
 
     - **`run-sanity-tests`** (**optional**): Set to `true` to run sanity tests after the environment is set up. Default: `false`
 
+    - **`run-ui-sanity-tests`** (**optional**): Set to `true` to run Kibana UI sanity tests after the environment is set up. Default: `false`
```
oren-zohar, line 48: Maybe link the tests docs here (.github/actions/kibana-ftr/README.md)

dev-docs/Cloud-Env-Testing.md
```diff
@@ -45,6 +45,10 @@ Follow these steps to run the workflow:
...
+
+    - **`ref_commit_sha`** (**optional**): Specifies the Kibana branch, tag, or commit SHA to check out for the UI sanity tests, which will be executed after the environment is set up. This should correspond to the version of the `elk-stack-version` provisioned by this workflow. Note that for the current version in development, use Kibana's `main` branch. Default: `main`
```
oren-zohar, line 50: field name is too generic IMO, consider making it more clear that it's Kibana's commit, e.g. `kbn_ref_commit_sha`, also since it's not referring only the commit sha, but also the branch name, maybe more generic on that side, so `kibana_ref`

.github/workflows/test-environment.yml
```diff
@@ -240,38 +258,7 @@ jobs:
...
-          echo "EC2_KSPM_KEY=$EC2_KSPM_KEY" >> $GITHUB_ENV
-
-          export KSPM_PUBLIC_IP=$(terraform output -raw ec2_kspm_public_ip)
-          echo "::add-mask::$KSPM_PUBLIC_IP"
-          echo "KSPM_PUBLIC_IP=$KSPM_PUBLIC_IP" >> $GITHUB_ENV
-
-          export CSPM_PUBLIC_IP=$(terraform output -raw ec2_cspm_public_ip)
-          echo "::add-mask::$CSPM_PUBLIC_IP"
-          echo "CSPM_PUBLIC_IP=$CSPM_PUBLIC_IP" >> $GITHUB_ENV
+        run: ../../.ci/scripts/set_environment_output.sh
```
oren-zohar, line 261: Is this a generic script that will be used in other workflows? If not, I would use a less generic name

dev-docs/Cloud-Env-Testing.md
oren-zohar, line 50: Can you add example supported values for each option? Just to make sure the format is clear for users

.github/actions/kibana-ftr/README.md
```diff
@@ -0,0 +1,76 @@
...
+      - name: Check out the repo
+        uses: actions/checkout@v4
+
+      - name: Run Kibana Tests Action
+        uses: ./.github/actions/kibana-ftr
+        with:
+          ref_commit_sha: ${{ github.event.inputs.ref_commit_sha || 'main' }}
+          test_kibana_url: ${{ secrets.TEST_KIBANA_URL }} # required
+          test_es_url: ${{ secrets.TEST_ES_URL }} # required
+          es_version: 8.15.0-SNAPSHOT # required
```
oren-zohar, line 46: why is this hard-coded? Or is it only for the example?

.github/actions/kibana-ftr/README.md
```diff
@@ -0,0 +1,76 @@
...
+          test_browser_headless: '1'
```
oren-zohar, line 47: why are we using `'1'` as a boolean?

oren-zohar (changes requested):

.github/actions/kibana-ftr/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    default: 'main'
+  test_browser_headless:
+    description: 'Whether to run tests in headless mode'
+    required: false
+    default: '1'
+  test_cloud:
+    description: 'Whether to enable cloud testing'
+    required: false
+    default: '1'
+  es_security_enabled:
```
orouz, line 26: could we remove this and go with a default of `true` ? is there a use case for running it with `false` ?

.github/actions/kibana-ftr/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    description: 'URL for the ES instance'
+    required: true
+  es_version:
+    description: 'Version of Elasticsearch to test against'
+    required: true
+  ref_commit_sha:
+    description: 'Kibana PR commit sha'
+    required: false
+    default: 'main'
+  test_browser_headless:
```
orouz, line 18: isn't it headless by default, since it's running from CI? not sure how we'd access a browser here

.github/actions/kibana-ftr/action.yml
```diff
@@ -0,0 +1,79 @@
...
+description: 'Run Kibana tests using FTR'
+
+inputs:
+  test_kibana_url:
+    description: 'URL for the Kibana instance to test'
+    required: true
+  test_es_url:
+    description: 'URL for the ES instance'
+    required: true
+  es_version:
```
orouz, line 11: i think this should just be the kibana version we're running, as kibana/elasticsearch version should match. the kibana version is available at `package.json#version`. maybe add a checkbox instead to indicate if this is a pre-released version

dev-docs/Cloud-Env-Testing.md
gurevichdmitry: Linked to the `kibana-ftr` action documentation

dev-docs/Cloud-Env-Testing.md
gurevichdmitry: changed to `kibana_ref`

gurevichdmitry: added examples for `branch`, `tag` and `commit SHA`.

.github/actions/kibana-ftr/README.md
gurevichdmitry: `es_version` in the README is just an example.

.github/workflows/test-environment.yml
gurevichdmitry: renamed to `set_cloud_env_params`

.github/actions/kibana-ftr/README.md
gurevichdmitry: This originates from the FTR runner, where setting the environment variable to '1' activates headless mode. However, as per @orouz's suggestion, I'll eliminate this input variable. Since all tests are executed in headless mode on the CI, no further adjustments are necessary.

.github/actions/kibana-ftr/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    default: '1'
+  test_cloud:
+    description: 'Whether to enable cloud testing'
+    required: false
+    default: '1'
+  es_security_enabled:
+    description: 'Whether Elasticsearch security is enabled'
+    required: false
+    default: '1'
+  node_version_file:
```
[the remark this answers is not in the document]
gurevichdmitry: Dropped the input parameter and updated the `node-version-file` to use the `package.json` file.

.github/actions/kibana-ftr/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    description: 'Path to the Node.js version file'
+    required: false
+    default: 'package.json'
+  config:
```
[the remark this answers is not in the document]
gurevichdmitry: Currently, only `config.cloud.ts` is used. Changes to test suites will be managed in `cloud_tests/index.ts` or through tags. Thus, this input parameter is being dropped.

.github/actions/kibana-ftr/action.yml
gurevichdmitry: You're correct. I'll set `TEST_BROWSER_HEADLESS: '1'` by default and remove the input parameters. Since we're running from CI, accessing a browser directly isn't feasible.

.github/actions/kibana-ftr/action.yml
gurevichdmitry: The `create-environment` workflow may deploy three types of versions: currently in development, pre-release, and released versions. According to the [documentation (3. Custom option)](https://www.elastic.co/guide/en/kibana/current/development-tests.html#_running_functional_tests), _you must run the same branch of tests as the version of Kibana you’re testing. To run against a previous minor version, use the --es-version <instance version> option._ Therefore, we need to pass the version string explicitly. If we replace this with a flag, additional logic would be required in the kibana-ftr action to calculate the version. Since the create-environment workflow already knows this version and provides it as an input, I believe it's more efficient to pass the version string directly rather than using a flag.

.github/actions/kibana-ftr/action.yml
```diff
@@ -0,0 +1,79 @@
...
+    required: true
+  ref_commit_sha:
+    description: 'Kibana PR commit sha'
+    required: false
+    default: 'main'
+  test_browser_headless:
+    description: 'Whether to run tests in headless mode'
+    required: false
+    default: '1'
+  test_cloud:
```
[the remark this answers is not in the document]
gurevichdmitry: According to the [FTR documentation](https://www.elastic.co/guide/en/kibana/current/development-tests.html#_running_functional_tests), when running against an Elastic Cloud instance, additional environment variables are required: `TEST_CLOUD` and `ES_SECURITY_ENABLED`. The `input.test_cloud` environment variable should be configured for our case. The `input.config` variable instructs FTR which test configuration to execute. Currently, it will only execute `config.cloud.ts`, so we can omit both input variables for now and extend to work with different test configurations later.

.github/actions/kibana-ftr/action.yml
gurevichdmitry: Removed from the input and set to '1' in the action step.

gurevichdmitry:
[This is the latest run](https://github.com/elastic/cloudbeat/actions/runs/9244239868/job/25429313625) after addressing review comments.

oren-zohar (approved):

Issue #2074, opened 2024-04-01
Account for rate limiting in AWS fetchers


orouz:
**Motivation**
we need to account for rate limiting in our AWS fetchers to avoid losing resources we want to evaluate.

**Definition of done**
- figure out the quotas for each AWS fetcher 
- when applicable, AWS fetchers method usage does not exceed the default quota 
- add a retry (with backoff) mechanism for failed requests

**Out of scope** 
- synced cloudbeat instances consuming the same quota

orouz:
turns out AWS Go SDK v2 comes with a built-in retry mechanism. verified by modifying the aws clients config to include logging:
```
awsConfig.Logger = logging.NewStandardLogger(os.Stdout)
awsConfig.ClientLogMode = aws.LogRetries | aws.LogRequestWithBody | aws.LogResponseWithBody 
```

the important log we get to see is:
```
Amz-Sdk-Request: attempt=1; max=3
```

which corresponds to the default settings of the [Standard](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html) retry of 3 attempts. 

regarding quotas for AWS APIs, there doesn't seem to be any publicly defined general limits nor does the client tries to take that into account, and looks like retrying-with-backoff is the approach taken. 


in terms of modifications to cloudbeat, i think adding `429` HTTP status code to [Retryable Errors](https://github.com/aws/aws-sdk-go-v2/blob/7ed7b4fc6595a7e554796424f3470b37287dde5c/aws/retry/standard.go#L44-L49) would be sufficient to ensure we retry when APIs are rate limited (this is a safe-measure addition, as a similar retry condition is already [present](https://github.com/aws/aws-sdk-go-v2/blob/7ed7b4fc6595a7e554796424f3470b37287dde5c/aws/retry/standard.go#L65))

adding the additional retry condition, should also be done with consolidation of AWS config initialization as we're currently doing it in 2 ways:
1. using the config provider, which uses libbeat
2. directly using libbeat

kubasobon:
Verified, no rate limit messages in long-living environment's log.

PR #2223, opened 2024-05-21
extend standard retry with 429 http status code

orouz:
### Summary of your changes

this PR:
1. adds a [Standard Retryer](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html) 
   - it seems that initializing AWS config via libbeat's `InitializeAWSConfig` function ends up returning an `aws.Config` with `Retryer` as `nil`, so without adding it we're not opting-in to the default retry mechanism 
3. adds `429` as an additional, retryable HTTP status code.

### Implementation notes

the first commit of this PR can be considered an alternative, standalone implementation of doing the same thing the 2nd commit does. the difference between the two, is that the 2nd commit removes the `ConfigProvider` in favor of a plain function to encapsulate AWS config initialization (via libbeat) + adding a retryer. to do so, some code was removed and now all AWS config initializations call the same function, and the only consumer of the now-deleted AWS config provider (EKS) sets the region manually (which is what the provider did) 

all in all - the 2nd commit's diff is bigger, but i think it does some nice cleanup too. although, we could go with the 1st if preferable. 

### Related Issues
 - closes https://github.com/elastic/cloudbeat/issues/2074

### Checklist


internal/resources/providers/awslib/config.go
```diff
@@ -18,30 +18,28 @@
...
-		return nil, err
+
+	awsConfig.Retryer = func() aws.Retryer {
+		return retry.NewStandard(func(o *retry.StandardOptions) {
+			o.Retryables = append(o.Retryables, retry.RetryableHTTPStatusCode{
+				Codes: map[int]struct{}{
+					http.StatusTooManyRequests: {},
+				},
+			})
+		})
```
orouz, line 41: this is essentialy the same as [some](https://github.com/aws/aws-sdk-go-v2/blob/8209abb7fa1aeb513228b4d8c1a459aeb6209d4d/credentials/endpointcreds/internal/client/client.go#L68-L74) AWS client does manually

internal/resources/providers/awslib/config.go
```diff
@@ -18,30 +18,28 @@
...
 	}
-	awsConfig.Region = metadata.Region
```
orouz, line 44: all our AWS config initializations are done via libbeat. the only extra thing the config provider did was set the region. it was only used by EKS, which now sets the region [itself](https://github.com/elastic/cloudbeat/pull/2223/files#diff-f297d6b58da94a9a8f8e33eed64ebdc64076afdf856ccd3d81e52e4b08225734R118)

internal/flavors/benchmark/eks_test.go
```diff
@@ -59,36 +58,18 @@ func TestEKS_Initialize(t *testing.T) {
...
-			cfg:                    awsCfg,
-			wantErr:                "some error",
-		},
 		{
 			name:                   "aws identity provider error",
-			awsCfgProvider:         mockAwsCfg(nil),
 			awsIdentityProvider:    mockAwsIdentityProvider(errors.New("some error")),
 			clientProvider:         mockKubeClient(nil),
-			awsMetadataProvider:    mockMetadataProvider(errors.New("not this error")),       // ignored
+			awsMetadataProvider:    mockMetadataProvider(nil),
```
orouz, line 65: this needs to not err because EKS now sets the region itself via the metadata provider

moukoublen (approved):

Issue #2226, opened 2024-05-23
Cloudbeat failed to get cluster name when using Karpenter


therealdwright:
**Describe the bug**
When using Cloudbeat and the KSPM integration to scan our cluster that deploys workers using Karpenter, the agent logs the error:

```
failed to get cluster name: failed to get cluster name from the Auto-scaling group: cluster name not found from autoscaling groups
```

This is because the cluster tag is `aws:eks:cluster-name` for Karpenter managed nodes. The check performed in the [cluster name provider](https://github.com/elastic/cloudbeat/blob/3dfeb066e98837df0c05017d6e4d9018b90dffea/internal/resources/providers/awslib/cluster_name_provider.go#L87) is a little too strict and could benefit from being a substring check.
 
**Preconditions**
EKS Cluster using Karpenter Node Provisioner

PR #2227, opened 2024-05-23
Use substring check for EKS cluster name provider

therealdwright:
### Summary of your changes
Karpenter managed clusters will use `aws:eks:cluster-name` for its
cluster name tag. This commit updates the logic to use a substring
check so both the existing tag of `eks:cluster-name` and karpenter
are supported.

### Related Issues
- Fixes: https://github.com/elastic/cloudbeat/issues/2226

### Checklist


moukoublen:
Hello @therealdwright, and thanks for opening the issue and the PR,

One minor concern: Would it be safer to compare using `ends with` instead of `contains`? 

I am not sure if it's actually needed. I just thought of cases where tags could perhaps contain `eks:cluster-name` in the name but refer to a different thing.

therealdwright:
I think this is a fair criticism and I am only personally interested in ends with but was trying to be robust but happy to boil this one down to YAGNI.

moukoublen:
LGTM

Thanks @therealdwright!

therealdwright:
Thanks @moukoublen - would you have any idea which elastic-agent release this will be in?

moukoublen:
@therealdwright most probably `8.15.0`.

Issue #2167, opened 2024-05-01
Fix bump bump version workflow


orouz:
**Motivation**

latest release the workflow [failed](https://github.com/elastic/cloudbeat/actions/runs/8709394933) to complete due to a git error where the workflow assumes it'll be the only one updating hermit, but hermit was updated manually in a different [PR](https://github.com/elastic/cloudbeat/pull/2000/files) 

**Definition of done**
- [ ] make sure the workflow skips a steps if it's not needed

orouz:
TODO: we should also add automation for bumping the new [CF credentials templates](https://github.com/elastic/integrations/pull/9828/files#diff-72fc27a4cf2fb28244c8fa9ed27c00e43935777a4875c5d61d91372f0d483425R127)

PR #2234, opened 2024-05-28
fix bump workflow git errors

orouz:
### Summary of your changes

the version bump workflow does the following:
1. cloudbeat
   - PR to update ARM templates, version.go, mergify (to `main`)
   - PR to update hermit (to `main`)
   - PR to update ARM templates (to `release-branch`)
   - upload CF templates

2. integrations
    - PR to bump version and update template URLs versions (arm, CF, cloudshell)

in the previous release, only the 1st cloudbeat PR was opened successfully, because the 2nd step - updating hermit, failed when the script tried to commit the changes to hermit but there were none, as it was already updated manually, and since the script has `set -e` it exited and didn't run any other step. 

the changes in this PR avoid committing if there are no changes, and by that avoid exiting due to errors and failing to complete the other steps. there's also some minor formatting changes. 

### Related Issues
- fixes https://github.com/elastic/cloudbeat/issues/2167


scripts/bump_cloudbeat.sh
```diff
@@ -21,8 +21,16 @@ echo "NEXT_CLOUDBEAT_VERSION: $NEXT_CLOUDBEAT_VERSION"
 echo "CURRENT_CLOUDBEAT_VERSION: $CURRENT_CLOUDBEAT_VERSION"
 echo "CURRENT_MINOR_VERSION: $CURRENT_MINOR_VERSION"
 
+# clear branches if they exists
+branches=("$NEXT_CLOUDBEAT_BRANCH" "$NEXT_CLOUDBEAT_HERMIT_BRANCH" "$RELEASE_CLOUDBEAT_BRANCH")
+for branch in "${branches[@]}"; do
+    if git ls-remote --exit-code --heads origin "$branch"; then
+        git push origin --delete "$branch"
+    fi
+done
```
orouz, line 30: the script is intended to be idempotent so if we run it multiple times, the result will be the same. to do so, we remove any existing branches (which will automatically close open PRs)

the thinking here was that if the script ran the first time and partially failed, for example - the user ran it before branching out so the release branch didn't exist, then after they create the release branch (or fix any other error, they can run it again.

scripts/bump_cloudbeat.sh
```diff
@@ -39,65 +47,78 @@ update_version_mergify() {
...
 EOF
     git add .mergify.yml
-    git commit -m "Update .mergify.yml"
-
-    gh label create "backport-v$CURRENT_CLOUDBEAT_VERSION"
+    if git diff --cached --quiet; then
+        echo "No changes to commit in .mergify.yml"
+    else
+        git commit -m "Update .mergify.yml"
+        gh label create "backport-v$CURRENT_CLOUDBEAT_VERSION" --force
```
orouz, line 54: we need `--force` in case the script runs a 2nd time

scripts/bump_integration.sh
```diff
@@ -24,37 +31,41 @@ get_next_integration_version() {
...
-    echo "• Update cloudshell_git_branch in manifest.yml"
+    echo "Update cloudshell_git_branch in manifest.yml"
     sed -i'' -E "s/cloudshell_git_branch=[0-9]+\.[0-9]+/cloudshell_git_branch=$MAJOR_MINOR_CLOUDBEAT/g" $MANIFEST_PATH
 
     # cis_aws + vuln_mgmt_aws
-    echo "• Update cloudformation-* in manifest.yml"
+    echo "Update cloudformation-* in manifest.yml"
     sed -i'' -E "s/cloudformation-cnvm-[0-9]+\.[0-9]+\.[0-9]+/cloudformation-cnvm-$CURRENT_CLOUDBEAT_VERSION/g" $MANIFEST_PATH
     sed -i'' -E "s/cloudformation-cspm-ACCOUNT_TYPE-[0-9]+\.[0-9]+\.[0-9]+/cloudformation-cspm-ACCOUNT_TYPE-$CURRENT_CLOUDBEAT_VERSION/g" $MANIFEST_PATH
+    sed -i'' -E "s/cloudformation-cspm-direct-access-key-ACCOUNT_TYPE-[0-9]+\.[0-9]+\.[0-9]+/cloudformation-cspm-ACCOUNT_TYPE-$CURRENT_CLOUDBEAT_VERSION/g" $MANIFEST_PATH
```
orouz, line 50: this was added to update the [new credentials deployment variable](https://github.com/elastic/integrations/pull/9828/files#diff-72fc27a4cf2fb28244c8fa9ed27c00e43935777a4875c5d61d91372f0d483425R127)

scripts/bump_cloudbeat.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euo pipefail
+set -xeuo pipefail
```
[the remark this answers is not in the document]
uri-weisman: We need to be aware we do not log sensitive data, e.g API keys.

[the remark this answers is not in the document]
orouz: good point! 💯 

i've updated it to remove logging when api keys are used (only once, when uploading cloud formation templates)

oren-zohar (changes requested):
I have to say, so docstring around the functions would be helpful here, it's not a small script anymore and we would like others to easily maintain it in the future

scripts/bump_cloudbeat.sh
```diff
@@ -39,65 +47,78 @@ update_version_mergify() {
...
     git add .mergify.yml
-    git commit -m "Update .mergify.yml"
-
-    gh label create "backport-v$CURRENT_CLOUDBEAT_VERSION"
+    if git diff --cached --quiet; then
+        echo "No changes to commit in .mergify.yml"
+    else
+        git commit -m "Update .mergify.yml"
+        gh label create "backport-v$CURRENT_CLOUDBEAT_VERSION" --force
+    fi
```
oren-zohar, line 55: Can we use `--only` to simplify the statement?
```suggestion
    git commit -m "Update .mergify.yml" --only .mergify.yml && gh label create "backport-v$CURRENT_CLOUDBEAT_VERSION" --force

```

orouz: no, we can't run `git commit` when there are no changes, as this will return an error code which will exit due to `set -e`, that's why it's wrapped in an if/else

scripts/bump_integration.sh
```diff
@@ -10,10 +10,20 @@ MAJOR_MINOR_CLOUDBEAT=$(echo "$CURRENT_CLOUDBEAT_VERSION" | cut -d. -f1,2)
 export MAJOR_MINOR_CLOUDBEAT
 
 checkout_integration_repo() {
-    echo "• Checkout integration repo"
+    echo "Checkout integration repo"
     gh auth setup-git
     gh repo clone $INTEGRATION_REPO
     cd integrations
+
+    # clear branch if it exists
```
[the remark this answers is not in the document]
orouz: it's just easier to delete all branches before executing the rest of the script

fetching/rebasing a non existing branch also results in an error so we'd need to account for it too. 

this is cleaner IMO

orouz:
> I have to say, so docstring around the functions would be helpful here, it's not a small script anymore and we would like others to easily maintain it in the future

anything in specific that is not clear that could use a comment with info that isn't reflected in the function name?

oren-zohar (approved):

PR #2241, opened 2024-05-31
Implement AWS VPC fetcher with sub-resources

kubasobon:
### Summary of your changes
Introduce AWS VPC fetcher for Asset Inventory to retrieve the following VPC resources:

- Internet Gateways
- NAT Gateways
- Network ACLs
- Network Interfaces
- Security Groups
- Subnets
- Transit Gateway Attachments
- Transit Gateways
- VPCs
- VPC Peering Connections

### Screenshot/Data
Tested:
image

### Related Issues
Fixes https://github.com/elastic/security-team/issues/9085

### Checklist

#### Introducing a new rule?
no


romulets (approved):
The PR looks good to me!

The only question that I have, is more from a product perspective. What is really the useful data? What is the use case we are levaraging?

I wonder if we need to have Transit Gateways and Transit Gateway Attachments as its own entity for example.

I have a feeling that we are going towards storing everything but without really considering what is relevant to be stored as a entity (es document).

Btw, what you did is correct, specially looking at [the list @tinnytintin10 crafted](https://docs.google.com/spreadsheets/d/1TO4EtqTYcaW-yPCLcTVkp2PC97HYLwCUz6rvSAfwmtI/edit#gid=731107840).

Concrete action: I suggest we take a step back to discuss a bit the goal of what we are doing from a product perspective. Is storing as much as we can indeed, or do we have use cases to leverage already? 

That doesn't necessarily blocks this from being merged. We can change it up later. You did what the task requested from you.

I just question the end usefulness.

PR #2245, opened 2024-06-02
Remove periodic-ci workflow

gurevichdmitry:
### Summary of your changes

This PR removes the unused workflow `periodic-ci.`

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


orouz:
@gurevichdmitry i think we can remove the `Build` command on the magefile, as it was only used in this workflow. maybe even some more magefile cleanup would be possible because of it.

gurevichdmitry:
> @gurevichdmitry i think we can remove the `Build` command on the magefile, as it was only used in this workflow. maybe even some more magefile cleanup would be possible because of it.

@orouz, I removed the `Build` command and the `build.sh` script, which had no references. Additionally, I updated the documentation where mage build was referenced.

orouz (approved):

PR #2266, opened 2024-06-10
add new wf

gurevichdmitry:
### Summary of your changes

This PR introduces a new scheduled workflow to automatically destroy expired environments. Additionally, the existing workflow has been updated to support this functionality. The` Create Environment` workflow now includes a new input parameter, `expiration_days`, which specifies the number of days until the environment expires. This date is calculated and stored in the `env_config.json` file.

The `Destroy Expired Environments` workflow comprises two jobs: one for identifying expired environments and another for executing the destroy process. Due to limitations in the GCP environment deletion script, the destroy process is currently set to run sequentially.

### Screenshot/Data

**Expiration Days**

Screenshot 2024-06-19 at 9 04 01

**Logging environments**

Screenshot 2024-06-19 at 9 12 29

**Destroy Expired Environment WF**
Screenshot 2024-06-19 at 9 13 23

Screenshot 2024-06-19 at 9 04 54

### Checklist


.github/workflows/test-environment.yml
```diff
@@ -33,10 +33,11 @@ on:
...
           type: boolean
-      run-ui-sanity-tests:
-          description: "Run UI sanity tests after provision"
-          default: false
-          type: boolean
+      # TODO: Reached max number of inputs. May be merged with run-sanity-tests
+      # run-ui-sanity-tests:
+      #     description: "Run UI sanity tests after provision"
+      #     default: false
+      #     type: boolean
```
gurevichdmitry, line 40: Since we encountered the issue of exceeding the [maximum number of input parameters](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs) in the `Create Environment` workflow, we need to consider removing or merging existing parameters to accommodate the new `expiration_days` input. After reviewing all input parameters, I propose the following options:
- Merging `run-sanity-tests` and `run-ui-sanity-tests`: This could be effective since both are used when entering the QA cycle, and marking one checkbox might suffice. However, it could be slightly problematic when creating new tests and wanting to debug only UI or integration tests.

- Removing `cleanup-env`: This parameter is used mainly to check the version before the QA cycle and has less usage in my opinion.

- Removing `ec-api-key`: This allows deployment on a private cloud account. Given that we are using a common organization account, I am not sure how useful this parameter is.

.github/workflows/destroy-expired-environment.yml
```diff
@@ -0,0 +1,59 @@
...
+on:
+  # TODO: check if workflow_dispatch is needed
+  workflow_dispatch:
+  # TODO: remove pull_request after debugging
+  pull_request:
+    branches:
+      - main
+      - "[0-9]+.[0-9]+"
+    types: [opened, synchronize, reopened]
+
```
gurevichdmitry, line 12: I'm using this option to run the workflow temporarily. After the review and necessary fixes, it will be removed.

kubasobon (approved):

.github/workflows/destroy-expired-environment.yml
```diff
@@ -0,0 +1,59 @@
+name: Destroy Expired Environment
+
+on:
+  # TODO: check if workflow_dispatch is needed
+  workflow_dispatch:
```
oren-zohar, line 5: I don't think this is needed

.github/workflows/destroy-expired-environment.yml
```diff
@@ -0,0 +1,59 @@
...
+  # TODO: remove pull_request after debugging
```
oren-zohar, line 6: don't forget

.github/workflows/destroy-expired-environment.yml
```diff
@@ -0,0 +1,59 @@
...
+    needs: ["envs_to_destroy"]
+    secrets: inherit
+    # Required for the 'Destroy' job in the 'destroy-environment.yml'
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    if: ${{ needs.envs_to_destroy.outputs.env_found == 'true' }}
+    strategy:
+      fail-fast: false
+      max-parallel: 1 # Destroy environments one by one
```
oren-zohar, line 55: why?

.github/workflows/test-environment.yml
oren-zohar: I'm in favor of getting rid of `cleanup-env` since the TTL makes it redundant

oren-zohar (commented):
Looks good, please don't forget to address the TODOs before merging

.github/workflows/destroy-expired-environment.yml
gurevichdmitry: Parallel execution might be problematic for GCP agent deletion due to [binding and unbinding of policies](https://github.com/elastic/cloudbeat/blob/main/deploy/test-environments/delete_gcp_env.sh#L17-L37) using the same GCP account. If two independent runs execute the same delete environment script, it may cause failures due to policy conflicts. Sequential run avoids this problem and might be sufficient since we don’t deploy many environments.

.github/workflows/destroy-expired-environment.yml
gurevichdmitry: removed

PR #2277, opened 2024-06-18
Add AWS SNS to Asset Inventory

kubasobon:
### Summary of your changes

Add an AWS SNS fetcher for Topics + Subscribtions.

### Related Issues

Closes https://github.com/elastic/security-team/issues/9086

### Checklist
- ~~I have added the necessary README/documentation (if appropriate)~~

#### Introducing a new rule?

no


internal/inventory/asset.go
```diff
@@ -37,9 +37,10 @@ type AssetSubCategory string
 const (
 	SubCategoryAuthorization AssetSubCategory = "authorization"
 	SubCategoryCompute       AssetSubCategory = "compute"
-	SubCategoryStorage       AssetSubCategory = "storage"
 	SubCategoryDatabase      AssetSubCategory = "database"
+	SubCategoryIntegration   AssetSubCategory = "integration"
```
romulets, line 41: I'm not completely sure about this sub-category. I personally would call it `messaging`. But maybe just a quick check with other engineers might solve our question :D

internal/resources/providers/awslib/sns/topic.go
```diff
@@ -0,0 +1,50 @@
...
+	region        string
+}
+
+func (v TopicInfo) GetResourceArn() string {
+	return pointers.Deref(v.Topic.TopicArn)
+}
+
+func (v TopicInfo) GetResourceName() string {
+	elems := strings.Split(v.GetResourceArn(), ":")
+	return elems[len(elems)-1]
```
romulets, line 41: There is a unlike possibility of this arn being empty or without a `:` which would cause a panic. Maybe just a quick check if `len(elems) > 0` is wise?

internal/inventory/asset.go
kubasobon: Yeah, I agree. I'm up for a discussion. Any recommendations welcome.

kubasobon: @tinnytintin10, any preference towards the classification of SNS Topics/Subscriptions?

internal/resources/providers/awslib/sns/topic.go
kubasobon: Good catch, fixed.

internal/inventory/asset.go
romulets: I vote for `messaging`. How I'd decide it: have a read on how analog techs (pubsub, kafka, rabbitmq and others) are called. If there is a common name, let's use what the market is using

romulets (approved):

Issue #2289, opened 2024-06-24
Supporting integration namespaces suffix


oren-zohar:
**Motivation**
Following:
- https://github.com/elastic/cloudbeat/issues/2229

Cloudbeat currently [hard-code](https://github.com/elastic/cloudbeat/blob/d4812c8710ef67045840df553bc8a03a950ba34f/internal/config/config.go#L40) the data stream namespace. We need to get the namespace config and update the data stream accordingly 

**Definition of done**
What needs to be completed at the end of this task
- [ ] Update cloudbeat's data-stream according to the integration's config

**Out of scope**
What should not be included in this task

**Related tasks/epics**
- https://github.com/elastic/security-team/issues/9558

kubasobon:
@kfirpeled I think so. Currently testing to confirm it works for CNVM as well.

kubasobon:
@kfirpeled I checked, and it needed some extra code changes, but I have achieved the same result for CNVM. Tested, screenshots in the PR.

kubasobon:
@oren-zohar Please find the results of me testing various namespace settings 

`agent_policy.namespace` | `package_policies[].namespace` | `data_stream.namespace` result
---|---|---
`default` | `default` | `default`
`default` | `""` | `default`
`default` | `test` | `test`
`test` | `default` | `default`
`test` | `""` | `test`
`test` | `test` | `test`

Conclusions:
1. Package Policy namespace overrides Agent Policy namespace setting
2. If Package Policy namespace is left empty, it uses Agent Policy namespace's value

PR #2291, opened 2024-06-25
Add support for namespace suffixes

kubasobon:
### Summary of your changes

This pull request allows Cloudbeat to publish events to non-default namespaces. For example, findings are sent to the `logs-cloud_security_posture.findings-{namespace}` Data Stream. Until now, the `{namespace}` always defaulted to `default`, regardless of Agent Policy. After merging this change, Cloudbeat will respect the Agent Policy and set publish events to the appropriate stream.

### Screenshot/Data

<details> <summary>CSPM using non-default namespace</summary>
Screenshot 2024-06-25 at 13 26 35
Screenshot 2024-06-25 at 13 25 40
</details>

<details> <summary>Non-default Data Streams present when running modified cloudbeat</summary>
Screenshot 2024-06-25 at 13 07 23
</details>

<details> <summary>Findings published to the new Data Stream successfully</summary>
Screenshot 2024-06-25 at 13 08 39
</details>

EDIT:

<details> <summary>Vulnerabilities published to the new Data Stream successfully</summary>
Screenshot 2024-06-25 at 16 57 24
</details>

### Related Issues

Closes https://github.com/elastic/cloudbeat/issues/2289

### Checklist
- ~~I have added the necessary README/documentation (if appropriate)~~


internal/config/config.go
```diff
@@ -55,6 +56,7 @@ type Config struct {
 	BundlePath            string                  `config:"bundle_path"`
 	PackagePolicyId       string                  `config:"package_policy_id"`
 	PackagePolicyRevision int                     `config:"revision"`
+	Index                 string                  `config:"index"`
```
kubasobon, line 59: Our integration has the [correct ECS `data_stream` fields](https://www.elastic.co/guide/en/ecs/current/ecs-data_stream.html) [defined](https://github.com/elastic/integrations/blob/317769902bd73c80c9de621f364d2de6547c2322/packages/cloud_security_posture/data_stream/findings/fields/base-fields.yml), and we could use those to build the index, like so `{data_stream.type}-{data_stream.dataset}-{data_stream.namespace}`.
However, the `namespace` field is empty for current release of the integration. The Policy/Config also comes with the `index` field, which contains a ready-to-use index name in the correct format. 

@kfirpeled Do you know where does `index` field come from and is it reliable enough for us to use?

internal/config/config.go
```diff
@@ -34,10 +34,11 @@ import (
...
 
 const (
-	DefaultNamespace             = "default"
-	VulnerabilityType            = "vuln_mgmt"
-	AssetInventoryType           = "asset_inventory"
-	ResultsDatastreamIndexPrefix = "logs-cloud_security_posture.findings"
+	DefaultNamespace                = "default"
+	VulnerabilityType               = "vuln_mgmt"
+	AssetInventoryType              = "asset_inventory"
+	defaultFindingsIndexPrefix      = "logs-cloud_security_posture.findings"
```
romulets, line 40: Why is this called default it doesn't mention any namespace?

internal/config/config.go
romulets: How I understand is that a integration has the permissions to to publish to a specific index. See the policy template once you install it. So essentially the integration already knows where you can publish to. The index template is decided by the integration

romulets (approved):

internal/config/config.go
kubasobon: Yeah, that I know :) I wanted to know if there is anything preventing us from relying on the `index` field.

internal/config/config.go
kubasobon: This is why it's a prefix. The whole index is constructed as `{prefix}-{namespace}`. The prefix can vary - see the prefix for Vulnerability events.

internal/config/config.go
kfirpeled: I'm not familiar with `index` in the agent config. I'm only familiar with agent's policy.

Here's a snippet:

```yaml
inputs:
  - id: logfile-system-1c2284c7-c5ae-481d-a7bc-d9e3ce61545a
    name: system-3
    revision: 1
    type: logfile
    use_output: default
    meta:
      package:
        name: system
        version: 1.58.1
    data_stream:
      namespace: default
    package_policy_id: 1c2284c7-c5ae-481d-a7bc-d9e3ce61545a
    streams:
      - id: logfile-system.auth-1c2284c7-c5ae-481d-a7bc-d9e3ce61545a
        data_stream:
          dataset: system.auth
          type: logs
```

So I can only guess from here how it is being done on the beats side

internal/config/config.go
kfirpeled: Here's an example what is being sent for CNVM

```yaml
  - id: cloudbeat/vuln_mgmt_aws-vuln_mgmt-6921c0f7-ae37-4a1b-910b-f158a7c1af64
    name: cnvm
    revision: 1
    type: cloudbeat/vuln_mgmt_aws
    use_output: default
    meta:
      package:
        name: cloud_security_posture
        version: 1.9.0
    data_stream:
      namespace: default
    package_policy_id: 6921c0f7-ae37-4a1b-910b-f158a7c1af64
    streams:
      - id: >-
          cloudbeat/vuln_mgmt_aws-cloud_security_posture.vulnerabilities-6921c0f7-ae37-4a1b-910b-f158a7c1af64
        data_stream:
          dataset: cloud_security_posture.vulnerabilities
          type: logs
        period: 24h
```

So the datastream is built by `<type>-<dataset>-<namespace>`

kfirpeled:
It would be best to approach `#fleet` or find some documentation regardless beats and how send events to datastreams and reading the agent policy. I'm not familiar with the best practices

PR #2374, opened 2024-07-31
Introduce Azure Asset Inventory and fetch Azure Account resources

kubasobon:
### Summary of your changes

This PR enables Asset Inventory flavor to handle a second cloud provider - Azure. It also adds a fetcher for the first two resources: Azure Tenants and Azure Subscriptions. 

Changes to config (`config.v1.asset_inventory_provider`) will be addressed in a PR to elastic/integrations: [the PR](https://github.com/elastic/integrations/pull/10663).

### Screenshot/Data
Screenshot 2024-07-31 at 11 47 23

_Kibana displaying one Tenant and one Subscription resource_

### Related Issues

Closes https://github.com/elastic/security-team/issues/10159

### Checklist


internal/config/config.go
```diff
@@ -37,6 +37,9 @@ const (
 	DefaultNamespace                = "default"
 	VulnerabilityType               = "vuln_mgmt"
 	AssetInventoryType              = "asset_inventory"
+	ProviderAWS                     = "aws"
+	ProviderAzure                   = "azure"
+	ProviderGCP                     = "gcp"
```
romulets, line 42: Just for sake of "consistency", maybe we should have this values inside of a provider file, like benchmark has?

https://github.com/elastic/cloudbeat/blob/main/internal/config/benchmark.go

internal/config/config.go
```diff
@@ -46,17 +49,18 @@ type Fetcher struct {
...
-	Period                time.Duration           `config:"period"`
-	Processors            processors.PluginConfig `config:"processors"`
-	BundlePath            string                  `config:"bundle_path"`
-	PackagePolicyId       string                  `config:"package_policy_id"`
-	PackagePolicyRevision int                     `config:"revision"`
-	Index                 string                  `config:"index"`
+	Benchmark              string                  `config:"config.v1.benchmark"`
+	Type                   string                  `config:"config.v1.type"`
+	Deployment             string                  `config:"config.v1.deployment"`
+	AssetInventoryProvider string                  `config:"config.v1.asset_inventory_provider"`
```
romulets, line 55: I understand why you did this way.

From a first thought, I had in mind to re-use the `benchmark` field, that is used to define the "provider" on CSPM.

I know they are different concepts. But they somehow feel connected to me. It's a "sub selection" of the type.

Do you think there is a way to make a generic config here?

As I type I realize that to update `benchmark` config key could lead to few changes on integrations and maybe even kibana. So I'm not sure if it's a good idea. But I'm still posting so you give some thought too :)

internal/flavors/asset_inventory.go
```diff
@@ -49,13 +53,26 @@ func newAssetInventoryFromCfg(b *beat.Beat, cfg *config.Config) (*assetInventory
 	logger := logp.NewLogger("asset_inventory")
 	ctx, cancel := context.WithCancel(context.Background())
 
-	logger.Info("Creating AWS AssetInventory")
-
-	awsFetchers, err := initAwsFetchers(ctx, cfg, logger)
+	var fetchers []inventory.AssetFetcher
+	var err error
+
+	switch cfg.AssetInventoryProvider {
```
romulets, line 59: I do like how on cspm this is extracted to an strategy file. It's clean.

You don't need to implement it, I'm just pointing how cspm does. Maybe it influences you.

https://github.com/elastic/cloudbeat/blob/main/internal/flavors/benchmark/strategy.go

internal/inventory/asset.go
```diff
@@ -129,6 +134,9 @@ var (
 	AssetClassificationAwsRds                      = AssetClassification{Category: CategoryInfrastructure, SubCategory: SubCategoryDatabase, Type: TypeRelationalDatabase, SubType: SubTypeRDS}
 	AssetClassificationAwsS3Bucket                 = AssetClassification{Category: CategoryInfrastructure, SubCategory: SubCategoryStorage, Type: TypeObjectStorage, SubType: SubTypeS3}
 	AssetClassificationAwsSnsTopic                 = AssetClassification{Category: CategoryInfrastructure, SubCategory: SubCategoryMessaging, Type: TypeNotificationService, SubType: SubTypeSNSTopic}
+	// Azure
+	AssetClassificationAzureTenant       = AssetClassification{Category: CategoryInfrastructure, SubCategory: SubCategoryManagement, Type: TypeCloudAccount, SubType: SubTypeAzureTenant}
```
romulets, line 138: I find management a super broad term. Is that something suggested by Tin? Maybe a quick ping on him would be cool

internal/inventory/azurefetcher/azurefetchers_test.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/elastic/cloudbeat/internal/inventory"
+)
+
+func collectResourcesAndMatch(t *testing.T, fetcher inventory.AssetFetcher, expected []inventory.AssetEvent) {
```
romulets, line 30: neat.

internal/resources/providers/azurelib/inventory/susbcriptions_provider.go
```diff
@@ -54,14 +66,37 @@ func NewSubscriptionProvider(log *logp.Logger, credentials azcore.TokenCredentia
 		},
 	}
 
+	tenantClient := tenantAzureClientWrapper{
+		Tenants: func(ctx context.Context, clientOptions *arm.ClientOptions, options *armsubscriptions.TenantsClientListOptions) ([]armsubscriptions.TenantsClientListResponse, error) {
+			c, err := armsubscriptions.NewTenantsClient(credentials, clientOptions)
```
romulets, line 71: Can't we get tenants from the graph api?

internal/resources/providers/azurelib/inventory/susbcriptions_provider.go
```diff
@@ -54,14 +66,37 @@ func NewSubscriptionProvider(log *logp.Logger, credentials azcore.TokenCredentia
...
+			if err != nil {
+				return nil, err
+			}
+			return readPager(ctx, c.NewListPager(options))
+		},
+	}
+
+	subscriptionClient := subscriptionAzureClientWrapper{
+		Subscriptions: func(ctx context.Context, clientOptions *arm.ClientOptions, options *armsubscriptions.ClientListOptions) ([]armsubscriptions.ClientListResponse, error) {
+			cl, err := armsubscriptions.NewClient(credentials, clientOptions)
```
romulets, line 81: Can't we get subscriptions from the graph api?

internal/config/config.go
kubasobon: That's great, definitely want to keep magic strings in one place if possible. Will do!

internal/config/config.go
kubasobon: Yeah, I was refraining from `Benchmark` for integration's sake. Another way I was considering was using the `Type`, as you noticed. It's currently `asset_inventory`, but we could make it `asset_inventory_<provider>` and apply logic to that. WDYT?

internal/flavors/asset_inventory.go
kubasobon: Makes sense, especially if we go with the `Type` route.

internal/inventory/asset.go
kubasobon: Yeah, I'm pulling all classification names from [this spreadsheet](https://docs.google.com/spreadsheets/d/1TO4EtqTYcaW-yPCLcTVkp2PC97HYLwCUz6rvSAfwmtI/edit?gid=1300582090#gid=1300582090) word for word. We can talk on CDR sync maybe?

internal/resources/providers/azurelib/inventory/susbcriptions_provider.go
kubasobon: I checked and I haven't found a way to do it. If you know a way, it would be cool.
Here's the ticket for [the resources we can pull using Resource Graph](https://github.com/elastic/security-team/issues/10160). Everything else has its own separate tickets.

internal/resources/providers/azurelib/inventory/susbcriptions_provider.go
kubasobon: ditto

I checked, but haven't found subscriptions in Resource Graph API.

internal/inventory/azurefetcher/azurefetchers_test.go
kubasobon: Actually I meant to extract it to a common lib, since it's a copy of your code from AWS tests 😄

internal/resources/providers/azurelib/inventory/susbcriptions_provider.go
romulets: I do see a resource called `microsoft.aadiam/tenants`, but using the graph explorer in my count leads to 0 results. Not sure if it's because this resource is about something else or permission issues. Either way I hate how little documentation and context azure has.

So let's go with your call!

internal/resources/providers/azurelib/inventory/susbcriptions_provider.go
romulets: Same comment as to `tenants` I see a `microsoft.resources/subscriptions` resource, but without results nor context nor documentation 🫠

internal/inventory/azurefetcher/azurefetchers_test.go
romulets: I had forgotten had done this 😝

internal/config/config.go
orouz: i could see a separate issue where we use something like `Kind` to express both `Benchmark` and `AssetInventoryProvider` (as in - a flavor's sub-type). that'll require some `.hbs` file changes, cloudbeat back-compatbility config resolver, and if we really want to clean up stuff - some non-mandatory minor kibana changes 

although it might be too much trouble just for tidiness sake

romulets (approved):

PR #2405, opened 2024-08-11
Update VM labels in environment creation workflow

orouz:
this is required in order to avoid VMs being suspended/removed by cloud custodian

we used to tag the deployment itself and not the VM

labels after deploying using [workflow](https://github.com/elastic/cloudbeat/actions/runs/10339504200/job/28618936309#step:24:1): 
Screenshot 2024-08-11 at 15 24 47


.github/workflows/test-environment.yml
```diff
@@ -324,7 +324,7 @@ jobs:
         id: cspm-gcp-agent
         working-directory: deploy/deployment-manager
         run: |
-          . ./set_env.sh && ./deploy.sh && gcloud deployment-manager deployments update "${DEPLOYMENT_NAME}" --update-labels "${GCP_DEFAULT_TAGS}"
+          . ./set_env.sh && ./deploy.sh && gcloud compute instances update "${DEPLOYMENT_NAME}" --update-labels "${GCP_DEFAULT_TAGS}" --zone=us-central1-a
```
gurevichdmitry, line 327: It might be better to use the zone from environment variables. For example, you can define a global environment variable like this:
`AZURE_ZONE: "us-central1-a"`

Then, in your script, you can use the environment variable as follows:
`
. ./set_env.sh && ./deploy.sh && gcloud compute instances update "${DEPLOYMENT_NAME}" --update-labels "${GCP_DEFAULT_TAGS}" --zone="${AZURE_ZONE}"
`

PR #2427, opened 2024-08-13
Bump Trivy to v0.54.1

romulets:
### Summary of your changes

Bump trivy and fix breaking changes:

- Fix registry version https://github.com/aquasecurity/trivy/pull/6219; 
-  Fix replace zap with slog https://github.com/aquasecurity/trivy/pull/6466; 
    - The fix with slog used a zap to slog bridge (official from zap, but exp). It didn't have a license file, so I hardcoded a commit version that had; 
- Fix go clear cache https://github.com/aquasecurity/trivy/discussions/7010
- Fix --vuln-type flag renamed into --pkg-types https://github.com/aquasecurity/trivy/pull/7104; 
- Adopt package relationships https://github.com/aquasecurity/trivy/pull/7237

Tested on EC2 instance and https://github.com/elastic/cloudbeat/pull/2426 actions


kubasobon (approved):
Looks good. Freeing disk space on GH jobs seems like a necessity these days.

internal/vulnerability/runner.go
```diff
@@ -82,24 +86,12 @@ func clearTrivyCache(ctx context.Context, log *logp.Logger) error {
...
 
-	// These are the three available cli settings for clean/reset translated to flag.Options object.
-	// 	{CacheOptions: flag.CacheOptions{ClearCache: true}},
-	// 	{DBOptions: flag.DBOptions{Reset: true}},
-	// 	{MisconfOptions: flag.MisconfOptions{ResetPolicyBundle: true}},
-	// In our case we will use only the ClearCache option.
-
-	errs := make([]error, 0, 2)
-	r, err := artifact.NewRunner(ctx, flag.Options{CacheOptions: flag.CacheOptions{ClearCache: true}})
+	err := clean.Run(ctx, flag.Options{CleanOptions: flag.CleanOptions{CleanScanCache: true}})
```
moukoublen, line 89: I think it might be better to initialize `GlobalOptions.Timeout` to at least some seconds, because `clean.Run` creates a context with timeout ([link](https://github.com/aquasecurity/trivy/blob/08cc14bd2171afdc1973c6d614dd0d1fb82b7623/pkg/commands/clean/run.go#L19-L20))

internal/vulnerability/runner.go
```diff
@@ -82,24 +86,12 @@ func clearTrivyCache(ctx context.Context, log *logp.Logger) error {
...
 	if err != nil {
 		if !errors.Is(err, artifact.SkipScan) {
```
moukoublen, line 91: This is no longer needed; `clean.Run` will not return `artifact.SkipScan` because it's no longer a scan command. Se we perhaps could `return clean.Run` directly.

PR #2460, opened 2024-08-28
GCP Assets Inventory integration fetcher

orouz:
### Summary of your changes

adds a new `gcpfetcher` to assets inventory flavor. uses gcplib/inventory provider to fetch assets.

### Screenshot/Data
Screenshot 2024-08-28 at 17 49 26

### Related Issues
- depends on https://github.com/elastic/security-team/issues/10437
- closes https://github.com/elastic/security-team/issues/10434


internal/inventory/asset.go
orouz, line 1: we need to add vendor-specific types to account for roughly-same-thing-but-different-fields cases, like `AssetNetwork` -> `AwsAssetNetwork` / `GcpAssetNetwork`. i plan to do this in a future PR

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -0,0 +1,100 @@
...
+func newAssetsInventoryFetcher(logger *logp.Logger, provider inventoryProvider) inventory.AssetFetcher {
+	return &assetsInventory{
+		logger:   logger,
+		provider: provider,
+	}
+}
+
+func (f *assetsInventory) Fetch(ctx context.Context, assetChan chan<- inventory.AssetEvent) {
+	for _, r := range ResourcesToFetch {
+		f.fetch(ctx, assetChan, r.assetType, r.classification)
```
orouz, line 64: we might want to run this in a goroutine with a wait group. thoughts?

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -0,0 +1,100 @@
...
+	defer f.logger.Infof("Fetching %s - Finished", assetType)
+
+	gcpAssets, err := f.provider.ListAllAssetTypesByName(ctx, []string{assetType})
+	if err != nil {
+		f.logger.Errorf("Could not fetch %s: %v", assetType, err)
+		return
+	}
+
+	for _, item := range gcpAssets {
+		assetChan <- inventory.NewAssetEvent(
```
orouz, line 79: item can be one of many asset types, and depending on which type that is we'll want to use different conditional enrichers (like network, machine, instance, etc)

i plan to do this in a future PR

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -0,0 +1,100 @@
...
+					Id:   item.CloudAccount.AccountId,
+					Name: item.CloudAccount.AccountName,
+				},
+				Organization: inventory.AssetCloudOrganization{
+					Id:   item.CloudAccount.OrganisationId,
+					Name: item.CloudAccount.OrganizationName,
+				},
+				Service: &inventory.AssetCloudService{
+					Name: assetType,
+				},
```
orouz, line 96: using the asset type correctly indicates which service was used, but
1 - it's repetitive, we're already sending that string
2 - format is not human friendly 

a better solution would be using [`Resource.DiscoveryName`](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/Asset#Resource:~:text=as%20Cloud%20Bigtable.-,discoveryName,-string)

but the docs say sometimes this field will be missing. so i want to test this some more and change it in a future PR.

internal/inventory/gcpfetcher/gcpfetchers.go
```diff
@@ -0,0 +1,31 @@
...
+import (
+	"github.com/elastic/elastic-agent-libs/logp"
+
+	"github.com/elastic/cloudbeat/internal/inventory"
+	gcpinventory "github.com/elastic/cloudbeat/internal/resources/providers/gcplib/inventory"
+)
+
+func New(logger *logp.Logger, provider gcpinventory.ServiceAPI) []inventory.AssetFetcher {
+	return []inventory.AssetFetcher{
+		newAssetsInventoryFetcher(logger, provider),
```
orouz, line 29: naming is a bit repetitive here. suggestions are welcome.

internal/inventory/gcpfetcher/fetcher_assets.go
kubasobon: Good idea. The only limitation that comes to mind: is the rate-limit back-off safe to use in goroutines? If we hit API rate limits, would it work consistently for code in parallel?

internal/inventory/gcpfetcher/fetcher_assets.go
kubasobon: Makes sense. This is what we do for most resources for now.

kubasobon (approved):
🚀 Good job

Issue #2217, opened 2024-05-20
Cloudbeat panics on configuration changes


orouz:
**Describe the bug**

cloudbeat panics after getting a new config (for example, updated integration version). it ends up restarting, but it shouldn't  panic

**To Reproduce**
1. Install integration and deploy an agent
2. Send reconfiguration by doing one of the following: 
    - edit integration values other than name, if possible 
    - update integration version
    - change agent log level
3. monitor agent logs and look for: 
     1. configuration change
     2. fatal panic error

**Expected behavior**

cloudbeat handles reconfiguration gracefully and restarts without panic

**Screenshots**

![Screenshot 2024-05-20 at 9 09 12](https://github.com/elastic/cloudbeat/assets/20814186/e8def343-dd63-4caf-a6cb-7ee7e791c5cf)

**Additional context**

- [long running environment logs](https://live-environment-do-not-delete-stack.kb.us-west2.gcp.elastic-cloud.com:9243/app/observability-logs-explorer/?pageState=(breakdownField:log.level,columns:!((fallbackFields:!(host.name,service.name),smartField:resource,type:smart-field,width:320),(fallbackFields:!(message),smartField:content,type:smart-field)),controls:(data_stream.namespace:(mode:include,selection:(selectedOptions:!(),type:options))),datasetSelection:(selection:(dataset:(name:%27logs-elastic_agent.cloudbeat-*%27,title:cloudbeat_logs),name:elastic_agent,title:%27Elastic%20Agent%27,version:%271.18.0%27),selectionType:single),filters:!(),query:(language:kuery,query:%27%27),refreshInterval:(pause:!t,value:60000),rowHeight:0,rowsPerPage:500,time:(from:%272024-04-18T13:00:00.000Z%27,to:%272024-04-18T14:23:06.694Z%27),v:1))

PR #2461, opened 2024-08-28
Add missing client to posture flavor

romulets:
https://github.com/elastic/cloudbeat/blob/main/internal/flavors/posture.go#L111 is panic once a config update is received.

Tested by adding the following snippet to simulate a real update:

`internal/launcher/launcher.go:134`
```go

	go func() {
		l.log.Info("Starting force reload")
		time.Sleep(2 * time.Minute)
		l.log.Info("Force reloading")
		l.reloader.(*Listener).Reload([]*reload.ConfigWithMeta{{Config: config.NewConfig()}})
	}()

```

By debugging I noticed the client was `nil`, and once I assigned it, posture would stop as expected (and config be updated).

### Related
- Closes https://github.com/elastic/cloudbeat/issues/2217

### DIFF
--- a/internal/flavors/posture.go
+++ b/internal/flavors/posture.go
@@ -86,6 +86,7 @@ func newPostureFromCfg(b *beat.Beat, cfg *config.Config) (*posture, error) {
 			publisher: publisher,
 			config:    cfg,
 			log:       log,
+			client:    client,
 		},
 		benchmark: bench,
 	}, nil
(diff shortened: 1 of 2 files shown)


orouz (approved):
nice! 

there is no way to enforce this in Go on the type level, right? as in - not being able to initialize `flavorBase` without providing `client`

PR #2480, opened 2024-09-08
Add cloudbeat.spec to DRA

orouz:
### Summary of your changes

it's better to maintain this file in our repo than on the elastic-agent repo
[came up](https://github.com/elastic/elastic-agent/pull/5422#discussion_r1747713828) during an update to add `cloudbeat/asset_inventory_gcp` input, which is also included in this PR

### DIFF
--- a/cloudbeat.spec.yml
+++ b/cloudbeat.spec.yml
@@ -1,0 +1,95 @@
+version: 2
+inputs:
+  - name: cloudbeat
+    description: "Cloudbeat"
+    platforms: &platforms
+      - linux/amd64
+      - linux/arm64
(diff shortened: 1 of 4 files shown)


kubasobon:
@orouz With last push 5 months ago, is this still pertinent, or can we close?

orouz:
@kubasobon it is, we should do this, it's just not really urgent. i'll try to get this done once i'm done i'm done with other tasks.

kubasobon:
Ack, just asking to see which of the dangling Cloudbeat PRs are abandoned. Thank you for the info!

PR #2491, opened 2024-09-10
Refactor CDR and CIS workflows

gurevichdmitry:
### Summary of your changes

This PR introduces the capability to install the CDR infrastructure with or without the CIS infrastructure. This enhancement allows for more flexibility in setting up the infrastructure based on the specific needs of the deployment.

Key Changes:
- Added support for installing CDR independently or in combination with CIS.
- Adjusted configuration options to ensure seamless integration between CDR and CIS when both are enabled.

### Screenshot/Data

Screenshot 2024-11-14 at 11 11 07

### Checklist

### Successful Workflow Test Runs

- [CDR integrations run](https://github.com/elastic/cloudbeat/actions/runs/11764809464/job/32770382171)
- CIS integration runs
    - [9.0.0-SNAPSHOT](https://github.com/elastic/cloudbeat/actions/runs/11817881084/job/32924227480)
    - [8.17.0-SNAPSHOT](https://github.com/elastic/cloudbeat/actions/runs/11836544457/job/32981653113)
    - [8.16.0 Release](https://github.com/elastic/cloudbeat/actions/runs/11835531711/job/32978501723)
- [CIS + CDR integrations run](https://github.com/elastic/cloudbeat/actions/runs/11812719117/job/32908387500)
- [Destroy environment run](https://github.com/elastic/cloudbeat/actions/runs/11817304613/job/32922431074)


gurevichdmitry:
Workflow runs:
- [CDR successful execution](https://github.com/elastic/cloudbeat/actions/runs/10814665950)
- [CIS successful execution](https://github.com/elastic/cloudbeat/actions/runs/10815093811)
- [CDR and CIS successful execution](https://github.com/elastic/cloudbeat/actions/runs/10813052965)
- [Destroy environment execution](https://github.com/elastic/cloudbeat/actions/runs/10828714185)

deploy/test-environments/manage_infrastructure.sh
orouz, line 1: personally i'd prefer to just call the relevant scripts directly (terraform or `*.sh`), there aren't that many usages anyway. 

wrapping it adds complexity in both new code and usage

.github/workflows/cdr-infra.yml
```diff
@@ -20,13 +20,35 @@ on:
         description: "The version of the ELK stack: For BC use version without hash 8.x.y, for SNAPSHOT use 8.x.y-SNAPSHOT"
         default: "8.16.0"
         type: string
+      cis-infra:
+        required: false
+        description: "Deploy the CIS infrastructure"
+        type: boolean
+        default: false
```
orouz, line 27: why do we need CIS infra inside the CDR workflow? aren't they supposed to be separated?

.github/workflows/test-environment.yml
```diff
@@ -98,11 +98,11 @@ on:
...
         required: false
-      cdr-infra:
-        description: "Flag to indicate that the CDR infrastructure is being created"
-        type: boolean
+      infra-type:
+        description: "Type of infrastructure to create"
+        type: string
         required: false
-        default: false
+        default: "cis"
```
orouz, line 105: maybe a boolean for CDR and CIS is better? otherwise the string needs to be known to whoever runs this workflow

.github/workflows/test-environment.yml
orouz, line 1: consider splitting this workflow to multiple actions, like `deploy-elk`, `deploy-cdr`, `deploy-cis` and using whichever is required based on inputs. 

i think doing so will make the workflow more approachable for future updates as it is getting a bit complex

deploy/test-environments/set_cloud_env_params.sh
orouz, line 1: i think if each deployment was under its own action, each of them could address its requirements (like outputs) in a more straightforward way, without requiring a wrapper script

.github/workflows/cdr-infra.yml
gurevichdmitry: After refactoring the Create Environment workflow, we can provide a choice, such as a combo box with options to select CIS, CDR, or All. From the CDR perspective, we only need to specify whether we want to install CIS as well. In this case, I believe a checkbox would be preferable.

.github/workflows/test-environment.yml
gurevichdmitry: This will be resolved by refactoring the test-environment workflow in [this task](https://github.com/elastic/security-team/issues/10815).

.github/workflows/test-environment.yml
gurevichdmitry: This will be resolved by refactoring the test-environment workflow in [this task](https://github.com/elastic/security-team/issues/10815). Regarding wrapping scripts around the Terraform code, I'm not sure I agree with that approach, as it would centralize all logic related to deployment. However, we can reconsider this when the new actions for CIS and CDR are created.

PR #2493, opened 2024-09-10
[Cloud Security] [Agentless] Add the creation of an agentless agent via Agentless API in the GHA Create Environment (new)

seanrathier:
### Summary of your changes
This change refactors the CreateEnvironment workflow to create an `agentless-agent` using the Agentless-API.

This should only be merged when production Serverless projects in the test region have the Agentless-API configuration in Kibana.

 We enable this for ESS by simply setting the [TEST_AGENTLESS](https://github.com/elastic/cloudbeat/pull/2493/files#diff-e92acdebda21024508b591a7dae2cd202f1e96f7741c94ca643a88fbf9905666L138) to true.
 
#### Testing
 To test this run the [CreateEnvironment workflow](https://github.com/elastic/cloudbeat/actions/workflows/test-environment.yml) with the following parameters
 
- Workflow branch: 10463-create-agentless-agent
- Stack version: 8.16.0-SNAPSHOT
- Select `Run UI sanity tests after provision` checkbox
- Kibana UI test branch: 10269-gha-sanity-tests-for-agentless-ess-deployments

(First time doing Python,  be gentle 😄 )

### Related Issues
- https://github.com/elastic/security-team/issues/10269
- closes https://github.com/elastic/security-team/issues/10463

### Checklist
    - not sure if this is needed for workflow changes


tests/integrations_setup/install_agentless_integrations.py
```diff
@@ -84,23 +83,34 @@ def generate_gcp_integration_data():
...
+            cfg=cnfg.elk_config,
+            agent_input=AGENTLESS_INPUT,
+            package_input=integration_data,
             stream_name="cloud_security_posture.findings",
         )
-        package_policy["force"] = True
 
-        logger.info(f"Creating {package_policy}")
+        logger.info("Create agentless-agent policy")
+        agent_policy_id = create_agent_policy(cfg=cnfg.elk_config, json_policy=agent_data)
```
gurevichdmitry, line 106: Since creating an agentless-agent policy is a new feature, we need to handle both cases: when an agentless policy already exists and when it needs to be installed (as is the case for you now). 
There are a couple of options for this:
- We could rely on an environment variable to determine if the agentless-agent policy should be installed.
- Alternatively, we could retrieve all installed agent policies from the Fleet and check if any are already agentless.

seanrathier: I believe I understand what you are saying, if I am not then please correct me.

By the time this PR gets merged, Serverless production we should not be using the older preconfigured Agentless policy solution so I don't think we need check for either solution.  

When ESS production has the new Agentless API configuration we should only need to set `TEST_AGENTLESS` to true

.github/workflows/test-environment.yml
```diff
@@ -135,15 +135,15 @@ jobs:
       TF_VAR_ess_region: ${{ inputs.ess-region }}
       DEPLOYMENT_NAME: ${{ inputs.deployment_name }}
       TF_VAR_serverless_mode: ${{ inputs.serverless_mode }}
-      TEST_AGENTLESS: ${{ inputs.serverless_mode }}
```
moukoublen, line 138: I could be missing something, but this seems to be committed by mistake. We should keep it conditional based on the workflow input value.

seanrathier: We want to deploy agentless for both ESS and Serverless now. I left the environment variable there in case we disable it a briefly (bugs,  changes, etc).

moukoublen (approved):

PR #2495, opened 2024-09-10
Make isort compatible with black

kubasobon:
### Summary of your changes

I have experienced issues when running `black` (which we also use as a pre-commit hook) with `isort`. They sometimes compete with each other, making a commit pass hooks locally, but fail lint action in the Github Actions. Luckily, [isort provides a special profile](https://github.com/PyCQA/isort/issues/1518) we can use to fix that issue.

### Related Issues

Upstream advice on running `isort` and `black` together: https://github.com/PyCQA/isort/issues/1518

### DIFF
--- a/deploy/azure/ARM-for-organization-account.dev.json
+++ b/deploy/azure/ARM-for-organization-account.dev.json
@@ -27,7 +27,7 @@
             "metadata": {
                 "description": "The version of elastic-agent to install"
             },
-            "defaultValue": "8.16.0"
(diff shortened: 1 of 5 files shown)


orestisfl (approved):
Weird that it doesn't pick it up from https://github.com/elastic/cloudbeat/blob/d69411c4cf12a7612894084306742cb074c6224f/pyproject.toml#L4-L5 but why not :shrug:

PR #2528, opened 2024-09-15
Fix GCP asset inventory CI

orouz:
### Summary of your changes

the failing test case seems to be a matter of timing. 

checking for data after 20 seconds - [fails](https://github.com/elastic/cloudbeat/actions/runs/10870390768/job/30162939379?pr=2528#step:4:438)
checking for data after 40 seconds - [succeeds](https://github.com/elastic/cloudbeat/actions/runs/10870322998/job/30162805992)

not sure why this started happening consistently just now, could be [gcp updates](https://github.com/elastic/cloudbeat/pull/2486/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R92) but i'm not sure i see something definitive in the [changelog](https://github.com/googleapis/google-cloud-go/releases?q=auth&expanded=true)

instead of just bumping the sleep period, i've added a new action - `wait-for-es-data` which is essentially the same thing but checks the index every 1s and if we have data it goes on to the next step (run the tests).  - [successful run](https://github.com/elastic/cloudbeat/actions/runs/10878341142/job/30181124690?pr=2528)

### DIFF
--- a/.github/actions/gcp-asset-inventory-ci/action.yml
+++ b/.github/actions/gcp-asset-inventory-ci/action.yml
@@ -35,9 +35,10 @@ runs:
       run: |
         ./cloudbeat -c deploy/asset-inventory/cloudbeat-gcp-asset-inventory.yml -d '*' &
 
-    - name: Wait for cloudbeat to send some events
-      shell: bash
-      run: sleep 20
+    - name: Wait for data
+      uses: ./.github/actions/wait-for-es-data
+      with:
+        index: .ds-logs-cloud_asset_inventory.asset_inventory-*
 
     - name: Check for assets
       working-directory: ./tests
--- a/.github/actions/wait-for-es-data/action.yml
+++ b/.github/actions/wait-for-es-data/action.yml
@@ -1,0 +1,35 @@
+name: "Wait for ES Data"
+description: "Waits for data on ES indices"
+inputs:
+  index:
+    description: "Which ES index to look at"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Wait for data
+      env:
+        ES_HOST: http://localhost:9200
+      shell: bash
+      run: |
+        URL="${ES_HOST}/${{ inputs.index }}/_search?size=0"
+        HEADER="Content-Type: application/json"
+        MAX_RETRIES=100
+
+        echo "Checking index: ${{ inputs.index }}"
+
+        for ((i = 1; i <= MAX_RETRIES; i++)); do
+            result=$(curl -s -X GET "$URL" -H "$HEADER" | jq '.hits.total.value')
+
+            if [ "$result" -gt 0 ]; then
+                echo "Data found: $result documents in index ${{ inputs.index }}"
+                exit 0
+            fi
+
+            echo "No data yet. Retrying in 1s... ($i/$MAX_RETRIES)"
+            sleep 1
+        done
+
+        echo "No data found after $MAX_RETRIES retries."
+        exit 1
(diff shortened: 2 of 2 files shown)


.github/actions/gcp-asset-inventory-ci/action.yml
```diff
@@ -35,9 +35,10 @@ runs:
       run: |
         ./cloudbeat -c deploy/asset-inventory/cloudbeat-gcp-asset-inventory.yml -d '*' &
 
-    - name: Wait for cloudbeat to send some events
-      shell: bash
-      run: sleep 20
+    - name: Wait for data
+      uses: ./.github/actions/wait-for-es-data
+      with:
+        index: .ds-logs-cloud_asset_inventory.asset_inventory-*
```
orouz, line 41: Screenshot 2024-09-15 at 14 36 15


in this case it took 26~ seconds

.github/actions/wait-for-es-data/action.yml
```diff
@@ -0,0 +1,35 @@
...
+  index:
+    description: "Which ES index to look at"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Wait for data
+      env:
+        ES_HOST: http://localhost:9200
```
gurevichdmitry, line 13: It may be more robust to set es_host as an input variable with a default value of http://localhost:9200, and then use the following configuration:
```
env:
  ES_HOST: ${{ inputs.es_host }}
```

gurevichdmitry, line 13: For the input index, it is better to set it as an environment variable and then use it in the inline script.
```
env:
  INDEX: ${{ inputs.index }}
```
...

```
URL="${ES_HOST}/${INDEX}/_search?size=0"
```
...
```
echo "Checking index: ${INDEX}"
```

.github/actions/wait-for-es-data/action.yml
```diff
@@ -0,0 +1,35 @@
...
+            result=$(curl -s -X GET "$URL" -H "$HEADER" | jq '.hits.total.value')
+
+            if [ "$result" -gt 0 ]; then
+                echo "Data found: $result documents in index ${{ inputs.index }}"
+                exit 0
+            fi
+
+            echo "No data yet. Retrying in 1s... ($i/$MAX_RETRIES)"
+            sleep 1
+        done
```
gurevichdmitry, line 32: The waiting loop is currently based on a hardcoded maximum retries count of 100. Since this action is intended for reuse, consider introducing a max_retries input parameter with a default value of 100, along with an explanation of how the retries work, as the timeout may vary depending on the connection to different providers.

Alternatively, you could rewrite the loop to use a time-based approach, specifying the total wait time in seconds rather than relying on the maximum retries.

.github/actions/wait-for-es-data/action.yml
orouz: i don't see how this is better, it's just longer. Could you explain how making an input variable also an environment variable adds value?

.github/actions/wait-for-es-data/action.yml
orouz: added `retries` input. because each retry is 1s, this is effectively time based (100 retries is roughly 100s, plus the time it takes to execute the curl command). 

note that the whole point of using this action is to avoid tweaking sleep period time for different consumers, so they can all just "wait for data". some may take 20 seconds, some 24, 26, 30 or 40. if it takes more than 100 seconds, something is probably not working.

oren-zohar (approved):

Issue #2502, opened 2024-09-12
Agentless service account deployment lacks organization permissions


orouz:
**Describe the bug**

in agentless, deploying a service account for CSPM GCP `organization-account` ends up not fetching resources due to lack of permissions. 

**To Reproduce**

1. deploy agentless in qa env
2. deploy CSPM GCP, pick agentless mode
3. open cloudshell, paste command to deploy the service account
4. copy the service account JSON and paste in integration installation
5. go to agent logs, see Permission Denied errors

**Expected behavior**

- no permission denied errors

**Additional context**

- fix was verified by manually granting the service account email access to the `"roles/cloudasset.viewer", "roles/browser"` roles on the organization level

---

- https://github.com/elastic/cloudbeat/pull/2118
- https://github.com/elastic/cloudbeat/issues/1781

PR #2529, opened 2024-09-15
Fix GCP deployment manager formatting

orouz:
### Summary of your changes

there were 2 issues preventing GCP CSPM on agentless from successfully deploying a service account for organizations. the first was a slightly off copy-paste command in [kibana](https://github.com/elastic/kibana/pull/192959) which ended up not passing `ORG_ID` to the deployment script, so the deployment assumed it's for a project. the second issue, after providing `ORG_ID`, the deployment script complained about it being a number and not a string (as per its schema). this is what this PR fixes.

after fixing both of these, i've deployed GCP CSPM on agentless and got findings for an organization account:

### Screenshot/Data
Screenshot 2024-09-15 at 17 59 46

### Related Issues
 - fixes https://github.com/elastic/cloudbeat/issues/2502

### DIFF
--- a/deploy/deployment-manager/deploy_service_account.sh
+++ b/deploy/deployment-manager/deploy_service_account.sh
@@ -35,7 +35,7 @@ fi
 
 result="$(gcloud deployment-manager deployments create --automatic-rollback-on-error "${DEPLOYMENT_NAME}" --project "${PROJECT_NAME}" \
     --template service_account.py \
-    --properties scope:"${SCOPE}",parentId:"${PARENT_ID}",serviceAccountName:"${SERVICE_ACCOUNT_NAME}")"
+    --properties "scope:'${SCOPE}',parentId:'${PARENT_ID}',serviceAccountName:'${SERVICE_ACCOUNT_NAME}'")"
 
 key="$(echo "$result" | awk '/serviceAccountKey/{getline; print}' | awk '{print $2}')"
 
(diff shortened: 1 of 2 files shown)


deploy/deployment-manager/deploy_service_account.sh
```diff
@@ -35,7 +35,7 @@ fi
 
 result="$(gcloud deployment-manager deployments create --automatic-rollback-on-error "${DEPLOYMENT_NAME}" --project "${PROJECT_NAME}" \
     --template service_account.py \
-    --properties scope:"${SCOPE}",parentId:"${PARENT_ID}",serviceAccountName:"${SERVICE_ACCOUNT_NAME}")"
+    --properties "scope:'${SCOPE}',parentId:'${PARENT_ID}',serviceAccountName:'${SERVICE_ACCOUNT_NAME}'")"
```
orouz, line 38: this format is required to ensure the values passed are treated as strings

see https://cloud.google.com/sdk/gcloud/reference/deployment-manager/deployments/create

deploy/deployment-manager/service_account.py
```diff
@@ -14,7 +14,7 @@ def generate_config(context):
         "name": sa_name,
         "type": "iam.v1.serviceAccount",
         "properties": {
-            "accountId": sa_name,
+            "accountId": f"{sa_name}-id",
```
orouz, line 17: i'm not entirely sure why this is needed in addition to `dependsOn`, but i ran into some flakiness when deploying the service account, this seems to fix it.

orouz:
moved to https://github.com/elastic/cloudbeat/pull/2530

PR #2530, opened 2024-09-15
Fix GCP deployment manager formatting

orouz:
### Summary of your changes

there were 2 issues preventing GCP CSPM on agentless from successfully deploying a service account for organizations. the first was a slightly off copy-paste command in [kibana](https://github.com/elastic/kibana/pull/192959) which ended up not passing `ORG_ID` to the deployment script, so the deployment assumed it's for a project. the second issue, after providing `ORG_ID`, the deployment script complained about it being a number and not a string (as per its schema). this is what this PR fixes.

after fixing both of these, i've deployed GCP CSPM on agentless and got findings for an organization account:

### Screenshot/Data
Screenshot 2024-09-15 at 17 59 46

### Related Issues
 - fixes https://github.com/elastic/cloudbeat/issues/2502

### DIFF
--- a/.github/workflows/test-gcp-dm.yml
+++ b/.github/workflows/test-gcp-dm.yml
@@ -16,6 +16,10 @@ on:
       - "deploy/deployment-manager/service_account.py"
       - "deploy/deployment-manager/service_account.py.schema"
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 env:
   TEST_ENVS_DIR: deploy/test-environments
   INTEGRATIONS_SETUP_DIR: tests/integrations_setup
--- a/deploy/deployment-manager/deploy_service_account.sh
+++ b/deploy/deployment-manager/deploy_service_account.sh
@@ -35,7 +35,7 @@ fi
 
 result="$(gcloud deployment-manager deployments create --automatic-rollback-on-error "${DEPLOYMENT_NAME}" --project "${PROJECT_NAME}" \
     --template service_account.py \
-    --properties scope:"${SCOPE}",parentId:"${PARENT_ID}",serviceAccountName:"${SERVICE_ACCOUNT_NAME}")"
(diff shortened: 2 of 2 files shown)


deploy/deployment-manager/deploy_service_account.sh
```diff
@@ -35,7 +35,7 @@ fi
 
 result="$(gcloud deployment-manager deployments create --automatic-rollback-on-error "${DEPLOYMENT_NAME}" --project "${PROJECT_NAME}" \
     --template service_account.py \
-    --properties scope:"${SCOPE}",parentId:"${PARENT_ID}",serviceAccountName:"${SERVICE_ACCOUNT_NAME}")"
+    --properties "scope:'${SCOPE}',parentId:'${PARENT_ID}',serviceAccountName:'${SERVICE_ACCOUNT_NAME}'")"
```
orouz, line 38: this format is required to ensure the values passed are treated as strings

see https://cloud.google.com/sdk/gcloud/reference/deployment-manager/deployments/create

in this specific case, ORG_ID is actually a number, but we need it as a string

.github/workflows/test-gcp-dm.yml
```diff
@@ -26,7 +26,23 @@ env:
...
+      STACK_VERSION: ${{ steps.stack_version.outputs.STACK_VERSION }}
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+
+      - name: Set stack version
+        id: stack_version
+        run: |
+          version=$(grep defaultBeatVersion version/version.go | cut -f2 -d "\"")
+          echo "STACK_VERSION=$version-SNAPSHOT" >> $GITHUB_OUTPUT
```
orouz, line 42: `9.0.0` builds don't work yet for ESS so we need to wait before merging this

orouz:
for some reason there is a bit of [flakiness](https://github.com/elastic/cloudbeat/actions/runs/11093739352/job/30820588148?pr=2530#step:11:177) in service account deployments which i can't exactly pin

oren-zohar (approved):

Issue #2546, opened 2024-09-20
Create Environment failing for serverless


opauloh:
**Describe the bug**

As described in this slack [thread](https://elastic.slack.com/archives/C03E5KGNWT1/p1726827523674779) @maxcold faced an issue when creating a serverless project with our GitHub action.

### Screenshots

![Image](https://github.com/user-attachments/assets/175eb6d3-0102-45a1-9407-9ecd2d81ecc8)

### Logs

```
Run ./install_agent_az_cli.sh
{
  "id": "/subscriptions/ef111ee2-6c89-4b09-92c6-5c2321f888df/resourceGroups/paulo-svl",
  "location": "eastus",
  "managedBy": null,
  "name": "paulo-svl",
  "properties": {
    "provisioningState": "Succeeded"
  },
  "tags": null,
  "type": "Microsoft.Resources/resourceGroups"
}
[...]

PR #2547, opened 2024-09-20
Fix Create Environment agent version for Serverless

opauloh:
### Summary of your changes

This fixes https://github.com/elastic/cloudbeat/issues/2546, by using the version inputted by the user in the Create Environment parameters instead of using the version from the `env.ELK_VERSION`.

While it is true that the version doesn't matter on Serverless in terms of Kibana and Elasticsearch, it does matter for deploying agents as the snapshots might be no longer available.


gurevichdmitry (approved):

opauloh:
@gurevichdmitry it seems I can't merge this PR even after it was approved, could you merge it?

PR #2576, opened 2024-10-01
[Asset Inventory][Azure] Add storage assets

kubasobon:
### Summary of your changes

Add 3 storage assets: Queues, Queue Services, and Blob Services.

### Related Issues

Closes https://github.com/elastic/security-team/issues/10162
Closes https://github.com/elastic/security-team/issues/10163

### Checklist


internal/inventory/asset.go
```diff
@@ -99,6 +101,9 @@ const (
 	SubTypeAzureSQLServer           AssetSubType = "azure-sql-server"
 	SubTypeAzureSnapshot            AssetSubType = "azure-snapshot"
 	SubTypeAzureStorageAccount      AssetSubType = "azure-storage-account"
+	SubTypeAzureStorageBlobService  AssetSubType = "azure-storage-blob-service"
+	SubTypeAzureStorageQueue        AssetSubType = "azure-storage-queue"
+	SubTypeAzureStorageQueueService AssetSubType = "azure-storage-queue-service"
```
romulets, line 106: What is the difference between azure-storage-queue and azure-storage-queue-service?

internal/inventory/azurefetcher/fetcher_storage.go
```diff
@@ -0,0 +1,121 @@
...
+	storageAccounts, err := f.provider.ListStorageAccounts(ctx, subscriptionIds)
+	if err != nil {
+		return nil, fmt.Errorf("error listing storage accounts: %v", err)
+	}
+
+	return storageAccounts, nil
+}
+
+func (f *storageFetcher) fetch(ctx context.Context, storageAccounts []azurelib.AzureAsset, resourceName string, function storageProviderFunc, classification inventory.AssetClassification, assetChan chan<- inventory.AssetEvent) {
+	f.logger.Infof("Fetching %s", resourceName)
```
romulets, line 95: Do we have this for every fetcher? I wonder if we are not over-logging

internal/resources/providers/azurelib/inventory/storage_provider.go
```diff
@@ -152,6 +180,62 @@ func (p *storageAccountProvider) ListStorageAccountBlobServices(ctx context.Cont
...
+				TenantId:       sa.TenantId,
+				Properties:     properties,
+				Extension: map[string]any{
+					ExtensionStorageAccountID:   sa.Id,
+					ExtensionStorageAccountName: sa.Name,
+				},
+			})
+		}
+		if errs != nil {
+			return nil, fmt.Errorf("error while transforming azure queue services for storage accounts %s: %w", sa.Id, errs)
```
romulets, line 232: I wonder if we should return nil here and don't store any StorageQueueService. 

The error is not a blocking one, right? We rather have the asset without properties anyway - and log the error.

Wdyt?

internal/resources/providers/azurelib/inventory/storage_provider.go
```diff
@@ -152,6 +180,62 @@ func (p *storageAccountProvider) ListStorageAccountBlobServices(ctx context.Cont
...
+func (p *storageAccountProvider) ListStorageAccountQueues(ctx context.Context, storageAccounts []AzureAsset) ([]AzureAsset, error) {
+	var assets []AzureAsset
+	for _, sa := range storageAccounts {
+		responses, err := p.client.AssetQueues(ctx, sa.SubscriptionId, nil, sa.ResourceGroup, sa.Name, nil)
+		if err != nil {
+			return nil, fmt.Errorf("error while fetching azure queues for storage accounts %s: %w", sa.Id, err)
+		}
+
+		queues, err := transformQueues(responses, sa)
+		if err != nil {
```
romulets, line 192: Same comment as StorageQueueService, should we stop the execution if the transform queue can't extract properties?

romulets (changes requested):

internal/inventory/azurefetcher/fetcher_storage.go
kubasobon: Yes we do. Assuming we'll have ~200 assets, even if we're scanning every minute, that's 400 lines/minute. I'd say that is a very tame amount of logs. If we want to change the approach, then let's do it in a separate PR for all of the fetchers.

internal/resources/providers/azurelib/inventory/storage_provider.go
kubasobon: Makes sense 👍

internal/resources/providers/azurelib/inventory/storage_provider.go
kubasobon: Makes sense 👍

romulets (approved):

PR #2625, opened 2024-10-18
[Asset Inventory][Azure] Add Acitive Directory fetcher to fetch Service Principals

kubasobon:
### Summary of your changes

This PR adds an Active Directory fetcher and fetchers Azure Service Principals. It utilizes [microsoftgraph/msgraph-sdk-go](https://github.com/microsoftgraph/msgraph-sdk-go), since this Active Directory API is not exposed in the SDK we've been using so far. This is the recommended approach.

### Screenshot/Data

Screenshot 2024-10-18 at 15 24 41

### Related Issues

Closes https://github.com/elastic/security-team/issues/10161 

### Checklist


internal/inventory/azurefetcher/fetcher_activedirectory.go
```diff
@@ -0,0 +1,84 @@
...
+		var tenantId string
+		if uuid := item.GetAppOwnerOrganizationId(); uuid != nil {
+			tenantId = uuid.String()
+		}
+		assetChan <- inventory.NewAssetEvent(
+			inventory.AssetClassificationAzureServicePrincipal,
+			[]string{pointers.Deref(item.GetId())},
+			pointers.Deref(item.GetDisplayName()),
+			inventory.WithRawAsset(
+				item.GetBackingStore().Enumerate(),
```
romulets, line 71: Why is the raw set the enumarated back store? I actually don't understand what we are getting here

internal/resources/providers/msgraph/provider.go
```diff
@@ -0,0 +1,91 @@
...
+		return nil, fmt.Errorf("error paging Azure Service Principals: %w", err)
+	}
+
+	items := []*models.ServicePrincipal{}
+	err = pageIterator.Iterate(ctx, func(pageItem *models.ServicePrincipal) bool {
+		items = append(items, pageItem)
+		return true // to continue the iteration
+	})
+	if err != nil {
+		p.log.Errorf("error iterating over Service Principals: %v", err)
```
romulets, line 88: shouldn't we return the error if principals are not returned?

internal/inventory/azurefetcher/fetcher_activedirectory.go
kubasobon: This is due to the nature of underlying GCP SDK. It uses its own wrapper for data, which it calls a `BackingStore`. `item` exposes some fields, not all, it's just a generic interface.
So to get "raw" asset in this case, we use `GetBackingStore() -> *gcpsdk.BackingStore -> Enumerate() -> map[string]any`. This gives us full resource representation in `map[string]any` format.

internal/resources/providers/msgraph/provider.go
kubasobon: This is a log, because we might still get some resources. The pagination failed, but `items` slice has been populated. We try to get the best scenario for us - getting data, even if incomplete.

kubasobon:
@romulets Please take another look

romulets (approved):

Issue #2645, opened 2024-10-28
`cloud.instance.id` and `resource.id` missing `i-` prefix for CNVM


maxcold:
**Describe the bug**
In AWS Instance ids usually start with `i-` prefix, but in the data coming from CNVM integration this prefix is stripped 

**Preconditions**
Have an environment with the data ingested from CNVM

**To Reproduce**
Write the exact actions one should perform in order to reproduce the bug.
1. Open any vulnerability finding and look for `cloud.instance.id` and `resource.id` fields, they will be missing `i-` prefix

**Expected behavior**

The instance id should not be stripped out of any symbols

![Image](https://github.com/user-attachments/assets/6d1ac7f3-9573-4240-9de8-d756d618b4e6)

animehart:
![Image](https://github.com/user-attachments/assets/df523447-e019-49e2-9a86-fe1394a14ef0)
![Image](https://github.com/user-attachments/assets/054d05a6-34e3-4e55-9bee-c4f5e249e79b)

`cloud.instance.id` and `resource.id` has i- prefix now
Verified

PR #2651, opened 2024-10-30
cnvm: include full ec2 instance id in findings fields

orouz:
### Summary of your changes

for ec2 instances, return the full id, including the `i-` prefix

### Screenshot/Data
| before 	| after 	|
|--------	|-------	|
|    Screenshot 2024-10-30 at 14 45 47    	| Screenshot 2024-10-30 at 15 35 24 |

### Related Issues
- fixes https://github.com/elastic/cloudbeat/issues/2645


romulets (approved):
Interesting, why did we ever do it? haha

PR #2654, opened 2024-10-31
[Asset Inventory][GCP] Populate `related.entity` field for resources implemented so far

kubasobon:
### Summary of your changes

Populate `related.entity` field for resources implemented so far.

### Screenshot/Data

Screenshot 2024-10-31 at 14 30 43

### Related Issues

Towards https://github.com/elastic/security-team/issues/10392

### Checklist


kubasobon:
Finished for all types of resources we have implemented so far.

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -103,3 +108,74 @@ func (f *assetsInventory) fetch(ctx context.Context, assetChan chan<- inventory.
...
+	if item.Resource != nil && item.Resource.Data != nil {
+		fields = item.GetResource().GetData().GetFields()
+	}
+
+	switch subType {
+	case inventory.SubTypeGcpInstance:
+		if v, ok := fields["networkInterfaces"]; ok {
+			for _, networkInterface := range v.GetListValue().GetValues() {
+				networkInterfaceFields := networkInterface.GetStructValue().GetFields()
+				ids = appendIfExists(ids, networkInterfaceFields, "network")
```
romulets, line 139: Why do we need to have a map with the name of the field?

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -103,3 +108,74 @@ func (f *assetsInventory) fetch(ctx context.Context, assetChan chan<- inventory.
...
+	case inventory.SubTypeGcpProject, inventory.SubTypeGcpBucket:
+		if item.IamPolicy == nil {
+			break
+		}
+		for _, binding := range item.IamPolicy.Bindings {
+			ids = append(ids, binding.Role)
+			ids = append(ids, binding.Members...)
+		}
+	default:
+		f.logger.Warnf("cannot find related asset IDs for unsupported sub-type %q", subType)
```
romulets, line 168: Do we really need a warn here? It seems normal to me that some types doesn't have related entities

romulets (approved):
You went deeper on this than I first did. It's good to know that aws and azure doesn't cover this part.

internal/inventory/gcpfetcher/fetcher_assets.go
kubasobon: I'm not entirely sure what you are asking about. Is it about "Why do we have to call `...GetFields(), which returns a map, that we have to query further?"?
If so, then the answer is Google's SDK is made that way 🤷

internal/inventory/gcpfetcher/fetcher_assets.go
kubasobon: True. I'll drop the `warn`.

PR #2691, opened 2024-11-07
Improve 8.x Dependency Updates

romulets:
### Summary of your changes

- Specific dependabot for 8.x
- Update CLI for 8.x


.github/dependabot.yml
```diff
@@ -29,7 +29,41 @@ updates:
       - "backport-v8.x"
```
oren-zohar, line 29: do we need to remove it?

romulets: Yes, we need.

oren-zohar (approved):

PR #2747, opened 2024-11-18
iam/user.go: Improve logging

orestisfl:
Avoid log & return (leads to double log), add more error information on `err` returns.

### DIFF
--- a/internal/resources/providers/awslib/iam/user.go
+++ b/internal/resources/providers/awslib/iam/user.go
@@ -46,7 +46,6 @@ const (
 func (p Provider) GetUsers(ctx context.Context) ([]awslib.AwsResource, error) {
 	apiUsers, err := p.listUsers(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/resources/providers/awslib/iam/user.go
```diff
@@ -46,7 +46,6 @@ const (
 func (p Provider) GetUsers(ctx context.Context) ([]awslib.AwsResource, error) {
 	apiUsers, err := p.listUsers(ctx)
 	if err != nil {
-		p.log.Errorf("fail to list users, error: %v", err)
 		return nil, err
```
romulets, line 49: Why not wrap this one too?

romulets (approved):

internal/resources/providers/awslib/iam/user.go
orestisfl: It's wrapped by `listUsers()`, too many wraps 🌯🫃

PR #2748, opened 2024-11-19
Add slack notifications to BK pipeline

gurevichdmitry:
### Summary of your changes

This PR introduces Slack notifications for the Buildkite pipeline with the following logic:
- If the pipeline fails, the appropriate user will be notified in Slack.
- User identification is based on the author's email.
   - If the email field is empty or the user is not found in the defined list, the search will proceed to the comment message for an email address.
   - If no user can be identified through either method, a default team group will be notified


kubasobon (approved):

.buildkite/scripts/notify.sh
```diff
@@ -0,0 +1,45 @@
...
+user_not_found="user_not_found"
+
+# Check if BUILDKITE_BUILD_CREATOR_EMAIL is set, defaulting to user not found
+build_creator="${BUILDKITE_BUILD_CREATOR_EMAIL:-$user_not_found}"
+build_message="${BUILDKITE_MESSAGE:-}"
+slack_channel="${SLACK_CHANNEL:-"#cloud-sec-ci"}"
+
+default_group_id=$(vault kv get -field="cloudbeat-eng-team" secret/ci/elastic-cloudbeat/slack-users)
+
+slack_user_id=$(vault kv get -field="$build_creator" secret/ci/elastic-cloudbeat/slack-users 2>/dev/null || echo "$user_not_found")
```
oren-zohar, line 20: nit: using "user_not_found" and then replacing it with `cloudbeat-eng-team` later is redundant, you can set `cloudbeat-eng-team` initially

gurevichdmitry: The script has been refactored

PR #2751, opened 2024-11-20
[8.17 freeze] bump cloudbeat versions

kubasobon:
Bump `cloudbeat` versions due to v8.17 feature freeze

Related issue: https://github.com/elastic/dev/issues/2878


bin/hermit.hcl
```diff
@@ -1,4 +1,4 @@
 env = {
-  "CLOUDBEAT_VERSION": "8.15.0",
+  "CLOUDBEAT_VERSION": "8.18.0",
```
moukoublen, line 2: I might be mistaken here but I think the docker images of 8.18.0 might not be ready yet and since we use that in ci it might brake. If I remember correctly.

Issue #2563, opened 2024-09-26
Bump `hermit.hcl` version whenever `version.go` is bumped


orouz:
**Motivation**

`hermit.hcl` includes env variables used in CI which determine which artifacts are used for testing deployments
currently minor releases are created using the bump-version workflow, which bumps both `version.go` and `hermit.hcl` in separate PRs, so we can merge the hermit PR once we know the artifacts CI will now require are available 

but whenever a patch release is made (as in - `8.15.0` -> `8.15.1`) the versions in `version.go` and `hermit.hcl` goes out of sync, which makes the CI workflows run with an outdated stack

what we could do is make a workflow that bumps `hermit.hcl` whenever the version in `version.go` is bumped

**Definition of done**
- [ ] versions in `hermit.hcl` and `version.go` are guaranteed to be eventually consistent

Issue #2492, opened 2024-09-10
Workflows updates after `8.x` branch-out


moukoublen:
**Motivation**
We branched out of the [8.x](https://github.com/elastic/cloudbeat/tree/8.x) branch, and both [8.x](https://github.com/elastic/cloudbeat/tree/8.x) and the main (9.x) branches will be long-lived branches from which we will branch out of the `8.x.y` and `9.x.y` releases.

Because of that we need to update our [bump-version.yml](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/bump-version.yml) and [updatecli.yml](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/updatecli.yml) workflows to follow that logic. 

### [A] `bump-version.yml`
The [bump-version.yml](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/bump-version.yml) should either be able to handle version bump for both `9.x.y` and `8.x.y` releases, or keep working for the latest version `9.x.y` and we will handle `8.x.y` bumps manually. 

### [B] `.github/workflows/publish-cloudformation.yml`
We should adjust the cloudformation deployment for the `8.x` branch (or ignore it). 

### [C] Dependencies and tools (updatecli.yml && dependabot)
The flow stays the same for the `main` branch as it currently is. 

For the `8.x` branch, we need to keep backporting everything (expect from beats) from the `main` up to the point that the two major releases will have different features planned (features/fixes in the `9.0.0` release that are not scheduled for the `8.x.y` release). 

But regarding dependencies updates, we should decide on one of the following: 
  1. Backport each dependency update PR and solve the conflict on each PR (if occurs).
  2. Establish a process in which we merge the dependencies updates PR to take place in the same order (when we merge the bump A for main next thing is to merge bump A for 8.x)
  3. Do not backport bumps and have a standalone updatecli +dependabot over the 8.x

More specifically the `beats` updates perhaps needs to follow the beats `8.x` branch for `cloudbeat` `8.x` branch and `beats` `main` branch for `cloudbeat` `main` branch.

Based on that we need to update [updatecli.yml](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/updatecli.yml) and dependabot accordingly.

romulets:
Reopening because we need to think about the bump flow

kubasobon:
### Update
ad. A: I opened a PR to fix the issue https://github.com/elastic/cloudbeat/pull/2757
[...]

PR #2757, opened 2024-11-21
[CI] Update bump version workflow to work for x.x branches

kubasobon:
### Summary of your changes

This PR aims to make "Bump Workflow" action usable for branches like `8.x`.
It also introduces a "Synchronize Cloudbeat Versions on release" workflow, which solves version drift in `bin/hermit.hcl` file.

**Work in progress. Testing pending**

### Related Issues

Closes https://github.com/elastic/cloudbeat/issues/2492
Closes https://github.com/elastic/cloudbeat/issues/2563


scripts/sync_internal_cloudbeat_version.sh
```diff
@@ -0,0 +1,30 @@
...
+        return
+    fi
+    echo "Versions changed, commiting changes"
+    git add $HERMIT_FILE
+    git commit -m "bump CLOUDBEAT_VERSION in $HERMIT_FILE to $CLOUDBEAT_VERSION"
+}
+
+find_current_cloudbeat_version
+set_hermit_cloudbeat_version
+create_pr_if_different
```
gurevichdmitry, line 30: @kubasobon Did you mean `commit_if_different` instead of `create_pr_if_different`?

.github/workflows/bump-version.yml
```diff
@@ -19,6 +19,7 @@ jobs:
       - name: Checkout Cloudbeat Repo
         uses: actions/checkout@v4
         with:
+          ref: ${{ github.base_ref || github.ref_name }}
```
gurevichdmitry, line 22: @kubasobon I'm not entirely sure about the use of `github.base_ref` in this workflow. Since the only trigger for this workflow is `workflow_dispatch`, `github.base_ref` will not be relevant and will always be empty. We only need to use `github.ref_name` to refer to the branch being processed. As mentioned [there](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs), the `github.base_ref` property is only available when the event triggering the workflow is either `pull_request` or `pull_request_target`. It will not be populated for other events like `workflow_dispatch`.

.github/workflows/bump-version.yml
```diff
@@ -39,8 +40,11 @@ jobs:
 
       - name: Bump Cloudbeat
         # bump_cloudbeat.sh will create multiple PRs with different HEAD branches
+        env:
+          GIT_BASE_BRANCH: ${{ github.base_ref || github.ref_name }}
```
gurevichdmitry, line 44: The same as mentioned before.

.github/workflows/bump-version.yml
```diff
@@ -39,8 +40,11 @@ jobs:
...
         run: scripts/bump_cloudbeat.sh
 
       - name: Bump Cloud Security Posture Integration
         # we need to run bump_integration.sh from the main branch
+        if: github.base_ref == 'main'
```
gurevichdmitry, line 49: This condition enforces that the workflow must run on the `main` branch for the bump integration. If this is the case, we need to document this behavior and also update the documentation to describe the version bump process.

scripts/bump_cloudbeat.sh
```diff
@@ -6,6 +6,7 @@ CURRENT_MINOR_VERSION=$(echo "$CURRENT_CLOUDBEAT_VERSION" | cut -d '.' -f1,2)
 export CURRENT_MINOR_VERSION
 
 # branches
+export BASE_BRANCH="${GIT_BASE_BRANCH:-main}"
```
gurevichdmitry, line 9: Since the bump version only works with the `workflow_dispatch` trigger, you can select any branch, including unconventional ones that may not be relevant for the bump process. I think we should add some checks on the `BASE_BRANCH` and restrict the script or workflow action `Bump Cloudbeat` to run only for specific patterns, such as `main`, `8.`, and `8.x`.

gurevichdmitry (commented):
Great work! It would be helpful to provide some workflow runs to ensure that the functionality is working as expected.

scripts/bump_cloudbeat.sh
kubasobon: That's a good idea, I've added a `validate_base_branch()` function to help with that.

scripts/bump_cloudbeat.sh
```diff
@@ -253,6 +254,21 @@ bump_snyk_branch_monitoring() {
 
 }
 
-run_version_changes_for_main
+validate_base_branch() {
+    if [[ "$BASE_BRANCH" == "main" || "$BASE_BRANCH" == "8.x" || "$BASE_BRANCH" == "9.x" ]]; then
```
gurevichdmitry, line 258: In the current state of the repository, the following important branches require version bumps:
- `main`: Currently points to version 9.0.
- `8.x`: The `main` branch for all `"8"` versions.
- `8.*`: Represents all released versions, such as `8.16`, `8.15`, etc.
- `9.*`: Future branches for "9" versions.
To handle these cases, we can verify whether $BASE_BRANCH == "main", additionally, we need to implement a regular expression to match branch patterns for 8. and 9.

Maybe we could consider something like this:
`[[ "$BASE_BRANCH" == "main" || "$BASE_BRANCH" =~ ^(8\..*|9\..*)$ ]]`

kubasobon:
Test PRs created by test run:
- https://github.com/elastic/cloudbeat/pull/2776
- https://github.com/elastic/cloudbeat/pull/2777

scripts/bump_cloudbeat.sh
gurevichdmitry: Since both if statements return the same result, I think it would be better to combine them into a single clause.

kubasobon: Done

PR #2769, opened 2024-11-25
Add entities.metadata

romulets:
### Summary of your changes

As part of implementing the first iteration of the [Cloud Entities in Entity Store](https://github.com/elastic/security-team/issues/10827), we need to populate the field `entities.metadata` with the relevant information

### Screenshot/Data
image

### Related Issues
- Related https://github.com/elastic/security-team/issues/11094
- Related https://github.com/elastic/security-team/issues/10961


internal/inventory/asset.go
```diff
@@ -329,9 +329,32 @@ type AssetResourcePolicy struct {
...
 
+func (a *AssetEvent) getEntityMetadata() map[string]entityMetadata {
+	ids := a.Asset.Id
+
+	if len(ids) == 0 {
+		return nil
+	}
+
+	// Picking up only first id, we need to make a decision on if we
+	// have a "primary" id or if we duplicate data
```
kubasobon, line 349: That's super important. We also have to pay attention to always put "the" asset ID in the `ids[0]` spot.

internal/inventory/asset.go
```diff
@@ -329,9 +329,32 @@ type AssetResourcePolicy struct {
 	Condition  map[string]any `json:"condition,omitempty"`
 }
 
+// EntityMetadata maps metadata required to use Entity Store
+type entityMetadata struct {
+	Category string `json:"category"`
+	Type     string `json:"type"`
```
kubasobon, line 335: Are there any other fields planned? What will happen if we change the taxonomy we have now (Category, Sub-Category, Type, Sub-type) to a less granular one?

kubasobon (approved):
The code looks good. Asking a couple of questions for my information. We can catch-up on Zoom about this if you'd like that better.

internal/inventory/asset.go
romulets: This is a product question I want to clear it up. I believe it was a bad decision in the end to have multiple ids.

We need to have a standardized id per resource. We need to make the decision for example is aws iam user arn or username? And one for each one of those.

The initial attempt of having a "correlation best chance" approach isn't going to help us in the asset inventory, unfortunately.

internal/inventory/asset.go
romulets: There might have other fields.

Once we change the taxonomy, we should follow the taxonomy.

romulets:
We won't move forward with this strategy

PR #2877, opened 2025-01-03
golangci-lint: Add more linters

orestisfl:
Adds new 1.63 linters:
- https://github.com/alingse/nilnesserr: Provides a concise way to detect return an unrelated/nil-values error.
- https://github.com/ldez/exptostd: Detects functions from golang.org/x/exp/ that can be replaced by std functions.
- https://github.com/ldez/usetesting: Detects when some calls can be replaced by methods from the testing package.

Additionally, add more older, useful linters.


kubasobon (approved):

romulets (approved):
I don't really like a lot linters that force certain naming styles...

But that's a personal opinion. I think the rest of the team is onboard with it.

I'll approve it, but wanted to share my feeling :P

orestisfl:
> I don't really like a lot linters that force certain naming styles...
> 
> But that's a personal opinion. I think the rest of the team is onboard with it.
> 
> I'll approve it, but wanted to share my feeling :P

In my opinion, what matters most is consistency. I don't necessarily agree with the naming style but this seems to be the Go idiom and I'd rather go with it rather than have each author name things with different naming patterns.

Issue #2887, opened 2025-01-06
[Agentless] healthy agent without cloudbeat logs


orouz:
**Describe the bug**
on ESS we have an elastic-agent `8.16.1` (agentless) reporting healthy status but when looking into Cloudbeat logs there are none for an extended time period (weeks)

**To Reproduce**

1. Visit [agent page on ESS long running env](https://keep-long-live-env-ess.kb.us-west2.gcp.elastic-cloud.com/app/fleet/agents/9e08cee7-6be4-4e81-8a85-ac896f246ea3/details) and check the logs for Cloudbeat

**Expected behavior**
- Cloudbeat logs what it's currently doing or had failed to do.

**Additional context**

this agent is using a policy that has assets inventory installed and is misconfigured - all 3 cloud vendors are enabled at once, which is not supported. might not be related though.

kubasobon:
There's definitely an interval issue. Running locally with a `5m` period:
![Image](https://github.com/user-attachments/assets/518a9ec6-f07e-46e1-b3e5-e5738f10ad9c)
No results after 10 minutes besides the initial run.

kubasobon:
This checks out. Look at [this piece of code](https://github.com/elastic/cloudbeat/blob/b493e0c32173dc4d03da5e1ee56f1b2c9b9625c1/internal/inventory/inventory.go#L66-L102).

The workflow looks like:
1. Start `fetcher.Fetch()` in a goroutine for each fetcher
2. Run a loop that forever checks for new events unless the context is cancelled

The issue is `fetcher.Fetch()` will only ever run once. That's it.

kubasobon:
With the fix:
![Image](https://github.com/user-attachments/assets/d2c221f4-767f-4d98-ad30-310f26e28fcc)

Still does not explain missing logs, but it's a start. We should see `"Interval reached without events"` every 10 seconds outside of fetch windows.

kubasobon:
Fix for intervals in https://github.com/elastic/cloudbeat/pull/2902

kubasobon:
PR merged. `8.x` backport can be found here: https://github.com/elastic/cloudbeat/pull/2905

kubasobon:
Ok, backported to `8.x`, `8.16` and `8.17` branches.

kubasobon:
Going to go ahead and close this one as resolved.

PR #2902, opened 2025-01-07
[Asset Inventory] Run fetchers periodically

kubasobon:
### Summary of your changes

Fixes an issue where fetchers **ran only once**.

### Related Issues

See https://github.com/elastic/cloudbeat/issues/2887

### Checklist


internal/inventory/inventory.go
```diff
@@ -72,13 +80,22 @@ func (a *AssetInventory) Run(ctx context.Context) {
...
 			a.publish(assetsBuffer)
 			return
 
+		case <-fetcherPeriod.C:
+			a.logger.Debug("starting a new fetch cycle")
+			for _, fetcher := range a.fetchers {
+				go func(fetcher AssetFetcher) {
+					fetcher.Fetch(ctx, a.assetCh)
+				}(fetcher)
+			}
```
romulets, line 97: Shouldn't we remove the the fetcher loop on top of this function?

romulets (commented):
So, has it never worked? 😬

internal/inventory/inventory.go
kubasobon: Well, no. I added this so all fetchers run every `fetcherPeriod` tick.

kubasobon:
Yes, unfortunately it seems like it always ran just once and that's it.

internal/inventory/inventory.go
kubasobon: And the extra one on top is so that it starts immediately, on second 0. Otherwise it would take until the first `fetcherPeriod` tick to run first fetchers.

romulets: Hm true, just checked the Ticker and we can't make it run straight away. 

Nitpick: Maybe move this code to a function? Just so we have a nice wrapped function, it prevents if in the future we need to fix something and we do in one place only

romulets (approved):
I left a suggestion, but looks good otherwise

kubasobon:
I've extracted duplicate code to `runAllFetchersOnce()`. I hope the comment and function name make it even clearer that it's not a looping situation.

internal/inventory/inventory.go
```diff
@@ -29,13 +29,17 @@ import (
 	"github.com/samber/lo"
 )
 
-const indexTemplate = "logs-cloud_asset_inventory.asset_inventory-%s_%s_%s_%s-default"
+const (
+	indexTemplate = "logs-cloud_asset_inventory.asset_inventory-%s_%s_%s_%s-default"
+	minimalPeriod = 30 * time.Second
```
orouz, line 34: maybe can omit this const? 

locally we can tweak 
https://github.com/elastic/cloudbeat/blob/56579fcae8736f2fb6775ff11aa8b2702a9a8e3e/deploy/asset-inventory/cloudbeat-gcp-asset-inventory.yml#L14

and users have a hard-coded value, for example - https://github.com/elastic/integrations/blob/56d0473a2ed4900b7b24b6a254795b9bc82cb713/packages/cloud_asset_inventory/data_stream/asset_inventory/agent/stream/gcp.yml.hbs#L1

kubasobon: I would rather keep it. It's not a default value, but a guardian value. If the customer sets value too low, this will start to spawn many goroutines that hit API. I can lower it if you think it makes sense, but we **definitely** need to make sure there is a reasonable minimum threshold. Obviously the worst-case scenario is `period: 0s`, where we get an infinite loop of new goroutines, but even `5s` or `10s` is too small IMHO.

kubasobon: PTAL: https://github.com/elastic/cloudbeat/pull/2902/files#diff-151451acd3d777673443e215b546390195bf895e79500c18a34bdf0029222fbfR57-R59

internal/inventory/inventory.go
moukoublen: A note: in manager (used in fetching of cspm mostly), we use a timer, first with a `0` value and then reset with the period value, to do the trick of "trigger now and then per period".

https://github.com/elastic/cloudbeat/blob/b493e0c32173dc4d03da5e1ee56f1b2c9b9625c1/internal/resources/fetching/manager/manager.go#L72-L87

I'm not saying we should do the same with asset inventory since I think the ticker is better for what we want.

moukoublen (approved):

internal/inventory/inventory.go
orouz: np with keeping it imo. but this value isn't configurable by the users, it's hardcoded in the integration package.

internal/inventory/inventory.go
kubasobon: Thanks. Yeah, I know the trick, but I always find calling `timer.Reset()` every time more messy. Nothing against the implementation, just personal taste I guess.

PR #2920, opened 2025-01-16
Add support for VERSION_QUALIFIER DRA

moukoublen:
### Summary of your changes
Add Support for VERSION_QUALIFIER argument during DRA production. 

Add a temporary condition to staging artifacts so can be triggered even from main branch when VERSION_QUALIFIER is passed. (required to upload the first `alpha1` pre release). 

Test for artifacts generation done locally, but the buildkite pipeline will be tested after merge. 

### Screenshot/Data
Screenshots after running what buildkite pipeline runs plus VERSION_QUALIFIER env var.
```bash
export VERSION_QUALIFIER='alpha1'
export WORKFLOW='staging'
bash ./.buildkite/scripts/package.sh 
```
Screenshot 2025-01-16 at 12 42 19 PM
Screenshot 2025-01-16 at 12 42 07 PM
Screenshot 2025-01-16 at 2 38 51 PM
Screenshot 2025-01-16 at 2 43 05 PM

### Related Issues

Fixes: https://github.com/elastic/security-team/issues/11536

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/.buildkite/scripts/package.sh
+++ b/.buildkite/scripts/package.sh
@@ -8,6 +8,7 @@ source ./bin/activate-hermit
 CLOUDBEAT_VERSION=$(grep defaultBeatVersion version/version.go | cut -d'=' -f2 | tr -d '" ')
 PYTHON_BIN=./build/ve/$(go env GOOS)/bin
 PYTHON=$PYTHON_BIN/python
+VERSION_QUALIFIER="${VERSION_QUALIFIER:-""}"
 
 if [ "$WORKFLOW" = "snapshot" ]; then
     export SNAPSHOT="true"
@@ -21,6 +22,9 @@ mage package
 
 CSV_FILE="build/dependencies-${CLOUDBEAT_VERSION}"
 [ -n "${SNAPSHOT+x}" ] && CSV_FILE+="-SNAPSHOT"
+if [[ -n "$VERSION_QUALIFIER" ]]; then
+  CSV_FILE+="-${VERSION_QUALIFIER}"
+fi
 
 echo "Generating $CSV_FILE.csv"
 $PYTHON ./.buildkite/scripts/generate_notice.py --csv "$CSV_FILE.csv"
--- a/.buildkite/scripts/package.sh
+++ b/.buildkite/scripts/package.sh
@@ -23,7 +23,7 @@ mage package
 CSV_FILE="build/dependencies-${CLOUDBEAT_VERSION}"
 [ -n "${SNAPSHOT+x}" ] && CSV_FILE+="-SNAPSHOT"
 if [[ -n "$VERSION_QUALIFIER" ]]; then
-  CSV_FILE+="-${VERSION_QUALIFIER}"
+    CSV_FILE+="-${VERSION_QUALIFIER}"
 fi
 
 echo "Generating $CSV_FILE.csv"
--- a/magefile.go
+++ b/magefile.go
@@ -49,8 +49,9 @@ import (
 )
 
 const (
-	snapshotEnv   = "SNAPSHOT"
-	agentDropPath = "AGENT_DROP_PATH"
+	snapshotEnv         = "SNAPSHOT"
+	agentDropPath       = "AGENT_DROP_PATH"
+	cloudbeatModulePath = "github.com/elastic/cloudbeat"
 )
 
 func init() {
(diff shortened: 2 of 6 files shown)


magefile.go
```diff
@@ -80,6 +81,9 @@ func Build() error {
 
 	args := devtools.DefaultBuildArgs()
 	args.CGO = false
+	if versionQualifier, versionQualified := os.LookupEnv("VERSION_QUALIFIER"); versionQualified {
```
romulets, line 84: I don't really understand this change.

If the `VERSION_QUALIFIER` env var is present, we set in the args map a key of a file name and value the qualifier. What does this do to the build? Includes a file in the build with the value as contents?

version/version.go
```diff
@@ -20,6 +20,8 @@ package version
 // name matches github.com/elastic/beats/v7/dev-tools/mage/settings.go parseBeatVersion
 const defaultBeatVersion = "9.0.0"
 
+var qualifier = ""
```
romulets, line 23: Will we manually update this upon new qualifiers? Why not a const?

.buildkite/scripts/package.sh
```diff
@@ -8,6 +8,7 @@ source ./bin/activate-hermit
 CLOUDBEAT_VERSION=$(grep defaultBeatVersion version/version.go | cut -d'=' -f2 | tr -d '" ')
 PYTHON_BIN=./build/ve/$(go env GOOS)/bin
 PYTHON=$PYTHON_BIN/python
+VERSION_QUALIFIER="${VERSION_QUALIFIER:-""}"
```
romulets, line 11: I'm trying to understand why here you used `:-` and on publish.sh `:=`

Reading https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html I think we should be using `:-` on both for substitution in case of not available?

magefile.go
moukoublen: No, what we do here is we set the variable `qualifier` that exists in `version.go`  at build time, (using golang `-ldflags` `-X`) 

https://pkg.go.dev/cmd/link#:~:text=%2DX%20importpath.name%3Dvalue

then because of that function change: 

```golang
// CloudbeatSemanticVersion returns the current cloudbeat version.
func CloudbeatSemanticVersion() string {
	if qualifier != "" {
		return defaultBeatVersion + "-" + qualifier
	}
	return defaultBeatVersion
}
```

This is displayed with we run `cloudbeat version` 
https://private-user-images.githubusercontent.com/1380039/403865316-ddb14708-33ca-4745-910a-a88d1739d324.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzcwNDQyMTcsIm5iZiI6MTczNzA0MzkxNywicGF0aCI6Ii8xMzgwMDM5LzQwMzg2NTMxNi1kZGIxNDcwOC0zM2NhLTQ3NDUtOTEwYS1hODhkMTczOWQzMjQucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDExNiUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTAxMTZUMTYxMTU3WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NTA0ODc1OWNlYmNlYjc3N2M3OGMzZTQ1ZWU5YWI5YmRhYTQyZjVmNzRhMTYzMWIzNGRiNDkzYWNmZWE0NDQ0NyZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.TbzbRlHauv8dZ0mm-j7ss5kvW9vkKA-HQbxJt3s6DKU

version/version.go
moukoublen: Its a variable that is set at build time, it can't be const.

.buildkite/scripts/package.sh
moukoublen: Essentially, there is no difference between the two since, in both cases, we assign the outcome of the expansion back to the variable. 

We would have different result if we didn't assign the expansion back to the variable.


I will make both the same for consistency.

oren-zohar (commented):
Thank you @moukoublen. Anything to update in our internal release manager docs?

moukoublen:
@oren-zohar Yes, I will create a PR for the docs tomorrow.

magefile.go
romulets: Super interesting! Thanks for explaining, I've never used this before!

romulets (approved):

PR #2948, opened 2025-01-26
Enrich GCP asset types with ECS fields

orouz:
### Summary of your changes

- adds specific asset type enrichers
- collects `tags` and `labels` for all assets (if available)

### Screenshot/Data

see [comment](https://github.com/elastic/security-team/issues/10435#issuecomment-2629356829)

### Related Issues
- closes https://github.com/elastic/security-team/issues/10435

### DIFF
--- a/internal/inventory/asset.go
+++ b/internal/inventory/asset.go
@@ -184,8 +184,6 @@ type Network struct {
 	Name      string `json:"name,omitempty"`
 	Direction string `json:"direction,omitempty"`
 	Type      string `json:"type,omitempty"`
-	// Protocol string `json:"protocol,omitempty"`
-	// Transport []string `json:"transport,omitempty"`
 }
 
 type Cloud struct {
--- a/internal/inventory/asset.go
+++ b/internal/inventory/asset.go
@@ -139,11 +139,31 @@ type AssetEvent struct {
 	Entity        Entity
 	Event         Event
 	Network       *Network
+	URL           *URL
+	Organization  *Organization
 	Cloud         *Cloud
+	Fass          *Fass
+	Orchestrator  *Orchestrator
+	Container     *Container
 	Host          *Host
 	User          *User
 	Labels        map[string]string
+	Tags          []string
 	RawAttributes *any
+}
+
+type Organization struct {
+	ID   string `json:"id,omitempty"`
+	Name string `json:"name,omitempty"`
+}
+
+type Fass struct {
+	Name    string `json:"name,omitempty"`
+	Version string `json:"version,omitempty"`
+}
+
+type URL struct {
+	Full string `json:"full"`
 }
 
 // Entity contains the identifiers of the asset
@@ -161,7 +181,11 @@ type Event struct {
 }
 
 type Network struct {
-	Name string `json:"name,omitempty"`
+	Name      string `json:"name,omitempty"`
+	Direction string `json:"direction,omitempty"`
+	Type      string `json:"type,omitempty"`
+	// Protocol string `json:"protocol,omitempty"`
+	// Transport []string `json:"transport,omitempty"`
 }
 
 type Cloud struct {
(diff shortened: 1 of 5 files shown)


internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -172,3 +194,173 @@ func appendIfExists(slice []string, fields map[string]*structpb.Value, key strin
...
+}
+
+func enrichFirewall(_ *gcpinventory.ExtendedGcpAsset, fields map[string]*structpb.Value) []inventory.AssetEnricher {
+	return []inventory.AssetEnricher{
+		inventory.WithNetwork(inventory.Network{
+			Name:      getStringValue("name", fields),
+			Direction: getStringValue("direction", fields),
+		}),
+	}
+}
```
orouz, line 302: we might be able to provide `network.transport` as well

the issue is that in GCP, an applicable value is at [`allowed[].IPProtocol`](https://cloud.google.com/compute/docs/reference/rest/v1/firewalls#:~:text=allowed%5B%5D.IPProtocol), meaning that there is an array of values to collect (not a single one)

also, if we do, it's unclear of `denied` rules should be collected as well (i think not, but not sure)

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -172,3 +194,173 @@ func appendIfExists(slice []string, fields map[string]*structpb.Value, key strin
...
+	return labelsMap
+}
+
+var assetEnrichers = map[string]func(item *gcpinventory.ExtendedGcpAsset, fields map[string]*structpb.Value) []inventory.AssetEnricher{
+	gcpinventory.IamRoleAssetType:               noopEnricher,
+	gcpinventory.CrmFolderAssetType:             noopEnricher,
+	gcpinventory.CrmProjectAssetType:            noopEnricher,
+	gcpinventory.StorageBucketAssetType:         noopEnricher,
+	gcpinventory.IamServiceAccountKeyAssetType:  noopEnricher,
+	gcpinventory.CloudRunService:                noopEnricher,
```
orouz, line 255: cloud run has [applicable values](https://cloud.google.com/run/docs/reference/rest/v1/Container) we can collect, the issue is that they are provided in `asset.spec.template.spec.containers[].*`, so each value we can technically get (`name`, `image`) would be an array, as opposed to the expected single value

orouz: in general, we could probably collect more ECS fields from each asset types if we dig deep enough

kubasobon:
Checking, it's a big PR 👀 😅

internal/inventory/asset.go
```diff
@@ -161,7 +181,11 @@ type Event struct {
 }
 
 type Network struct {
-	Name string `json:"name,omitempty"`
+	Name      string `json:"name,omitempty"`
+	Direction string `json:"direction,omitempty"`
+	Type      string `json:"type,omitempty"`
+	// Protocol string `json:"protocol,omitempty"`
+	// Transport []string `json:"transport,omitempty"`
```
kubasobon, line 188: Is this a leftover from development or future-proofing?

internal/inventory/gcpfetcher/fetcher_assets.go
kubasobon: Thanks for the clarification 👍

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -172,3 +194,173 @@ func appendIfExists(slice []string, fields map[string]*structpb.Value, key strin
...
+			Direction: getStringValue("direction", fields),
+		}),
+	}
+}
+
+func enrichSubnetwork(_ *gcpinventory.ExtendedGcpAsset, fields map[string]*structpb.Value) []inventory.AssetEnricher {
+	return []inventory.AssetEnricher{
+		inventory.WithNetwork(inventory.Network{
+			Name: getStringValue("name", fields),
+			Type: strings.ToLower(getStringValue("stackType", fields)),
```
kubasobon, line 308: Do we need to lower the stack type here? Why not keep it consistent with what we get from upstream?

kubasobon (approved):
Just a couple of comments, LGTM otherwise

internal/inventory/asset.go
orouz: we might be able to add those but i'll remove it from this PR. maybe next iteration

Issue #2556, opened 2024-09-24
[Cloud Connectors] Configuration and Flow


moukoublen:
**Motivation**
We need to create a separate configuration for cloud connectors. In that case an indication that cloud connectors is used should be provided (e.g. a `enabled` flag) and the customer remote-role arn should be also proved via configuration.

During the flow and if the cloud connectors is enabled, cloudbeat should first assume the implicit role provided by the EKS POD Identity association and then having assumed that cloudbeat should assume the customer remote role.

Using the credentials from the customer remote-role cloudbeat should continue to the scan flow.

POC branch: https://github.com/moukoublen/cloudbeat/tree/cc_8.14

**Definition of done**
- AWS Single account flow
- AWS Org account flow
- CSPM Integration configuration 

**Related tasks/epics**
[Reference related issues and epics](https://github.com/elastic/security-team/issues/10552)

moukoublen:
@eyalkraft I think for the remote `assume` to work we need a default config first (so aws SDK to get the token provided by the EKS Pod Identithy agent from the k8s mounted volume) using `awsconfig.LoadDefaultConfig` and then using that credentials (the elastic `super-role`) assume the remote one. 

I think using the already implemented flow of "role arn to assume" didn't work for me in POC, but I didn't put the effort in, so perhaps it's something else that needs fixing. 


So, I will check the flow of "role to assume" as it is implemented from `gibbet' further and post my findings here.

PR #2960, opened 2025-01-29
cloud connectors role chaining

moukoublen:
### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/2556

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/internal/resources/providers/awslib/config.go
+++ b/internal/resources/providers/awslib/config.go
@@ -59,46 +59,65 @@ func CloudConnectorsExternalID(resourceID, externalIDPart string) string {
 }
 
 func InitializeAWSConfigCloudConnectors(ctx context.Context, cfg config.AwsConfig) (*aws.Config, error) {
-	// 1. Load initial config - Chain Part 1 - Elastic Super Role Local implicitly assumed.
+	const defaultDuration = 20 * time.Minute
+
+	// 1. Load initial config - Chain Step 1 - Elastic Super Role Local implicitly assumed through IRSA.
 	awsConfig, err := awsconfig.LoadDefaultConfig(ctx)
 	if err != nil {
 		return nil, err
 	}
 
-	// Create an STS client using the base credentials
-	firstClient := sts.NewFromConfig(awsConfig)
-
-	const defaultDuration = 20 * time.Minute
-
-	// Chain Part 2 - Elastic Super Role Global
-	globalSuperRoleProvider := stscreds.NewAssumeRoleProvider(
-		firstClient,
-		cfg.CloudConnectorsConfig.GlobalRoleARN,
-		func(aro *stscreds.AssumeRoleOptions) {
-			aro.RoleSessionName = "cloudbeat-super-role-global"
-			aro.Duration = defaultDuration
+	chain := []AWSRoleChainingStep{
+		// Chain Step 2 - Elastic Super Role Global
+		{
+			RoleARN: cfg.CloudConnectorsConfig.GlobalRoleARN,
+			Options: func(aro *stscreds.AssumeRoleOptions) {
+				aro.RoleSessionName = "cloudbeat-super-role-global"
+				aro.Duration = defaultDuration
+			},
 		},
-	)
-	globalSuperRoleCredentialsCache := aws.NewCredentialsCache(globalSuperRoleProvider)
-
-	// Chain Part 3 - Elastic Super Role Local
-	customerRemoteRoleCfg := awsConfig
-	customerRemoteRoleCfg.Credentials = globalSuperRoleCredentialsCache
-	customerRemoteRoleProvider := stscreds.NewAssumeRoleProvider(
-		sts.NewFromConfig(customerRemoteRoleCfg),
-		cfg.Cred.RoleArn, // Customer Remote Role passed in package policy.
-		func(aro *stscreds.AssumeRoleOptions) {
-			aro.RoleSessionName = "cloudbeat-remote-role"
-			aro.Duration = cfg.Cred.AssumeRoleDuration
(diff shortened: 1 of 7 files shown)


kubasobon (approved):

internal/resources/providers/awslib/config.go
```diff
@@ -18,28 +18,102 @@
...
 }
+
+func CloudConnectorsExternalID(resourceID, externalIDPart string) string {
+	return fmt.Sprintf("%s-%s", resourceID, externalIDPart)
+}
+
+func InitializeAWSConfigCloudConnectors(ctx context.Context, cfg config.AwsConfig) (*aws.Config, error) {
+	// 1. Load initial config
+	// (TODO: check directly assuming the first role in chain and/or libbeataws.InitializeAWSConfig(cfg))
+	// (TODO: consider os.Setenv("AWS_EC2_METADATA_DISABLED", "true"))
```
olegsu, line 64: I am not sure what to consider, to set it or to check the behavior when it was previously set?
Also, in MKI we block the IMDS ip address, how it affect the flow?

internal/resources/providers/awslib/config.go
```diff
@@ -18,28 +18,102 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Create an STS client using the base credentials
+	firstClient := sts.NewFromConfig(awsConfig)
+
+	const defaultDuration = 5 * time.Minute
+
+	// Chain Part 1 - Elastic Super Role Local
```
olegsu, line 75: Super role is misleading term, indicate that the role permissions are elevated when we need only [audit](https://www.elastic.co/guide/en/security/current/cspm-get-started.html#cspm-set-up-cloud-access-section) ([SecurityAudit built-in AWS](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_job-functions.html#jf_security-auditor))

internal/resources/providers/awslib/config.go
```diff
@@ -18,28 +18,102 @@
...
+		firstClient,
+		cfg.CloudConnectorsConfig.LocalRoleARN,
+		func(aro *stscreds.AssumeRoleOptions) {
+			aro.RoleSessionName = "cloudbeat-super-role-local"
+			aro.Duration = defaultDuration
+		},
+	)
+	localSuperRoleCredentialsCache := aws.NewCredentialsCache(localSuperRoleProvider)
+
+	// Chain Part 2 - Elastic Super Role Global
```
olegsu, line 86: Super role is misleading term, indicate that the role permissions are elevated when we need only [audit](https://www.elastic.co/guide/en/security/current/cspm-get-started.html#cspm-set-up-cloud-access-section) ([SecurityAudit built-in AWS](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_job-functions.html#jf_security-auditor))

internal/resources/providers/awslib/config.go
```diff
@@ -18,28 +18,102 @@
...
+		sts.NewFromConfig(globalSuperRoleCfg),
+		cfg.CloudConnectorsConfig.GlobalRoleARN,
+		func(aro *stscreds.AssumeRoleOptions) {
+			aro.RoleSessionName = "cloudbeat-super-role-global"
+			aro.Duration = defaultDuration
+		},
+	)
+	globalSuperRoleCredentialsCache := aws.NewCredentialsCache(globalSuperRoleProvider)
+
+	// Chain Part 3 - Elastic Super Role Local
```
olegsu, line 99:
```suggestion
	// Chain Part 3 - Elastic Remote SecurityAudit
```

internal/resources/providers/awslib/config.go
```diff
@@ -18,28 +18,102 @@
...
+	awsConfig.Retryer = awsConfigRetrier
 
 	return &awsConfig, nil
 }
+
+func CloudConnectorsExternalID(resourceID, externalIDPart string) string {
+	return fmt.Sprintf("%s-%s", resourceID, externalIDPart)
+}
+
+func InitializeAWSConfigCloudConnectors(ctx context.Context, cfg config.AwsConfig) (*aws.Config, error) {
```
olegsu, line 61: I wonder if we better to have a more generic flow in case we need longer chain. 
The chain can be part of the agent policy and be used also for non cloud connector use cases

internal/resources/providers/awslib/config.go
moukoublen: I believe this is something to discuss in RFC level, wdyt?

internal/resources/providers/awslib/config.go
moukoublen: The global super role has not the audit policy, it has no policy part from the "assume anything". Feel free to refer RFC regarding the terminology.

internal/resources/providers/awslib/config.go
moukoublen: The comment was for me (I will remove them) to check id we need to use the AWS_EC2_METADATA_DISABLED in case we needed to remove the firewall rules. We don't (we checked the flow with @amirbenun a while ago) so all good. 

No changes needed. I will remove both TODOs.

internal/resources/providers/awslib/config.go
moukoublen: I thought about implementing it in a "loop" mode, but: 

1. this will eventually be moved to libbeat so a more generic approach can be implemented there
2. After some tests that I just completed, there are even fewer steps now; it's the init and 2 extras assume instead of 3. 

I believe a more generic approach (if it is worth the effort) can be implemented when that code gets transferred to libbeat. 

Wdyt?

moukoublen: Implemented it in a loop-like logic.

PR #2989, opened 2025-02-05
mergify: support backport-active-8, backport-active-9, backport-active-all

v1v:
### Summary of your changes

Enable `backport-active-all`, `backport-active-8` and `backport-active-9` GitHub labels in Mergify to backport to a set of active branches in addition to the existing automation with the `backport-<major>.<minor>` labels. This can help with the current 6 active branches, which we have today.

See https://docs.mergify.com/workflow/actions/backport/#parameters

image

I can find `regexes` too:

image

But I'm not sure whether that can solve our current release model easily

### Why is it important?

> Maybe this is solved already, but is there a way we can get an automatically kept up to date “backport-active-minors” label that auto backports to the set of minor releases that are still supported? I’d want this in Beats, Elastic Agent, and Fleet Server if it already exists. I’ve noticed an increase in forgotten backports because there are more releases to keep track of. If we can abstract this it’d make life easier, I think Kibana does something like this

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


.mergify.yml
```diff
@@ -378,3 +381,56 @@ pull_request_rules:
...
+          - "{{ author }}"
+        # NOTE: this list needs to be changed when a new minor branch is created
+        #       or an existing release branch reached EOL.
+        branches:
+          - "9.0"
+          - "8.18"
+          - "8.17"
+          - "8.16"
+          - "8.x"
+          - "7.17"
```
moukoublen, line 433:
```suggestion
          - "7.17"
```

Since cloudbeat does not have `7.17` branch perhaps we can remove that?

v1v: That's an option, I kept as I was figuring out when running the automation to update that list , it will be much simpler if using https://storage.googleapis.com/artifacts-api/snapshots/branches.json without doing any specific manipulation.

However, since `7.17` EOL might happen eventually, maybe we can remove it. Up to you :)

moukoublen (approved):

PR #3017, opened 2025-02-13
Ensure cloudbeat is installed with >= 9.x elastic-agent

kubasobon:
### Summary of your changes

> [!warning] 
> Do **not** backport this PR to 8.x

Adding `--install-servers` flag to deployment templates so that cloudbeat is installed in v9.x elastic-agent.

### Screenshot/Data

Screenshot 2025-02-13 at 10 36 20

_1. Showcasing the new `elastic-agent install` flag in 9.x_

Screenshot 2025-02-13 at 13 55 08

_2. Showing installed v9.0.0-beta1 agent_

### Related Issues

Towards https://github.com/elastic/security-team/issues/11500

### DIFF
--- a/tests/integration/tests/test_sanity_checks.py
+++ b/tests/integration/tests/test_sanity_checks.py
@@ -22,7 +22,7 @@ from configuration import elasticsearch
 from loguru import logger
 
 CONFIG_TIMEOUT = 120
-GCP_CONFIG_TIMEOUT = 600
+GCP_CONFIG_TIMEOUT = 1200
 CNVM_CONFIG_TIMEOUT = 3600
 
 # The timeout and backoff for waiting all agents are running the specified component.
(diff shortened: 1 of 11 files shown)


tests/integration/tests/test_sanity_checks.py
```diff
@@ -304,18 +304,29 @@ def wait_components_list(actual: AgentComponentMapping, expected: AgentExpectedM
     if not elasticsearch.kibana_url:
         return [""]
 
-    actual.load_map()
+    try_loading_map(actual)
```
kubasobon, line 307: @gurevichdmitry This is the fix. When we call `actual.load_map()` too soon we will get a response, but it will be missing `components` key. I've wrapped the method to handle errors like this and now it successfully awaits `actual` to be populated. Not sure why this came up just now, but it's all in timing.

gurevichdmitry (approved):

Issue #3018, opened 2025-02-16
Enrich asset types with multiple values


orouz:
**Motivation**
for `keyword` ECS fields, we [should](https://github.com/elastic/cloudbeat/pull/2948#discussion_r1939717606) collect multiple values instead of singular

GCP Examples:

| type 	| ECS 	| GCP data  	|
|--------	|-------	|---	|
| `run.googleapis.com/Service`     	| `container.name`     	|   `asset.spec.template.spec.containers[].name`	|
| `run.googleapis.com/Service`      	| `container.image.name`     	|   `asset.spec.template.spec.containers[].image`	|
| `compute.googleapis.com/Instance`      	| `network.name`     	|   `asset.networkInterfaces[].name`	|

**Definition of done**
- map currently used ECS fields of the `keyword` type and enable asset enrichers to assign multiple values

**Options**

this could be a simple `string` -> `[]string` replacement or maybe something more fancy.

orouz:
after discussing with product, the decision is to not send multiple values to ECS where a single value is expected. the raw data would still be available. #3028 will be modified to do so.

PR #3028, opened 2025-02-17
Enrich GCP asset types with multiple values

orouz:
### Summary of your changes

1. adds cloud run enricher
2. adds utility function to traverse maps/arrays and extract leaf string values

with this PR, all asset types required for GA are fetched and enriched plus additional enrichments we may want to include next would be easy to collect.

### Screenshot/Data
WIP

### Related Issues
- closes https://github.com/elastic/cloudbeat/issues/3018

### DIFF
--- a/internal/inventory/asset.go
+++ b/internal/inventory/asset.go
@@ -189,9 +189,9 @@ type Event struct {
 }
 
 type Network struct {
-	Name      string `json:"name,omitempty"`
-	Direction string `json:"direction,omitempty"`
-	Type      string `json:"type,omitempty"`
+	Name      []string `json:"name,omitempty"`
+	Direction string   `json:"direction,omitempty"`
+	Type      string   `json:"type,omitempty"`
 }
 
 type Cloud struct {
@@ -237,9 +237,9 @@ type Orchestrator struct {
 }
 
 type Container struct {
-	ID        string `json:"id,omitempty"`
-	Name      string `json:"name,omitempty"`
-	ImageName string `json:"image.name,omitempty"`
+	ID        string   `json:"id,omitempty"`
+	Name      []string `json:"name,omitempty"`
+	ImageName []string `json:"image.name,omitempty"`
 }
 
 // AssetEnricher functional builder function
--- a/internal/inventory/gcpfetcher/fetcher_assets.go
+++ b/internal/inventory/gcpfetcher/fetcher_assets.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/samber/lo"
-	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/elastic/cloudbeat/internal/infra/clog"
 	"github.com/elastic/cloudbeat/internal/inventory"
@@ -112,7 +111,7 @@ func getAssetEvent(classification inventory.AssetClassification, item *gcpinvent
 	// Asset type specific enrichers
 	if hasResourceData(item) {
 		if enricher, ok := assetEnrichers[item.AssetType]; ok {
-			enrichers = append(enrichers, enricher(item, item.GetResource().GetData())...)
+			enrichers = append(enrichers, enricher(item, item.GetResource().GetData().AsMap())...)
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


internal/inventory/asset.go
```diff
@@ -158,6 +158,11 @@ type AssetEvent struct {
 	Labels        map[string]string
 	Tags          []string
 	RawAttributes *any
+	Related       *Related
+}
+
+type Related struct {
+	Id []string `json:"id,omitempty"`
```
orouz, line 165: in order to test `WithRelatedAssetIds` i needed an exported field, so i replaced `Entity.relatedEntityId` with `Related.Id`

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -125,54 +123,37 @@ func getAssetEvent(classification inventory.AssetClassification, item *gcpinvent
...
+		ids = append(ids, values([]string{"networkInterfaces", "network"}, pb)...)
+		ids = append(ids, values([]string{"networkInterfaces", "subnetwork"}, pb)...)
+		ids = append(ids, values([]string{"serviceAccounts", "email"}, pb)...)
+		ids = append(ids, values([]string{"disks", "source"}, pb)...)
+		ids = append(ids, values([]string{"machineType"}, pb)...)
+		ids = append(ids, values([]string{"zone"}, pb)...)
+
+	case gcpinventory.ComputeFirewallAssetType, gcpinventory.ComputeSubnetworkAssetType:
+		ids = append(ids, values([]string{"network"}, pb)...)
+	case gcpinventory.CrmProjectAssetType, gcpinventory.StorageBucketAssetType:
```
orouz, line 156: i'm not sure why we're adding `binding.Role` and `binding.Members` just for projects and buckets asset types. 

unrelated to this PR tho

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -263,104 +215,148 @@ var assetEnrichers = map[string]func(item *gcpinventory.ExtendedGcpAsset, fields
...
+
+func noopEnricher(_ *gcpinventory.ExtendedGcpAsset, _ map[string]any) []inventory.AssetEnricher {
 	return []inventory.AssetEnricher{}
 }
 
-func getStringValue(key string, f map[string]*structpb.Value) string {
-	if value, ok := f[key]; ok {
-		return value.GetStringValue()
+// returns string values of keys in a map/array
+func values(keys []string, current any) []string {
```
orouz, line 328: extracting values from a protobuf is kind of a pain. (`GetStructValue().GetFields()`, `GetListValue().GetValues()`)

this function simplifies extraction by enabling map/array traversal with provided keys and defaults to returning an empty slice. 

for example, this is how we collect values for a container name:
```go
containerNames: = values([]string{"spec", "template", "spec", "containers", "name"}, pb)
```

instead of:
```go
var containerNames []string

specField, ok := s.Fields["spec"]
if !ok {
  return nil, fmt.Errorf("'spec' field not found")
}

templateField, ok := specField.GetStructValue().Fields["template"]
if !ok {
  return nil, fmt.Errorf("'template' field not found")
}

specTemplateField, ok := templateField.GetStructValue().Fields["spec"]
if !ok {
  return nil, fmt.Errorf("'spec' field inside template not found")
}

containersField, ok := specTemplateField.GetStructValue().Fields["containers"]
if !ok {
  return nil, fmt.Errorf("'containers' field not found")
}

containersList, ok := containersField.GetListValue().GetValues()
if !ok {
  return nil, fmt.Errorf("'containers' is not an array")
}

for _, container := range containersList {
  if nameField, ok := container.GetStructValue().Fields["name"]; ok {
    containerNames = append(containerNames, nameField.GetStringValue())
  }
}

return containerNames
```

orouz: um yeah definitely, panic is bad™, can you share an example when will that happen? currently there are 2 `return []string{}` at the end of each switch case, which i thought was sufficient

kubasobon:
One more thing - it probably needs `backport-v9.x` label as well.

internal/inventory/asset.go
```diff
@@ -179,19 +184,16 @@ type Entity struct {
...
 
 type Event struct {
 	Kind string `json:"kind"`
 }
 
 type Network struct {
-	Name      string `json:"name,omitempty"`
-	Direction string `json:"direction,omitempty"`
-	Type      string `json:"type,omitempty"`
+	Name      []string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
orouz: this is a very good point, thanks for bringing this up. i was under the impression this was legit, and each value will be indxed as a keyword, which is correct, but still has issues, mainly - correlating data. for example, in Cloud Run Service, we can have multiple values for `container.name` and `container.image`, and sending both as an array would work but the user would lose the ability to know which container name matches which container image.

i'm going to revert these changes and i'll bring up the issue for discussion in our sync. what we can do though, is either not collect values in the wild that don't match the ECS schema (bad option) or once we know multiple values are available, send more than 1 documents. so in the case of Cloud Run Service, if we have `containers[].image/name` - we would send as many documents as `containers[].length` is. a bit of a stretch but this would provide the most value for the user.

internal/inventory/gcpfetcher/fetcher_assets.go
```diff
@@ -263,104 +215,141 @@ var assetEnrichers = map[string]func(item *gcpinventory.ExtendedGcpAsset, fields
...
```
orestisfl, line 328: This is a hard-to-understand function. If that's the way we are going, I suggest to cover it with dedicated unit tests:
```go
func Test_values(t *testing.T) {
	tests := []struct {
		value any
		keys  []string
		want  []string
	}{
		{
			want: []string{},
		},
		{
			value: 1,
			keys:  nil,
			want:  []string{},
		},
		{
			value: "x",
			keys:  nil,
			want:  []string{"x"},
		},
		{
			value: map[string]any{"name": "value"},
			keys:  []string{"not-exist"},
			want:  []string{},
		},
		{
			value: map[string]any{"name": "value"},
			keys:  []string{"name"},
			want:  []string{"value"},
		},
		{
			value: map[string]any{"name": map[string]any{"name": "value"}},
			keys:  []string{"name", "name"},
			want:  []string{"value"},
		},
		{
			value: []any{"some string", map[string]any{"name": map[string]any{"name": "value"}}, map[string]any{"name": map[string]any{"name": "value"}}},
			keys:  []string{"name", "name"},
			want:  []string{"value", "value"},
		},
		{
			value: map[string]any{"name": []any{"a", "b", "c"}},
			keys:  []string{"name"},
			want:  []string{"a", "b", "c"},
		},
	}
	for _, tt := range tests {
		name := fmt.Sprintf("%v(%v)=>%v", tt.value, tt.keys, tt.want)
		t.Run(name, func(t *testing.T) {
			assert.Equal(t, tt.want, values(tt.keys, tt.value))
		})
	}
}
```

orouz: thanks for the feedback @orestisfl! there is a test for the function, but i'll make sure to add more cases and clean it up based on your example up so it's easier to understand.

i agree the function is a bit of a stretch, and i am hesitant to add it, but it does make collecting values [nicer](https://github.com/elastic/cloudbeat/pull/3028#discussion_r1959318267) 

wdyt? would you prefer if we stick to conventional extraction of values instead of "smart" functions?

PR #3030, opened 2025-02-17
Change updatdcli to run monthly

orouz:
weekly is too much noise


kubasobon (approved):
Should we backport this to 8.x and 9.x?

orouz:
@kubasobon i think merging to `main` and `8.x` is enough as those are the main branches

PR #3086, opened 2025-03-06
Fix ECH and Serverless test workflows

gurevichdmitry:
### Summary of your changes

This PR fixes and improves the following:

- Retrieve stack version from the Hermit environment instead of using a hardcoded value.
- Ensure the correct version is used in Kubernetes manifests for both EKS and vanilla clusters.
- Increase the AWS provider `max_retries` counter.
- Update tests to run queries with the agent version only on known hosts.

### Screenshot/Data

Serverless workflow [successful run](https://github.com/elastic/cloudbeat/actions/runs/13702829516)
ECH workflow [successful run](https://github.com/elastic/cloudbeat/actions/runs/13759523856/job/38474293507)

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


tests/integration/tests/test_sanity_checks.py
```diff
@@ -181,7 +181,6 @@ def test_cspm_aws_findings(
         query_list = build_query_list(
             benchmark_id="cis_aws",
             match_type=match_type,
-            version=AGENT_VERSION,
             agent=agent,
```
gurevichdmitry, line 184: Sanity checks are used for both `agent-based` and `agentless` deployments. Since we don't have control over the agentless agent version deployment, we cannot use this parameter in the query. However, we can still correctly identify the retrieved documents because we use `agent.id` for identification.

bin/hermit.hcl
```diff
@@ -1,4 +1,4 @@
 env = {
-  "CLOUDBEAT_VERSION": "9.0.0",
+  "CLOUDBEAT_VERSION": "9.1.0",
```
oren-zohar, line 2: This will be backported to 8.x?

gurevichdmitry: No, I'll remove this change, and the version bump will be done in a [separate PR](https://github.com/elastic/cloudbeat/pull/3088).

deploy/test-environments/cis/main.tf
```diff
@@ -1,5 +1,6 @@
 provider "aws" {
-  region = var.region
+  region      = var.region
+  max_retries = 30
```
orouz, line 3: why is this needed? 30 seems kind of a lot

.github/workflows/weekly-serverless.yml
```diff
@@ -11,12 +11,22 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       deployment_name: ${{ steps.set_deployment_name.outputs.date-name }}
+      stack_version: ${{ steps.set_deployment_name.outputs.stack-version }}
     steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+
+      - name: Hermit Environment
+        uses: ./.github/actions/hermit
```
orouz, line 20: i don't see anything new that requires it. what am i missing?

.github/workflows/weekly-serverless.yml
```diff
@@ -11,12 +11,22 @@ jobs:
...
+
       - name: Set deployment name
         id: set_deployment_name
+        env:
+          STACK_VERSION: ${{ env.ELK_VERSION }}
```
orouz, line 25: is this set by hermit? is this why we added that step?

gurevichdmitry: Exactly, instead of using a hardcoded version, we can retrieve it from the Hermit environment variable.

.github/workflows/weekly-serverless.yml
gurevichdmitry: In order to retrieve `ELK_VERSION`, we need to initialize Hermit.

deploy/test-environments/cis/main.tf
gurevichdmitry: The default value is [25](https://registry.terraform.io/providers/hashicorp/aws/4.15.0/docs#max_retries-3), but we still encountered an API error when creating the EKS cluster. So, I explicitly increased this value to 30.

orouz (approved):

PR #3087, opened 2025-03-07
Performance improvements (quick wins)

moukoublen:
### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


internal/flavors/publisher.go
```diff
@@ -98,5 +98,14 @@ func (p *Publisher) publish(events *[]beat.Event) {
 	p.log.With(ecsEventActionField, ecsEventActionValue, ecsEventCountField, len(*events)).
 		Infof("Publishing %d events to elasticsearch", len(*events))
 	p.client.PublishAll(*events)
-	*events = nil
+	*events = (*events)[:0] // reuse the capacity and set len to 0.
+
+	// if for some reason capacity exceeds 4*threshold, drop the slice and create a new one. (it will never get here, just a precaution)
```
romulets, line 103: Therefore you are expecting the capacity to grow?

The flow is the following:

- Init with capacity of 1.5x the threshold
- Append to event once the channel receives it
- If the capacity is 4x higher than threshold, restart the capacity

The question is, how would capacity grow if `publish` is a blocking action once the threshold has been met?

moukoublen: Our flow is pretty much (in high level) here: 

https://github.com/elastic/cloudbeat/blob/5eea2c30d3aa8ba9e1dc6abe708d517bffd23c38/internal/flavors/benchmark/builder/benchmark.go#L52-L56

and here 

https://github.com/elastic/cloudbeat/blob/5eea2c30d3aa8ba9e1dc6abe708d517bffd23c38/internal/flavors/posture.go#L96-L104

___


resources -> findings (plural) per resource -> create (plural) beat events -> publish.


SO the channel in this case receives multiple events ` []beat.Event`

```golang
case event, ok := <-ch:
// ...
eventsToSend = append(eventsToSend, event...)
```

So in theory if a resource produces multiple findings (more than the threshold) the capacity will grow. And because we re-use we will keep using the "extended" capacity for ever (not nececerilly a bad thing) but its good to have a fallback. 


So if we have a single resource that produced an enormous number of findings (if), lets reset the slice in that specific case.

orestisfl (changes requested):
Quite important to add tests that cover cases with events exceeding 4x the threshold

internal/flavors/publisher.go
moukoublen: I think its not batched per resource type but per single resource. 

So the `case event, ok := <-ch:` basically means (if I am not mistaken) rule findings for a single resource id.

orestisfl:
Would be nice (outside of this PR) to set some kind of continued benchmarking tool to compare our performance / memory usage for go benchmarks: https://github.com/benchmark-action/github-action-benchmark

internal/flavors/publisher_test.go
```diff
@@ -208,3 +210,140 @@ func sleepOnCycleEnd(t *testing.T, size int, eventCount int, interval time.Durat
...
+	}
+}
+
+func events(ids []int) []beat.Event {
+	s := make([]beat.Event, len(ids))
+	for i := range len(ids) {
+		s[i] = event(ids[i])
+	}
+	return s
+}
```
orestisfl, line 349:
```suggestion
func generateEventBatches(batchesOf ...int) [][]beat.Event {
	batches := make([][]beat.Event, 0, len(batchesOf))

	id := 1
	for _, singleBatchCount := range batchesOf {
		batches = append(batches, eventsFromIds(lo.RangeFrom(id, singleBatchCount)))
		id += singleBatchCount
	}

	return batches
}

func eventsFromIds(ids []int) []beat.Event {
	s := make([]beat.Event, len(ids))
	for i := range len(ids) {
		s[i] = beat.Event{
			Fields: mapstr.M{"id": ids[i]},
		}
	}
	return s
}
```

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -398,7 +398,7 @@ func getAssetsByProject[T any](assets []*ExtendedGcpAsset, log *clog.Logger, f T
 
 func (p *Provider) getAllAssets(ctx context.Context, request *assetpb.ListAssetsRequest) []*assetpb.Asset {
 	p.log.Infof("Listing asset types: %v of type %v for %v", request.AssetTypes, request.ContentType, request.Parent)
-	results := make([]*assetpb.Asset, 0)
+	results := make([]*assetpb.Asset, 0) // TODO: use PageInfo().Remaining() to initialize with capacity <issue link>
```
orestisfl, line 401: Let's remove this TODO if we don't have an issue for it and if it's not a compatible change with streaming each page anyway

internal/flavors/publisher_test.go
```diff
@@ -208,3 +210,140 @@ func sleepOnCycleEnd(t *testing.T, size int, eventCount int, interval time.Durat
 		time.Sleep(waitPeriod)
 	}
 }
+
+func TestPublisher_HandleEvents_Buffer(t *testing.T) {
+	tests := map[string]struct {
+		interval                            time.Duration
```
orestisfl, line 216:
```suggestion
```

internal/flavors/publisher_test.go
```diff
@@ -208,3 +210,140 @@ func sleepOnCycleEnd(t *testing.T, size int, eventCount int, interval time.Durat
...
+			for i, publishedSliceIDs := range tc.expectedPublishedBatchesIDs {
+				calls[i] = client.EXPECT().PublishAll(events(publishedSliceIDs)).Run(
+					func(_ []beat.Event) {
+						assert.Equalf(t, tc.expectedBufferCapacityOnEachPublish[i], cap(publisher.eventsBuffer), "buffer capacity miss-match on publish number %d", i)
+					},
+				).Call
+			}
+			mock.InOrder(calls...)
+
+			publisher = NewPublisher(log, tc.interval, tc.threshold, client)
```
orestisfl, line 302:
```suggestion
			publisher = NewPublisher(log, time.Hour, tc.threshold, client)
```

orestisfl (commented):
The interval doesn't play an role in this test so let's not mislead by putting it in the table

internal/flavors/publisher_test.go
```diff
@@ -208,3 +210,140 @@ func sleepOnCycleEnd(t *testing.T, size int, eventCount int, interval time.Durat
...
+			expectedPublishedBatchesIDs:         [][]int{{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24}},
+			expectedBufferCapacityOnEachPublish: []int{27},
+			expectedBufferCapacityEnd:           7,
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			log := testhelper.NewLogger(t)
+			defer goleak.VerifyNone(t, goleak.IgnoreCurrent())
```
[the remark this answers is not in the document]
orestisfl: hmm, we have the one that sends the events but it's only for testing, not any actual code leaks. Up to you.

internal/flavors/publisher_test.go
```diff
@@ -208,3 +210,140 @@ func sleepOnCycleEnd(t *testing.T, size int, eventCount int, interval time.Durat
...
+
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			var publisher *Publisher
```
orestisfl, line 289:
```suggestion
			client := newMockClient(t)
			publisher := NewPublisher(log, time.Hour, tc.threshold, client)
```

IMO a bit cleaner

internal/flavors/publisher_test.go
```diff
@@ -208,3 +210,140 @@ func sleepOnCycleEnd(t *testing.T, size int, eventCount int, interval time.Durat
...
+			expectedBufferCapacityOnEachPublish: []int{7, 7},
+			expectedBufferCapacityEnd:           7,
+		},
+		"single batch over threshold": {
+			interval:                            2 * time.Second,
+			threshold:                           3,
+			incomeEventBatchesOf:                []int{5},
+			expectedPublishedBatchesIDs:         [][]int{{1, 2, 3, 4, 5}},
+			expectedBufferCapacityOnEachPublish: []int{8},
+			expectedBufferCapacityEnd:           8, // 4 * 2
```
orestisfl, line 261:
```suggestion
			expectedBufferCapacityEnd:           4*2,
```

internal/flavors/publisher_test.go
```diff
@@ -208,3 +210,140 @@ func sleepOnCycleEnd(t *testing.T, size int, eventCount int, interval time.Durat
 		time.Sleep(waitPeriod)
 	}
 }
+
+func TestPublisher_HandleEvents_Buffer(t *testing.T) {
+	tests := map[string]struct {
+		interval                            time.Duration
+		threshold                           int
+		incomeEventBatchesOf                []int
```
orestisfl, line 218:
```suggestion
		batchSizes                []int
```

internal/flavors/publisher.go
```diff
@@ -41,6 +41,8 @@ type Publisher struct {
 	interval  time.Duration
 	threshold int
 	client    client
+
+	eventsBuffer []beat.Event
```
orestisfl, line 45: As a note, this changes `Publisher` from being thread-safe to being thread-unsafe

internal/flavors/publisher_test.go
moukoublen: I believe it's a matter of pov, I prefer having the mock created and initialized together.

internal/flavors/publisher.go
moukoublen: I think we use the `HandleEvents` from a single go routine both on posture and vuln, I did n't find any case where we use it as a worker and spawn multiple go routines using the same struct instance. 

Also there is no other way to test the events buffer capacity growth and reset, unless you either provide the buffer externally (which might be a bit ugly API wise) or put the buffer in the struct's state.  


So to me that's fine since there is no need to do otherwise. But we can explicitly note it with a comment perhaps.

internal/flavors/publisher_test.go
moukoublen: I think you suggest that because the channel is closed in the test sender? 

I was planning (and forgot) to make it part of the test case `closeChannel` as the current one. It gives better flexibility to the testcase rather than, essentially, disabling the interval at all. (I will add the `closeChannel` boolean to testcase struct).

internal/flavors/publisher_test.go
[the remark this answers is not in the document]
moukoublen: The concept initially was to ensure that the test sender (go routine) is stopped consciously to avoid scenarios like hanging over channel send.

internal/flavors/publisher.go
orestisfl: Yes, sorry should have been clearer: I don't see any issue, just wanted to explicitly state in my review.

internal/flavors/publisher_test.go
orestisfl: hmm, my preference is because the client mock is dereferencing a nil pointer. Ofc it's in a callback and it is not dangerous in any way currently, but still feels "wrong".

Up to you.

internal/flavors/publisher_test.go
orestisfl: I am suggesting it in combination with the other suggestion in this review and the top-level comment:
> The interval doesn't play an role in this test so let's not mislead by putting it in the table

internal/flavors/publisher_test.go
moukoublen: resolving because I refactored and made them inline

PR #3088, opened 2025-03-10
Bump the version used for test environments.

gurevichdmitry:
### Summary of your changes

This PR updates the version used in test environments


bin/hermit.hcl
```diff
@@ -1,4 +1,4 @@
 env = {
-  "CLOUDBEAT_VERSION": "9.0.0",
+  "CLOUDBEAT_VERSION": "9.1.0",
```
orouz, line 2: i thought we always need this to be one version behind? is this not the case anymore?

gurevichdmitry: No, this one is used for testing. The problem is that when we bump the version, we still cannot execute tests because we need to wait for the snapshot to be created. Once the snapshot is ready, we can bump the version and use it for testing.

orouz (approved):

Issue #1181, opened 2023-08-02
[CNVM] Clean snapshots in a graceful shutdown


amirbenun:
### Motivation
Today cloudbeat fails to clean snapshots when context is done, instead we should create a termination context and have the cleanup anyway as part of a graceful shutdown.

### Definition of done
- [ ] All created snapshots should be deleted even on a signal

oren-zohar:
related: 
- https://github.com/elastic/cloudbeat/issues/950

amirbenun:
Another approach here is to start every CNVM cycle be deleting all the snapshots that were created by us more than 24 hours ago.
That way we can promise that as long as our CNVM solution runs on the AWS account we will keep deleting snapshots that should have been deleted.

orestisfl:
> Another approach here is to start every CNVM cycle be deleting all the snapshots that were created by us more than 24 hours ago. That way we can promise that as long as our CNVM solution runs on the AWS account we will keep deleting snapshots that should have been deleted.

Agreed, but we need a different ticket for that @oren-zohar

PR #3090, opened 2025-03-11
cnvm: Delete snapshots after scanning them

orestisfl:
### Summary of your changes
Fixes various underlying issues with CNVM snapshot deletion. The logic here is to do a best-effort attempt to clean up snapshots created during the run both continuously (after we are done scanning the snapshot) and on shutdown. Cleaning old snapshots that we don't use anymore is part of https://github.com/elastic/cloudbeat/issues/3105. Issues fixed:
- `internal/flavors/vulnerability.go`: Wait for `Run()` to finish, this ensures that final snapshot clean-up is done after execution finishes
- `internal/resources/providers/awslib/ec2/provider.go`: Give extra retries to snapshot deletion, mainly avoiding "too many requests" errors
- `internal/vulnerability/snapshot.go`: New snapshot manager to handle creation, deletion and clean-up of snapshots. The deletion extends the `context.Context` with an extra 30s timeout to give a grace period to clean-up snapshots during shutdown/restart.
- `internal/vulnerability/replicator.go`: Add dependency to the snapshot manager instead of `provider` to track created snapshots
- `internal/vulnerability/scanner.go`: Delete snapshot after scanning
- `internal/vulnerability/worker.go`: `defer` a call snapshot manager's cleanup

### Screenshot/Data
1. The way I verified we avoid leftover snapshots is to change the name of the snapshots:
  ```diff
  diff --git a/internal/resources/providers/awslib/ec2/provider.go b/internal/resources/providers/awslib/ec2/provider.go
  index 14abc5bf..3faeef7d 100644
  --- a/internal/resources/providers/awslib/ec2/provider.go
  +++ b/internal/resources/providers/awslib/ec2/provider.go
  @@ -78,7 +78,7 @@ func (p *Provider) CreateSnapshots(ctx context.Context, ins *Ec2Instance) ([]EBS
 			  {
 				  ResourceType: "snapshot",
 				  Tags: []types.Tag{
  -					{Key: aws.String("Name"), Value: aws.String(fmt.Sprintf("elastic-vulnerability-%s", *ins.InstanceId))},
  +					{Key: aws.String("Name"), Value: aws.String(fmt.Sprintf("orestis-elastic-vulnerability-%s", *ins.InstanceId))},
 					  {Key: aws.String("Workload"), Value: aws.String("Cloudbeat Vulnerability Snapshot")},
 				  },
 			  },
  ```
2. Deploy cloudbeat based on this PR + the diff above.
3. Do various tests like:
    - kill cloudbeat while its running
    - restart the agent
    - run continuously over the weekend (note: CNVM takes quite a while to finish a single cycle)
4. Verify no leftovers are with command:
  ```shell
  aws ec2 describe-snapshots --owner-ids self --query 'Snapshots[?StartTime>=`2025-03-10` && Tags[?Key==`Name` && starts_with(Value, `orestis-elastic-vulnerability-`)]] | [].SnapshotId' --output table
  ```

### Related Issues
See `Development` to the right

### Checklist


internal/vulnerability/scanner.go
```diff
@@ -201,6 +207,13 @@ func (f VulnerabilityScanner) scan(ctx context.Context, snap ec2.EBSSnapshot) {
 		}
 	}
 
+	go func() {
```
moukoublen, line 210: I was thinking that we might want to make that `defer` and not go routine for two reasons:

  1. Previously the `f.cleaner.CleanSnapshots(ctx)` was a blocking call to ensure that every "delete snapshot" was done before `VulnerabilityWorker.Run` ended. 
  2. [concern, not an issue currently] Potentially we have <number of snapshots>*routines. Although the snapshots are handled (I believe) serially and we might not end up in an un-capped go routine spawning situation. Still, perhaps it leaves the door open to a situation like this if we modify the snapshot scanning at some point in the future.

orestisfl: 1. Very good point, moved the goroutine
2. Even though I am not concerned with the resource impact of goroutines (I doubt this will be the bottleneck if we scale to millions of CNVM snapshots), it might be a good idea in case there are rate limits in AWS' side. That said, I feel the goroutine gives a statistically better chance to successfully complete the cleanup as of now.

moukoublen:
While I personally agree with the "delete after scan" approach, I would like to verify the reason for the initial design (separate cleaner worker over channel). 

For example, was there any concern regarding snapshot deletion duration? If yes, then perhaps the initial approach was to have the results available "as fast as possible"? 

I don't think it's possible, but it might be good to verify.

orestisfl:
> For example, was there any concern regarding snapshot deletion duration? If yes, then perhaps the initial approach was to have the results available "as fast as possible"?

Hmm, in this approach, the results are not delayed by the snapshot clean-up

moukoublen:
Technically they are (in the whole picture), because each now scan starts after the previous one has finished, both the scan and the delete, while previously it was only scan and the clean/delete was happening independently after the scan and the other tasks.  

(I don't believe it's an issue since CNVM could take hours; I just wanted to verify that there wasn't any specific reason that we might have missed it.)

orestisfl:
> Technically they are (in the whole picture), because each scan now starts after the previous one has finished, both the scan and the delete, while previously it was only scan and the clean/delete was happening independently after the scan and the other tasks.

This part didn't change unless if I missed something

orestisfl:
@moukoublen @kubasobon  added a context grace period to deal with #1181

moukoublen (approved):

PR #3095, opened 2025-03-12
Add support for running ECH and Serverless test workflows on 8.x versions

gurevichdmitry:
### Summary of your changes

This PR introduces the execution of ECH and Serverless test workflows for the main and 8.x branches.
Currently, the jobs run sequentially, but in the future, dependencies can be removed to enable parallel execution.

Changes in this PR:
- Introduced two new workflows:
    - **test-runner.yml** – Manages execution order and test run permutations.
    - **test-e2e-flow.yml** – Contains three jobs: `setup`, `deploy` (which includes test execution), and `destroy`.
- Additionally, since the new workflow serves as a common solution for nightly runs, the following workflows will be removed:
    - [weekly-serverless.yml](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/weekly-serverless.yml)
    - [test-ech-agentless.yml](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/test-ech-agentless.yml)

### Screenshot/Data

This is an [example](https://github.com/elastic/cloudbeat/actions/runs/14104021329) of a workflow run

Screenshot 2025-03-27 at 13 38 09

This is an [example](https://github.com/elastic/cloudbeat/actions/runs/14104193233) of a parallel workflow run

Screenshot 2025-03-27 at 13 40 27


romulets:
What do you mean with `dependencies can be removed to enable parallel execution`?

.github/workflows/test-e2e-flow.yml
```diff
@@ -0,0 +1,76 @@
...
+      branch:
+        required: true
+        type: string
+      deployment-type:
+        required: true
+        type: string
+      region:
+        required: true
+        type: string
+      serverless:
```
romulets, line 16: Doesn't `type` already covers if it's serverless or not?

romulets (approved):

gurevichdmitry:
> What do you mean with `dependencies can be removed to enable parallel execution`?

Currently, in the test-runner, jobs are defined in a sequence: `run-ech-8x` → `needs: run-ech-main`, and `run-serverless-main` → `needs: run-ech-8x`. Removing these dependencies will allow all jobs to execute in parallel.

.github/workflows/test-e2e-flow.yml
gurevichdmitry: @romulets, you're right, we can determine that from the deployment-type. I'll refactor and remove the serverless parameter.

gurevichdmitry: I've decided to remove `deployment-type` and keep `serverless`, as removing `deployment-type` requires significantly fewer changes.

romulets (approved):

PR #3138, opened 2025-03-24
Update flaky-test.yml descriptions

orestisfl:


.github/ISSUE_TEMPLATE/flaky-test.yml
```diff
@@ -15,15 +15,15 @@ body:
     id: link
     attributes:
       label: "Link"
-      description: "Link to file/line number in GitHub."
+      description: "Link to failing Github Action"
```
romulets, line 18: Github action logs aren't kept forever, therefore linking only the gh action isn't useful for future reference. Better have both, action and file/number

Issue #3105, opened 2025-03-17
CNVM: Clean-up old snapshots on startup


orestisfl:
**Motivation**
This came up in https://github.com/elastic/sdh-security-team/issues/1168 and in our own AWS accounts. Old snapshots that for some reason have failed to be deleted by cloudbeat during the normal cycle, persist in accounts without a purpose. We should delete older snapshots created by cloudbeat in a background job.

**Definition of done**
What needs to be completed at the end of this task
- [ ] Background cleaner is implemented
- [ ] Verify results in AWS account

**Out of scope**
Graceful shutdown https://github.com/elastic/cloudbeat/issues/1181

**Related tasks/epics**
- https://github.com/elastic/sdh-security-team/is[sues/1168
- https://github.com/elastic/cloudbeat/issues/2221
- https://github.com/elastic/cloudbeat/issues/1181
- https://github.com/elastic/cloudbeat/issues/950
- https://github.com/elastic/cloudbeat/pull/3090

oren-zohar:
is there a way to create snapshots with a defined TTL?

orestisfl:
> is there a way to create snapshots with a defined TTL?

None that I can find. Amazon Data Lifecycle Manager can automatically archive snapshots but only those created by that Manager: https://aws.amazon.com/blogs/storage/automatically-archive-amazon-ebs-snapshots-with-amazon-data-lifecycle-manager/

orestisfl:
Verified in eu-west-1 (oldest snapshot is 2 days old which is the limit).

<img width="3082" height="939" alt="Image" src="https://github.com/user-attachments/assets/eca62f47-fd0d-4f14-a868-48c0b26a2dbe" />

PR #3143, opened 2025-03-25
cnvm: Delete old snapshots on startup

orestisfl:
### Summary of your changes

Change CNVM to add a background job cleaning up old snapshots.

The logic here is:
1. Background routine is started. Since the amount of leftover snapshots can be quite high (e.g. 40k in our account),  3 cleanup workers are used
2. `IterOwnedSnapshots()` is called which goes over all snapshots, returning an iterator. Snapshots are selected if:
    1. They are more than 48 hours old
    2. They have a tag with key "Name" and value starting with "elastic-vulnerability"
    3. They are "self-owned"
3. Cleanup returns. On context cancelled, no extra grace period is added so the process doesn't block restarts/shutdown.

### Screenshot/Data

Deleted 1178 snapshots from `2025-03-25T16:17:08.384Z` to `2025-03-25T16:20:19.504Z`

### Related Issues

Closes https://github.com/elastic/cloudbeat/issues/3105
Closes https://github.com/elastic/sdh-security-team/issues/1168

### Checklist

#### Introducing a new rule?
No


internal/resources/providers/awslib/ec2/provider.go
```diff
@@ -311,6 +326,60 @@ func (p *Provider) DescribeSnapshots(ctx context.Context, snapshot EBSSnapshot)
...
+func filterSnap(snap types.Snapshot, before time.Time) bool {
+	if aws.ToTime(snap.StartTime).After(before) {
+		return false
+	}
+
+	for _, tag := range snap.Tags {
+		if aws.ToString(tag.Key) == "Name" {
+			return strings.HasPrefix(aws.ToString(tag.Value), snapshotPrefix)
+		}
+	}
```
romulets, line 379: Since we are already querying already the snapshots here:

```
                     Filters: []types.Filter{
					{
						Name:   aws.String("tag:Name"),
						Values: []string{fmt.Sprintf("%s-*", snapshotPrefix)},
					},
				},
				```
				
Do we need this piece of filtering?

File: internal/vulnerability/snapshot.go [hunk 2]
```diff
@@ -109,3 +144,33 @@ func runWithGrace(ctx context.Context, grace time.Duration, f func(ctx context.C
 	defer stop() // if the callback finishes in time, stop the AfterFunc
 	f(newCtx)    // finally, call the actual callback!
 }
+
+type contextualChan[T any] struct {
+	ch chan T
+}
```
moukoublen, line 150: [ just sharing thoughts - not suggesting to change it ]
I wonder which one of "wrapper struct vs just generic functions", is leaner in that case. 

eg instead of a wrapper just having something like that
```golang
func SendToChannel[T any](ctx context.Context, ch chan<- T, t T) bool
func ReadFromChannel[T any](ctx context.Context, ch <-chan T) (T, bool)
```

moukoublen (approved):

internal/vulnerability/snapshot.go
orestisfl: I wanted to completely avoid exposing the channel to prevent misusing it. Both can be valid approaches if direct access to the channel is needed.

internal/resources/providers/awslib/ec2/provider.go
orestisfl: We need to filter for time anyway so the name check is there just as a sanity check.

Issue #3081, opened 2025-03-05
CSPM GCP RAM footprint


moukoublen:
**Describe the bug**
We have records of high memory footprint (30 GB) on cloudbeat in cases where the GCP account under scan has a vast number of assets. 

Currently, GCP (and Azure) fetchers work by reading all pages of paginated responses, in memory, and then parsing and sending the assets to the channel for further processing. 

This could lead to the memory footprint being scaled linearly with the amount of scanned GCP assets. In cases that the account/project contains a vast amount of resources, the RAM footprint could peak accordingly.

A pprof run shows a big memory footprint also by the elastic event encoder, but it has not been verified whether it scales linearly with the amount of resources or it's capped.

**Preconditions**
GCP project/account with big number of resources.

**Additional context**
We should refactor the GCP fetchers to stream each page rather than wait to collect all pages when calling the gcp SDK to fetch the resources. 

We should check if there is any implication / any change in the rest steps of the mechanism to support that stream-based approach (e.g. does the rest of the mechanism expect all the resources grouped by type or it can support multiple batches per type). 

This can be done step by step (fetcher by fetcher) and test accordingly. 

To test the improvement we need a GCP account with a lot of resources **or** to mock the GCP part. 

Note: _The same pattern has been followed in Azure pagination. We can investigate Azure on a later step based on findings on GCP refactor_

orouz:
We've made some progress on this by reducing big documents size (#3216, #3249, #3240) and improving memory usage (#3152) and then provided a pre-release cloudbeat binary to the customer to test. 

the results were initially promising, [reported](https://elastic.slack.com/archives/C03U2T8B2S3/p1750325733146229) at 4x improvements for ES and 6x for cloudbeat in terms of memory. but after 2 hours, both the transform failed and the VM running cloudbeat crashed. 

the next step is to look at the diagnostics and get cloudbeat logs from the customer. 

for the transforms, we can see why they failed (chunk size too big): 

![Image](https://github.com/user-attachments/assets/b55cc6aa-7239-49f3-affc-b1315102db5e)
![Image](https://github.com/user-attachments/assets/b10ad260-1815-4538-a85b-e88d61ba9258)
![Image](https://github.com/user-attachments/assets/7c1dbf06-62cb-4a13-b991-7941dd9ddc7d)

[...]

Issue #2182, opened 2024-05-07
Use cycle-aware cache for GCP cloud metadata


orouz:
**Motivation**

we currently initiate a cache in the GCP asset inventory provider to hold cloud account metadata:
https://github.com/elastic/cloudbeat/blob/ded46cab11e3966a0f9c3f15d0edc2f96b3c7219/internal/resources/providers/gcplib/inventory/provider.go#L178

but if the project/org names changed between cycles (unlikely, but possible) we'd still get the old values. we still need the cache itself, as resources tend to have same project/org names, but we need to make sure the values are fresh. 

**Definition of done**
- [ ] make sure we get fresh values for cloud account metadata

**Related tasks/epics**
- raised in https://github.com/elastic/cloudbeat/pull/2055#discussion_r1568303037

PR #3152, opened 2025-03-30
make GCP provider work concurrently

orouz:
### Summary of your changes
WIP

### Screenshot/Data
WIP

### Related Issues
- Related: https://github.com/elastic/cloudbeat/issues/3081


internal/resources/fetching/fetchers/gcp/assets_fetcher.go
```diff
@@ -94,33 +79,31 @@ func NewGcpAssetsFetcher(_ context.Context, log *clog.Logger, ch chan fetching.R
...
-
-		for _, asset := range assets {
-			select {
-			case <-ctx.Done():
-				f.log.Infof("GcpAssetsFetcher.Fetch context err: %s", ctx.Err().Error())
+	f.log.Info("GcpAssetsFetcher.Fetch start")
+	defer f.log.Info("GcpAssetsFetcher.Fetch done")
+
+	resultsCh := make(chan *inventory.ExtendedGcpAsset)
+	go f.provider.ListAssetTypes(ctx, lo.Keys(reversedGcpAssetTypes), resultsCh)
```
orouz, line 86: instead of fetching each asset type separately, we now send a request for all asset types and get back an iterator for all of them

internal/resources/fetching/fetchers/gcp/networks_fetcher.go
orouz, line 1: this is a new fetcher for network assets, it was added to avoid an enrichment attempt we did on all assets instead of just network assets:

https://github.com/elastic/cloudbeat/blob/afdb27ccc6c636019c68407a8a3e4885e848d063/internal/resources/providers/gcplib/inventory/provider.go#L213

these assets are now fetched and enriched in their own fetcher, avoiding said attempt when fetching all other assets in `assets_fetcher`

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -91,242 +75,327 @@ type dnsPolicyFields struct {
...
-	ListAllAssetTypesByName(ctx context.Context, assets []string) ([]*ExtendedGcpAsset, error)
-
-	// ListMonitoringAssets List all monitoring assets by project id
-	ListMonitoringAssets(ctx context.Context, monitoringAssetTypes map[string][]string) ([]*MonitoringAsset, error)
-
-	// ListLoggingAssets returns a list of logging assets grouped by project id, extended with folder and org level log sinks
-	ListLoggingAssets(ctx context.Context) ([]*LoggingAsset, error)
-
-	// ListServiceUsageAssets returns a list of service usage assets grouped by project id
-	ListServiceUsageAssets(ctx context.Context) ([]*ServiceUsageAsset, error)
```
orouz, line 111: these two were combined in `ListProjectAssets`

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -91,242 +75,327 @@ type dnsPolicyFields struct {
...
-	p.enrichNetworkAssets(ctx, extendedAssets)
+	go p.getAllAssets(ctx, p.config.Parent, assetpb.ContentType_RESOURCE, assetTypes, resourceCh)
+	go p.getAllAssets(ctx, p.config.Parent, assetpb.ContentType_IAM_POLICY, assetTypes, policyCh)
+	go p.mergeAssets(ctx, mergeCh, resourceCh, policyCh)
+	go p.enrichAssets(ctx, mergeCh, enrichCh)
 
-	return extendedAssets, nil
+	for asset := range enrichCh {
+		out <- asset
+	}
```
orouz, line 146: this method is used by the main fetcher (`assets_fetcher`) and the way it works is that we start fetching resources and then merge them by name. the merging process sends assets to the enrichment channel if: 1) an asset has a resource and a policy, or 2) an asset has either a resource or a policy, and the other channel is closed, so we don't need to wait for merging with the other content type

after an asset is enriched it's sent to the out channel which is received in the fetcher and passed to the `resourceCh` to start the relevant flavor (cspm/assets inventory) pipeline

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -91,242 +75,327 @@ type dnsPolicyFields struct {
...
 }
 
-// ListMonitoringAssets returns a list of monitoring assets grouped by project id
-func (p *Provider) ListMonitoringAssets(ctx context.Context, monitoringAssetTypes map[string][]string) ([]*MonitoringAsset, error) {
-	logMetrics, err := p.ListAllAssetTypesByName(ctx, monitoringAssetTypes["LogMetric"])
-	if err != nil {
-		return nil, err
+func (p *Provider) ListMonitoringAssets(ctx context.Context, out chan<- *MonitoringAsset) {
```
orouz, line 149: this method used to fetch all resources and policies for 2 asset types, then group them by project, and once all assets are grouped, return a slice with all groups

it now does the same thing, but sends each group once it's ready, instead of waiting for all of them to be ready. also, only resources are fetched (no policies, as CIS rules don't require it. not all these asset types necessarily have a policy, but trying to fetch them was redundant)

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -91,242 +75,327 @@ type dnsPolicyFields struct {
...
+			CloudAccount: logAssets[0].CloudAccount,
+		}
 	}
+}
 
-	alertPolicies, err := p.ListAllAssetTypesByName(ctx, monitoringAssetTypes["AlertPolicy"])
-	if err != nil {
-		return nil, err
-	}
+func (p *Provider) ListProjectsAncestorsPolicies(ctx context.Context, out chan<- *ProjectPoliciesAsset) {
```
orouz, line 189: this used to send a slice of items that each represent a group of assets of the same project. it now does the same, but sends each group separately so we don't wait for all of them.

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -91,242 +75,327 @@ type dnsPolicyFields struct {
...
+			CloudAccount: asset.CloudAccount,
+			Policies:     append([]*ExtendedGcpAsset{asset}, p.getAssetAncestorsPolicies(ctx, asset.Ancestors[1:], policiesCache)...),
 		}
 	}
+}
 
-	var assets []*ExtendedGcpAsset
-	assets = append(append(assets, logMetrics...), alertPolicies...)
-	monitoringAssets := getAssetsByProject[MonitoringAsset](assets, p.log, typeGenerator)
+func (p *Provider) ListProjectAssets(ctx context.Context, assetTypes []string, out chan<- *ProjectAssets) {
```
orouz, line 207: this function now handles the same thing `ListLoggingAssets` and `ListServiceUsageAssets` used to. they both used to fetch both resource and policy for some asset types, group them all by project and send a slice with each group.

it now does the same thing, but starts by fetching projects and then for each project fetches only the resource of the given asset types and the sends it out. the difference is that we don't wait for all assets from all projects, plus we don't fetch policies as they weren't used by the relevant CIS rules.

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -91,460 +75,397 @@ type dnsPolicyFields struct {
...
 }
 
 type ProviderInitializerAPI interface {
-	// Init initializes the GCP asset client
 	Init(ctx context.Context, log *clog.Logger, gcpConfig auth.GcpFactoryConfig) (ServiceAPI, error)
 }
 
-func (p *ProviderInitializer) Init(ctx context.Context, log *clog.Logger, gcpConfig auth.GcpFactoryConfig) (ServiceAPI, error) {
+func newAssetsInventoryWrapper(ctx context.Context, log *clog.Logger, gcpConfig auth.GcpFactoryConfig) (*AssetsInventoryWrapper, error) {
 	limiter := NewAssetsInventoryRateLimiter(log)
```
orouz, line 97: we use rate limiting for the GCP Assets Inventory `ListAssets` method, and because we're making more concurrent calls, the cycle fetching takes longer (~30seconds vs ~18secs) as we wait between calls.

internal/resources/fetching/fetchers/gcp/assets_fetcher.go
```diff
@@ -94,33 +79,31 @@ func NewGcpAssetsFetcher(_ context.Context, log *clog.Logger, ch chan fetching.R
...
-			continue
-		}
-
-		for _, asset := range assets {
-			select {
-			case <-ctx.Done():
-				f.log.Infof("GcpAssetsFetcher.Fetch context err: %s", ctx.Err().Error())
+	f.log.Info("GcpAssetsFetcher.Fetch start")
+	defer f.log.Info("GcpAssetsFetcher.Fetch done")
+	defer f.provider.Clear()
```
orouz, line 84: `Clear()` will clear out the cache used for cloud account metadata, which is accessed by all assets fetched (mostly all hits)

it used to not get cleared at all (see https://github.com/elastic/cloudbeat/issues/2182)
now it gets cleared when each fetcher exits, which is better but also not good, as it should get called when all fetchers are done. i think a better place would be if the registry would be the one calling a `Clear` method on the `Fetcher` interface, allowing them to clear up caches when fetching cycle is done.

moukoublen (approved):

internal/resources/providers/gcplib/inventory/resource_manager.go
```diff
@@ -0,0 +1,142 @@
...
+		OrganizationName: orgName,
+	}
+}
+
+func (c *ResourceManagerWrapper) Clear() {
+	c.accountMetadataCache.Clear()
+}
+
+func getOrganizationId(ancestors []string) string {
+	last := ancestors[len(ancestors)-1]
```
moukoublen, line 124: nit: perhaps a check if the ancestors has zero length (is it possible?)

orouz: ancestors can't be empty
https://github.com/googleapis/google-cloud-go/blob/656d319252e5d973366a2c10f57f072b170dee1f/asset/apiv1p5beta1/assetpb/assets.pb.go#L95-L103

PR #3250, opened 2025-04-29
Increase gcp ListAssets timeout and page size

moukoublen:
### Summary of your changes
on gcp asset client:
  * Increased the default timeout on `ListAssets` to 4 minutes (configurable).
  * Increased the default page size to 200 (from 100) on `ListAssets` (configurable).
  * Added overwrite functionality using env vars `CLOUDBEAT_GCP_LIST_ASSETS_PAGE_SIZE` and `CLOUDBEAT_GCP_LIST_ASSETS_TIMEOUT`.
  * Created a wrapper on gax Retryer that logs on each retry. 

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -272,7 +272,8 @@ func overwritesFromEnvVars(c *Config) {
 	logErr := func(name string, value string, err error) {
 		log.Errorw(
 			"error trying to parse config env variable",
-			logp.String(name, value),
+			logp.String("name", name),
+			logp.String("value", value),
 			logp.Error(err),
 		)
 	}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -31,6 +31,7 @@ import (
 	"github.com/elastic/beats/v7/x-pack/libbeat/common/aws"
 	"github.com/elastic/elastic-agent-libs/config"
 
+	"github.com/elastic/cloudbeat/internal/infra/clog"
 	"github.com/elastic/cloudbeat/internal/launcher"
 )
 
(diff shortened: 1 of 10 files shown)


orestisfl (commented):
I know it's simpler but why have we opted to use envvars? Is this something that we want to configure specifically for agentless? If users are to be instructed to override this, wouldn't it be more user-friendly to do that via cloudbeat's config?

moukoublen:
@orestisfl user can edit the cloudbeat config since its part of the config struct. 

> Is this something that we want to configure specifically for agentless? 
Yes, the env vars are another way of setting those values in case we need to do it from our side for agentless.

internal/config/config.go
```diff
@@ -231,3 +259,22 @@ func newCloudConnectorsConfig() CloudConnectorsConfig {
 		ResourceID:    os.Getenv(CloudResourceIDEnvVar),
 	}
 }
+
+const (
+	CloudbeatGCPListAssetPageSizeEnvVar = "CLOUDBEAT_GCP_LIST_ASSETS_PAGE_SIZE"
+	CloudbeatGCPListAssetTimeoutEnvVar  = "CLOUDBEAT_GCP_LIST_ASSETS_TIMEOUT"
+)
+
+func overwritesFromEnvVars(c *Config) {
```
olegsu, line 268: I suggest to log errors that can occur while overwriting

moukoublen: Done

internal/config/config.go
```diff
@@ -231,3 +260,31 @@ func newCloudConnectorsConfig() CloudConnectorsConfig {
...
+	log := clog.NewLogger("config")
+	logErr := func(name string, value string, err error) {
+		log.Errorf("error trying to parse variable %s with value %s: %s", name, value, err.Error())
+	}
+
```
olegsu, line 274: I think it worth to keep the message predicted and the values as keys and values to make later search in o11y easier, wdyt?

internal/resources/providers/gcplib/inventory/grpc_rate_limiter.go
```diff
@@ -29,14 +29,43 @@ import (
...
+func (r retryer) Retry(err error) (time.Duration, bool) {
+	pause, shouldRetry := r.gaxRetryer.Retry(err)
+
+	if shouldRetry {
+		s := ""
+		// This should always be not nil, but let's be extra cautious.
+		if err != nil {
+			s = err.Error()
+		}
+		r.log.Debugf("gax Retryer retries based on error (%s) pause %s", s, pause.String())
```
olegsu, line 47: Similar approach might be useful here, i suggest to have something like
```golang
		r.log.Debug("gax retryer", "error", s, "pause", pause.String())
```

internal/config/config.go
moukoublen: In general, I agree, I am all about structured logging. I just tried to follow the cloudbeat's code way, where we don't use structured logging almost anywhere, but yes, we can change those two.

olegsu (approved):

Issue #3230, opened 2025-04-22
[CSPM AWS ORG] Initial operations happens without assuming the cloudbeat-root role


moukoublen:
## Describe the bug
On AWS ORG CSPM some initial operations are happening without assuming the `cloudbeat-root` role. 

#### `ListAccountAliases`
`AWSOrg::getIdentity` that calls `a.IdentityProvider.GetIdentity` [link](https://github.com/elastic/cloudbeat/blob/5eea2c30d3aa8ba9e1dc6abe708d517bffd23c38/internal/flavors/benchmark/aws_org.go#L234).

[This](https://github.com/elastic/cloudbeat/blob/5eea2c30d3aa8ba9e1dc6abe708d517bffd23c38/internal/flavors/benchmark/aws_org.go#L79) call happens using the initial aws config and thus not assumed the cloudbeat-root role. 

This call eventually invokes `ListAccountAliases` from the AWS SDK that requires `iam:ListAccountAliases` policy that is included in `cloudbeat-root` role. 

#### `GetRole` (critical)

`AWSOrg::IAMProvider` is [initialized](https://github.com/elastic/cloudbeat/blob/5eea2c30d3aa8ba9e1dc6abe708d517bffd23c38/internal/flavors/benchmark/aws_org.go#L91) using the initial config as well and thus not the assumed `cloudbeat-root` role. The `IAMProvider` is used to [query](https://github.com/elastic/cloudbeat/blob/5eea2c30d3aa8ba9e1dc6abe708d517bffd23c38/internal/flavors/benchmark/aws_org.go#L175) the cloudbeat root role. But for that to be invoked, the `iam:GetRole` policy is required. 

#### Impact on VM / Agent-Based 
On VM that has been created using our cloudformation template the mentioned operations succeed because the VM uses the [`IamInstanceProfile`](https://github.com/elastic/cloudbeat/blob/242df19b8c31034e4f4de07554dbdc46b61e5f4b/deploy/cloudformation/elastic-agent-ec2-cspm-organization.yml#L163) that is linked with the [`cloudbeat-root`](https://github.com/elastic/cloudbeat/blob/242df19b8c31034e4f4de07554dbdc46b61e5f4b/deploy/cloudformation/elastic-agent-ec2-cspm-organization.yml#L140) role. Having that in place, every SDK operation runs inside the VM has already assumed the cloudbeat-root role (and re-assumes it later). 

#### Impact on Direct Credentials CloudFormation
[...]

PR #3289, opened 2025-05-22
AWS CSPM ORG: initial operations fix. Use cloudbeat-root role

moukoublen:
### Summary of your changes

Changed the aws cspm org flow to first assume the `cloudbeat-root` role (if not already assumed) and then perform the `AWSOrg::getIdentity` (`ListAccountAliases`) and `AWSOrg::IAMProvider` (`GetRole`) that require the policies that `cloudbeat-role` has.

We have 4 different cases:

  * case A [**EC2 Instance**]: in this case we have already assumed `cloudbeat-root` role automatically because of `InstanceProfile`.
  * case B [**Direct Credentials**]: In this case, we have **not** assumed the `cloudbeat-root` role, but we have the same policies as the cloudbeat-root role, added to the user. _(need to assume `cloudbeat-root` role as first step)_
  * case C [Cloud Connectors]: in this case we have **not** assumed the `cloudbeat-root` role **nor** have the same policies with cloudbeat-root. _(need to assume `cloudbeat-root` role as first step)_
  * [Custom Setup] On custom setup like manual authentication for organization-level onboarding. _(need to assume `cloudbeat-root` role as first step)_

### Related Issues

Fixes: https://github.com/elastic/cloudbeat/issues/3230

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests


internal/flavors/benchmark/aws_org.go
```diff
@@ -217,26 +218,63 @@ func (a *AWSOrg) pickManagementAccountRole(ctx context.Context, log *clog.Logger
...
+		// [Cloud Connectors] On cloud connectors this ends up assuming the customer remote role (using role chaining)
+		awsConfig, err = awslib.InitializeAWSConfigCloudConnectors(ctx, cfg.CloudConfig.Aws)
+	} else {
+		// [EC2 Instance] On EC2 created with our cloud formation, the identity is inferred by the EC2 instance InstanceProfile which has the cloudbeat-root role.
+		// [Direct Credentials] On Direct credentials this identity is the user created by the cloud formation.
+		// [Custom Setup] On custom setup like manual authentication for organization-level onboarding.
+		awsConfig, err = awslib.InitializeAWSConfig(cfg.CloudConfig.Aws.Cred)
+	}
+
+	return awsConfig, err
```
kubasobon, line 277:
```suggestion
func (a *AWSOrg) getInitialAWSConfig(ctx context.Context, cfg *config.Config) (*awssdk.Config, error) {
	if cfg.CloudConfig.Aws.CloudConnectors {
		// [Cloud Connectors] On cloud connectors this ends up assuming the customer remote role (using role chaining)
		return awslib.InitializeAWSConfigCloudConnectors(ctx, cfg.CloudConfig.Aws)
	}
	
    // [EC2 Instance] On EC2 created with our cloud formation, the identity is inferred by the EC2 instance InstanceProfile which has the cloudbeat-root role.
    // [Direct Credentials] On Direct credentials this identity is the user created by the cloud formation.
    // [Custom Setup] On custom setup like manual authentication for organization-level onboarding.
	return awslib.InitializeAWSConfig(cfg.CloudConfig.Aws.Cred)
```

kubasobon (approved):
Looks good, just one comment. 
In an ideal world it would be nice to test this on AWS Gov Cloud as well.

PR #3412, opened 2025-06-26
Bump cloudbeat version

cloudsecmachine:
Bump cloudbeat version - `9.2.0`


moukoublen:
ARM Lint workflow no longer works because `microsoft/action-armttk` uses deprecated version of `actions/upload-artifact@v3`.

Will merge once the other workflows are successful.

PR #3429, opened 2025-07-02
remove evidence field from CIS GCP

orouz:
### Summary of your changes

this PR removes the `result.evidence` field from all CIS GCP rules by omitting it from the rego rules. the motivation for this is to reduce raw findings size. regardless, as seen in the changed files, majority of the rules already send the entire resource as the evidence, and since that resource is already available as `resource.raw` we can just point to it in kibana instead of duplicating it. 

findings count and size were checked with:
- `POST logs-cloud_security_posture.findings-default/_count`
- `GET _cat/indices/logs-cloud_security_posture.findings-default?v&h=index,store.size`

| 	| before (`main`)	|  after 	|
|-------	|--------	|---	|
| count     	| `1322`      	|  `1322` 	|
|  size     	|    `1126.4kb`	|  `914.3kb` 	|

this makes the default findings index about `18.83%` smaller

### Related Issues

- closes https://github.com/elastic/security-team/issues/12315
- related https://github.com/elastic/kibana/pull/227263


security-policies/bundle/compliance/cis_gcp/rules/cis_1_17/rule.rego
```diff
@@ -9,10 +9,7 @@ default has_cusomter_encrypted_key := false
 finding := result if {
 	data_adapter.is_dataproc_cluster
 
-	result := common.generate_result_without_expected(
-		common.calculate_result(has_cusomter_encrypted_key),
-		data_adapter.resource,
-	)
+	result := common.generate_evaluation_result(common.calculate_result(has_cusomter_encrypted_key))
 }
```
orouz, line 13: this is the common theme in this PR - replace `generate_result_without_expected` with `generate_evaluation_result` which results in sending an object with just `evaluation` field instead of a one including `evidence` too

uri-weisman (approved):
LGTM 🚀 

1. Before merging please verify that this PR didn't change the number of failed/passed findings for the same dataset.
2. Merge is blocked on the corresponding Kibana PR.

orouz:
> @orouz should it be backported to `9.1.0`?

generally speaking, yeah. but that's in June 24, which isn't so far from now, and i'm waiting to see the customer feedback before merging this to main or any backports.

olegsu:
Should also to backport to 9.x?

uri-weisman:
@olegsu 

> Should also to backport to 9.x?

We agreed to backport to the next patch versions (`8.19.1` and `9.1.1`) as we want more time to be assure those changes didn't introduce issues.

PR #3434, opened 2025-07-07
Improve assets resources and policies fetching

orouz:
### Summary of your changes

Replace the per-asset-type parallel fetch (resources + policies) with a two-pass sequential strategy that minimizes the in-memory buffer:
1. old: `assetStore` (max - `number of asset types * number of policies * number of resources * number of projects`)
2. new: `policyByName` (max - `number of policies * number of projects`)

### Related Issues
- closes https://github.com/elastic/cloudbeat/issues/3081


internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -113,36 +113,37 @@ func newAssetsInventoryWrapper(ctx context.Context, log *clog.Logger, gcpConfig
...
-	for asset := range assetsCh {
-		select {
-		case <-ctx.Done():
+	for _, t := range assetTypes {
+		if ctx.Err() != nil {
 			return
-		case out <- asset:
+		}
+
+		resCh := p.getAssets(ctx, p.config.Parent, []string{t}, assetpb.ContentType_RESOURCE)
```
uri-weisman, line 142: nit: It makes more sense to send only the type and append it to a list inside `getAssets` logic.
```suggestion
		resCh := p.getAssets(ctx, p.config.Parent, t, assetpb.ContentType_RESOURCE)
```

internal/resources/providers/gcplib/inventory/provider.go
```diff
@@ -297,41 +298,101 @@ func (p *Provider) getAssetAncestorsPolicies(ctx context.Context, ancestors []st
...
+
+		for polCh != nil || resCh != nil {
+			select {
+			case <-ctx.Done():
+				return
+
+			case a, ok := <-polCh:
+				if !ok {
+					polCh = nil
+					flushAssets(out, store, resCh, polCh)
```
uri-weisman, line 329: Isn't a single `flushAssets` call when both channels are closed is sufficient? we already check in `mergeAsset` if one of the channels is already closed in order to expedite processing.

uri-weisman (commented):
Good Job 💪

internal/resources/providers/gcplib/inventory/provider.go
orouz: it would, but `getAssets` is also used elsewhere that requires passing an array.

internal/resources/providers/gcplib/inventory/provider.go
orouz: for correctness - yeah, we could do with a single `flushAssets`. but for performance - we shouldn't, as these 2 additional calls mean we drain the store faster which directly impacts memory usage.

without it, after one channel closes (let's assume policies channel closed), `mergeAsset` will only send **the next** asset that comes because 1) it found its policy pair, or 2) it didn't but the policies channel is closed so we don't need to wait. but what about all the resources already in the map? by now we know they don't have a policy, and if we don't flush, we'll have to wait until the resource channel is closed too.

internal/resources/providers/gcplib/inventory/provider.go
uri-weisman: Didn't find a place that sends more than one type, did I miss it?

uri-weisman (approved):
LGTM

orouz:
the customer tried this and we went from 12gb (previous pre-release) to 9gb peak RAM in this iteration. kinda disapponted tbh, was expecting much more than 25% improvement given the test i ran with the actual asset counts they have and a big asset size (derived from their biggest document). the test did run in isolation though, testing only the fetching phase, so it's possible the rest of the pipeline has also room for improvements too.

in anyway, this is a good change and we should merge it.

moukoublen (approved):

PR #3452, opened 2025-07-14
Adding JSON tag in JSON in a text

animehart:
### Summary of your changes

This PR addresses the issue where JSON text is not being tagged properly

### DIFF
--- a/security-policies/dev/common.py
+++ b/security-policies/dev/common.py
@@ -176,26 +176,35 @@ def add_new_line_after_period(text):
     # Join the lines back into a single string and return the result
     return "\n".join(lines)
 
-
 def format_json_in_text(text):
-    try:
-        # Search for JSON-like content in the text
-        start_index = text.find("{")
-        end_index = text.rfind("}") + 1
-        json_str = text[start_index:end_index]
+    def fix_and_format_json(json_candidate):
+        try:
+            # Attempt to load directly
+            parsed = json.loads(json_candidate)
+        except json.JSONDecodeError:
+            try:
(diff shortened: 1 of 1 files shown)


uri-weisman:
@animehart thanks for contributing this code, can you please run the script and we'll see the change to the existing metadata files?

security-policies/dev/common.py
```diff
@@ -197,9 +321,16 @@ def format_json_in_text(text):
...
-    text = format_json_in_text(text)
+    if rule_number in {"1.17", "3.5"} and benchmark_id == "cis_aws":
+        text = format_json_in_text(text)
+    elif (rule_number in {"3.10", "3.11"} and benchmark_id == "cis_aws") or (
+        rule_number in {"5.1.5"} and benchmark_id == "cis_azure"
+    ):
+        text = format_json_in_string_command(text)
+    else:
+        text = format_and_fix_json_in_text(text)
     return check_and_fix_numbered_list(text)
```
uri-weisman, line 334: I’d prefer that we don’t apply a function tailored to a specific rule. Instead, we should implement a fallback mechanism, if the output doesn’t match the expected format or an error was thrown - we trigger the next JSON formatting function in line.

security-policies/dev/common.py
```diff
@@ -161,6 +161,76 @@ def check_and_fix_numbered_list(text):
     return "\n".join(lines)
 
 
+def fix_broken_json_block(text: str) -> str:
```
uri-weisman, line 164: Consider to divide to small functions to increase readability.

PR #3519, opened 2025-08-11
JSON Indentation Fix -Adding JSON tag and/or Indentations-

animehart:
### Summary of your changes

this PR only handles adding JSON tag and not fixing any broken JSON.

It won't add a JSON tag in the code block IF:
1. The code inside the code block is not a valid JSON
2. The code inside the code contains valid JSON but is within a string (like for example a command)

In the case its not a Valid JSON but still "looks like" valid JSON (as in encompassed within { } properly without missing brackets) then it will just add proper indentation to it


security-policies/bundle/compliance/cis_aws/rules/cis_3_6/data.yaml
```diff
@@ -39,10 +39,10 @@ metadata:
 
     ```
     {
-        "LoggingEnabled": {
-            "TargetPrefix": "<Prefix_Test>",
-            "TargetBucket": "<Bucket_name_for_Storing_Logs>"
-        }
+     "LoggingEnabled": {
```
niros1, line 42: seems like a regression

security-policies/bundle/compliance/cis_aws/rules/cis_3_6/data.yaml
```diff
@@ -39,10 +39,10 @@ metadata:
 
     ```
```
niros1, line 40: missing json

security-policies/dev/common.py
```diff
@@ -177,24 +178,37 @@ def add_new_line_after_period(text):
...
+        return f"```json\n{pretty}\n```"
+    except json.JSONDecodeError:
+        # Fallback: Try to find JSON-like substring inside code block
+        try:
+            start = code_block.find("{")
+            end = code_block.rfind("}") + 1
+            if start == -1 or end == -1:
+                raise ValueError("No JSON-like structure found")
+
+            json_str = code_block[start:end]
```
niros1, line 196: this one might take part of the JSON, why aren't use a simple function that trying to add missinf brackets by counting open and closed brackets?

security-policies/dev/common.py
```diff
@@ -177,24 +178,37 @@ def add_new_line_after_period(text):
...
+        parsed = json.loads(code_block)
+        pretty = json.dumps(parsed, indent=4)
+        return f"```json\n{pretty}\n```"
+    except json.JSONDecodeError:
+        # Fallback: Try to find JSON-like substring inside code block
+        try:
+            start = code_block.find("{")
+            end = code_block.rfind("}") + 1
+            if start == -1 or end == -1:
+                raise ValueError("No JSON-like structure found")
```
niros1, line 194: not sure the is a point for error here, you can just return code_block, if u need a log just add a log.

niros1 (commented):
Can you give a more meaningful name then fix 1?

security-policies/dev/common.py
animehart: I thought we just want to add JSON tag for this PR ?

security-policies/bundle/compliance/cis_aws/rules/cis_3_6/data.yaml
animehart: I need help figuring why this one didn't get detected as JSON 😵

security-policies/dev/common.py
```diff
@@ -177,24 +178,37 @@ def add_new_line_after_period(text):
...
-        return text
+        # Try parsing the entire code block as JSON
+        parsed = json.loads(code_block)
+        pretty = json.dumps(parsed, indent=4)
+        return f"```json\n{pretty}\n```"
+    except json.JSONDecodeError:
+        # Fallback: Try to find JSON-like substring inside code block
+        try:
+            start = code_block.find("{")
+            end = code_block.rfind("}") + 1
```
niros1: this one will never be -1

uri-weisman (commented):
Nice! minor comments

security-policies/bundle/compliance/cis_aws/rules/cis_2_1_1/data.yaml
```diff
@@ -48,11 +48,28 @@ metadata:
 
     Run either 
     ```
```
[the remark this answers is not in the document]
animehart: This is intentional
This is not JSON by itself, its part of a command string as such I didn't add the JSON tag here as I only want the tag for only pure JSON string

PR #3531, opened 2025-08-18
[aws single account]: use aws account name if alias is not configured

moukoublen:
### Summary of your changes
In AWS (both for CSPM and Asset Inventory) we have 2 ways to initialize the `cloud.Identity` (that contains the account number and account name). 

On the single-account we use `IdentityProvider.GetIdentity` that used to call aws `iam.ListAccountAliases` to fetch the configured **aliases** for the current aws account.

On the organization setup, we use `AccountProvider.ListAccounts` that calls aws `organizations.ListAccounts` to fetch all the org accounts with the **name** (not alias) of each one. 

In case we have single-account and **no alias configured** in the aws account, we don't resolve any name because we used to fetch only aliases in the `IdentityProvider.GetIdentity`.

In this PR:
  * Keep fetching aliases as first resolve attempt to keep backwords compatibility
  * If no alias is found, use aws `organizations.DescribeAccount` to fetch the account **name**.
  * AWS [SecurityAudit](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/SecurityAudit.html) already contains `"organizations:Describe*"`, so its safe to call the `organizations.DescribeAccount`.

### Related Issues

Fixes: https://github.com/elastic/kibana/issues/226433

### Checklist

#### Introducing a new rule?

- [ ] Add relevant unit tests

### DIFF
--- a/internal/flavors/benchmark/aws_org.go
+++ b/internal/flavors/benchmark/aws_org.go
@@ -123,12 +123,12 @@ func (a *AWSOrg) initialize(ctx context.Context, log *clog.Logger, cfg *config.C
 
 // getAwsAccounts returns all the aws accounts of the org.
 // For each account it bundles together the cloud.Identity and the credentials for the cloudbeat-securityaudit role of that account.
-// It requires cloudbeat-root credentials (requires iam:ListAccountAliases and iam:GetRole).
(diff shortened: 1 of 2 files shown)


amirbenun (approved):
LGTM

PR #3566, opened 2025-09-04
go.mod: Remove unecessary replace directives

orestisfl:
### Summary of your changes

The sarama, jwx and kubernetes replaces have no effect.

The goproxy replace seem to be arbitrary, without a core reason.


moukoublen:
Just a note that ee need to be in sync with https://github.com/elastic/beats/blob/main/go.mod#L520-L529 

(we might not need all of the replaces but the ones we need we need them to be the same)

moukoublen (approved):

PR #3775, opened 2025-12-17
Add retry logic for Hermit initialization

gurevichdmitry:
### Summary of your changes

There is no built-in GitHub Actions mechanism to retry a step on failure. Since this is a relatively simple use case, it is preferable to handle retries with a custom script rather than introducing a third-party action that would require additional version management and maintenance.

This PR introduces a retry script and applies it to the Hermit initialization command to make the CI workflows more resilient to transient failures.

### DIFF
--- a/.github/workflows/destroy-environment.yml
+++ b/.github/workflows/destroy-environment.yml
@@ -62,97 +62,43 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ${{ env.AWS_REGION }}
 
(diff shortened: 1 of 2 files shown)


uri-weisman:
I understand those changes allows us to retry on a single steps, but other steps might also be flaky and we might not want to apply the retry block for all of them.
Should we consider using [rerun-flaky-workflows](https://github.com/elastic/cloudbeat/blob/main/.github/workflows/rerun-flaky-workflows.yml) to automatically re-run the GitHub Action if it fails?

uri-weisman (approved):

Issue #3554, opened 2025-09-01
[Asset Discovery] Migrate to new nested `*.entity.*` schema


romulets:
Based on the [discussion,](https://github.com/elastic/security-team/discussions/13139#discussioncomment-14093414) we want to avoid entity duplications.

We need to mind that a migration is a breaking change. I suggest doing a small research on the complexity and cost of introducing a breaking change and possible migration. However, if the cost is too high, I suggest using the [alias field type](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/field-alias) to ease the migration (mind that this will probably be a forever decision too, making the migration might be a hard pill to swallow, but will make the data better for the future).

Depends on:
- [[Entity Store] Support *.entity.id fields on transform pivot #233603](https://github.com/elastic/kibana/issues/233603).

**DoD**
- [ ] Previous versions of Asset Discovery can be upgraded without issues to the new format
- [ ] User, Host and Service publish entity data in the new structure,

uri-weisman:
FYI @caitlinbetz  - We need to modify the schema for asset discovery events (a breaking change) and are exploring the best approach to do so with the smallest possible impact.

chennn1990:
@uri-weisman A follow up [task](https://github.com/elastic/cloudbeat/issues/3808) to remove `entity.*`

PR #3791, opened 2025-12-24
copy entity to user & hosts structs

chennn1990:
### Summary of your changes
Asset Inventory change:
create a nested hierarchy inside User & Host so the entity will be appeared also there instead of a sibling.

### Related Issues

- Related: 
https://github.com/elastic/cloudbeat/issues/3554
https://github.com/elastic/kibana/issues/233603

### Checklist


internal/inventory/azurefetcher/fetcher_resource_graph.go
```diff
@@ -107,6 +107,7 @@ func (f *resourceGraphFetcher) fetch(ctx context.Context, resourceName, serviceN
 					Name: vmProperties.Extended.InstanceView.ComputerName,
 					Type: vmProperties.HardwareProfile.VmSize,
 				}
+				asset.Host.Entity = &asset.Entity
```
uri-weisman: Not necessary related to your changes, but I wonder why we don't use the `inventory.WithHost` utility here.

uri-weisman (approved):
LGTM 🚀  A few notes:

1. Let’s validate this on GCP and Azure before merging.
2. We should open a ticket to remove the redundant `entity.*` fields for host and user entities once we pivot to `user|host.entity.id`. This can be tracked as an action item under [https://github.com/elastic/kibana/issues/233603](https://github.com/elastic/kibana/issues/233603).
3. There are currently no mappings for these fields. we’ll need to add them in the [integrations repo](https://github.com/elastic/integrations/tree/5246e51180fd8d2537088b49e48299768c0eef36/packages/cloud_asset_inventory/data_stream/asset_inventory/fields):
4. Please don't backport to `9.3`

internal/inventory/azurefetcher/fetcher_resource_graph.go
uri-weisman: Not with linting or something like that, maybe unit tests can help enforce it, but I'd prefer the tests to be focus on a single class and not on the utilities.

PR #3836, opened 2026-01-11
feat: add GCP Cloud Connectors Terraform infrastructure

amirbenun:
Adds a new Terraform module for deploying GCP Cloud Connectors infrastructure using GCP Infrastructure Manager. This enables customers to set up keyless AWS-to-GCP federation through Workload Identity, allowing Elastic's AWS role to impersonate a customer-owned GCP service account for security monitoring. The module creates a Workload Identity Pool and Provider, a target service account with Cloud Asset Viewer and Browser roles, and implements two-layer security validation (AWS role ARN + session name matching). Includes deployment scripts for Cloud Shell and comprehensive documentation.

### DIFF
--- a/deploy/infrastructure-manager/gcp-cloud-connectors/deploy.sh
+++ b/deploy/infrastructure-manager/gcp-cloud-connectors/deploy.sh
@@ -1,12 +1,15 @@
 #!/bin/bash
 set -e
 
+# Get the directory where this script lives (for Terraform source files)
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+
 # Configure GCP project
 PROJECT_ID=$(gcloud config get-value core/project)
 SERVICE_ACCOUNT="infra-manager-deployer"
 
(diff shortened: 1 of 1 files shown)


deploy/infrastructure-manager/gcp-cloud-connectors/main.tf
```diff
@@ -0,0 +1,55 @@
...
+}
+
+provider "google" {
+  project = var.project_id
+}
+
+locals {
+  # Use suffix from deploy.sh to ensure all resource names stay within GCP limits
+  resource_suffix = var.resource_suffix
+  sa_name         = "${var.target_service_account_name}-${local.resource_suffix}"
```
jeniawhite: We can have an overflow on the service account name.
I suggest validation including the suffix.

- https://docs.cloud.google.com/iam/docs/service-accounts-create

```
The ID must be between 6 and 30 characters, and can contain lowercase alphanumeric characters and dashes
```

PR #3893, opened 2026-01-28
fix project metadata request

orouz:
## Changes
- add rate limiting for GCP Projects API to prevent 429 quota errors
- add singleflight to deduplicate concurrent requests for the same project metadata

## Context
- a while ago i merged a PR to cache the org name so we don't get rate limited for the org api. 
- latest errors show we're hitting rate limit errors for project api, this is what this PR fixes

```
googleapi: Error 429: Quota exceeded for quota metric 'Project V3 get requests' 
```

The quota is 600 requests per minute. with many projects and assets arriving concurrently, we could exceed this before the cache was populated.

### DIFF
--- a/.buildkite/scripts/package.sh
+++ b/.buildkite/scripts/package.sh
@@ -17,7 +17,6 @@ if [ "$WORKFLOW" = "snapshot" ]; then
 fi
 
 mage pythonEnv
-$PYTHON_BIN/pip install setuptools
 mage package
 
 CSV_FILE="build/dependencies-${CLOUDBEAT_VERSION}"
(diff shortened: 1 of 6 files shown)


internal/resources/providers/gcplib/inventory/resource_manager.go
```diff
@@ -50,36 +55,43 @@ func NewResourceManagerWrapper(ctx context.Context, log *clog.Logger, gcpConfig
...
-			if err != nil {
-				log.Errorf("error fetching GCP Org: %s, error: %s", parent, err)
-				return ""
-			}
-			orgName = org.DisplayName
+		projectsRateLimiter:  rate.NewLimiter(rate.Every(time.Minute/600), 1),
+	}
+	wrapper.getProjectDisplayName = func(ctx context.Context, parent string) string {
+		if err := wrapper.projectsRateLimiter.Wait(ctx); err != nil {
+			log.Errorf("rate limiter error for GCP Project: %s, error: %s", parent, err)
```
romulets: Is rate limiting error really an error? The expected behaviour is just to block and then release once we have room for more requests?

orouz: > The expected behaviour is just to block and then release once we have room for more requests?

yes

> Is rate limiting error really an error?

it is if we fail to `Wait()`. the likely reason is a canceled context. we don't log it just for blocking.

PR #3973, opened 2026-02-16
Revert "[9.3] Bump github.com/sigstore/cosign/v2 and github.com/sigstore/rekor"

gurevichdmitry:
Reverts elastic/cloudbeat#3956


gurevichdmitry:
We have found a way to fix all issues related to linting and unit tests [there](https://github.com/elastic/cloudbeat/pull/3966), so there is no need to revert this PR.

PR #4355, opened 2026-04-12
Fix destroy environment workflow failure due to missing variable

gurevichdmitry:
### Summary of your changes

The destroy environment workflow currently fails because a required variable is missing.
This PR adds the missing variable and sets it to the correct value to ensure the workflow runs successfully.

### DIFF
--- a/deploy/test-environments/manage_infrastructure.sh
+++ b/deploy/test-environments/manage_infrastructure.sh
@@ -4,9 +4,10 @@
 run_terraform() {
     local dir=$1
     local operation=$2
+    local terraform_rc
 
     echo "Running Terraform $operation in $dir..."
-    cd "$dir" || exit
+    cd "$dir" || exit 1
 
     case $operation in
     "apply")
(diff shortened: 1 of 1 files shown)


deploy/test-environments/manage_infrastructure.sh
```diff
@@ -20,6 +21,10 @@ run_terraform() {
             echo "Removing aws_auth resource from state in cis..."
             terraform state rm "$(terraform state list | grep "kubernetes_config_map_v1_data.aws_auth")"
         fi
+        # Destroy still evaluates module variable validation; CDR apply sets TF_VAR_* in CI, generic destroy does not.
```
uri-weisman, line 24: Can you please explain why this is needed?

gurevichdmitry: The CDR stack passes `var.windows_elastic_defend_winrm_ingress_cidr` into modules/aws/ec2-windows, which [rejects an empty string (and 0.0.0.0/0)](https://github.com/elastic/cloudbeat/blob/main/deploy/test-environments/modules/aws/ec2-windows/variables.tf#L30-L38)
**Destroy** process did not set that env var, so the variable stayed at its default `""`
As a result **Destroy** then failed variable validation (winrm_ingress_cidr must be set…) even though you only wanted to tear resources down.

```
╷
│ Error: Invalid value for variable
│ 
│   on main.tf line 103, in module "aws_ec2_elastic_defend_windows":
│  103:   winrm_ingress_cidr    = var.windows_elastic_defend_winrm_ingress_cidr
│     ├────────────────
│     │ var.winrm_ingress_cidr is ""
│ 
│ winrm_ingress_cidr must be set to a restrictive CIDR and must not be
│ 0.0.0.0/0, as this would expose WinRM to the public internet.
│ 
│ This was checked by the validation rule at
│ ../modules/aws/ec2-windows/variables.tf:35,3-13.
╵
```

uri-weisman (approved):

PR #4381, opened 2026-04-14
Update naming convention for GCP deployments

gurevichdmitry:
### Summary of your changes
When deploying CDR and CIS integrations, the same name is currently used, which causes conflicts due to an existing account.
This PR updates the naming convention to ensure separate resource creation and avoid these conflicts.


opauloh (approved):
lgtm, thanks for the fix!

uri-weisman (approved):

PR #4382, opened 2026-04-14
Add new EA test environment workflow

gurevichdmitry:
### What this PR does

- Adds a new GitHub Actions workflow you can run by hand to spin up an Elastic test environment for Entity Analytics work.
- Turns on the right Kibana options for that environment.
- After the stack is up, loads org data using `security-documents-generator`.
- Runs ML starter job.

### DIFF
--- a/.github/workflows/create-env-ea.yml
+++ b/.github/workflows/create-env-ea.yml
@@ -1,0 +1,252 @@
+name: create-env-ea
+run-name: EA env ${{ inputs.deployment_name }} by @${{ github.actor }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      deployment_name:
+        type: string
+        description: |
+          Name with letters, numbers, hyphens; start with a letter. Max 20 chars. e.g., 'ea-env-01'
+        required: true
+      elk-stack-version:
+        required: true
+        description: "Stack version: released 8.x.y / 9.x.y, BC 8.x.y-hash, SNAPSHOT 8.x.y-SNAPSHOT"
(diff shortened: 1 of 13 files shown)


olegsu (approved):

.github/workflows/create-env-ea.yml
```diff
@@ -0,0 +1,271 @@
...
+          serverless-mode: false
+          elk-stack-version: ${{ env.STACK_VERSION }}
+          ess-region: ${{ env.ESS_REGION }}
+          ec-api-key: ${{ env.TF_VAR_ec_api_key }}
+          ec-url: ${{ env.EC_URL }}
+          kibana-security-solution-experimental: false
+          kibana-enable-entity-analytics-settings: true
+          env-s3-bucket: "${{ env.S3_BASE_BUCKET }}/${{ env.DEPLOYMENT_NAME }}_${{ env.TF_STATE_FOLDER }}"
+          tag-project: ${{ github.actor }}
+          tag-owner: ${{ github.actor }}
```
uri-weisman, line 154: Shouldn't we need to pass here `docker-image-version-override` as well? to be propagated to `./.github/actions/elk-stack`?

uri-weisman (approved):

.github/workflows/create-env-ea.yml
gurevichdmitry: No, the Docker image override is used only for the Elastic Agent (for KSPM manifests). For the stack, we only provide the stack version.

PR #4384, opened 2026-04-15
Add ML node option to Elasticsearch configuration

gurevichdmitry:
### Summary of your changes

Since we are enabling ML jobs, the current setup creates an ML node only after the jobs are applied, which introduces delays. To improve this, it is better to start the stack with an ML node already configured, especially when using API requests that depend on ML functionality.

This PR updates the Elasticsearch Terraform configuration to include a basic ML node setup.


uri-weisman (approved):

opauloh (approved):
awesome

PR #6991, opened 2026-06-26
[Cloud Asset Inventory] Add extra resource fields #1

kubasobon:
## Summary

Follow-up to InfoSec's review of #6991. Adds the remaining EC2 and ELB
fields InfoSec needs to retire their custom AWS CLI scripts for the FedRAMP
asset inventory. All new fields land in `entity.attributes.*` (non-ECS,
UpperCamelCase) — no ECS mapping change required.

Tracking: https://github.com/elastic/security-team/issues/17750

## Changes

**EC2** (`entity.attributes`):
- `ImageId`, `Platform`, `VpcId`, `SubnetId`, `State`, `RoleArn`
- `Owner`, `CostCenter` — resolved case-insensitively from instance tags

**ELB v1 + v2** (`entity.attributes`):
- `LoadBalancerType` — v2 reports the real type; v1 reports `classic`
- `AccountID`
- `State` — v2 only (`State.Code`); classic ELB exposes no state field
- `IPAddresses` — NLB static IPs; ALB/classic are DNS-only (empty)
- `OwnerTag` — required adding `DescribeTags` to both ELB `Client`
  interfaces and providers (batched to the 20-item AWS limit)

## Notes
- No `elastic/integrations` change: `entity.attributes` is `flattened`, so
  new keys need no mapping and there's no migration.
- No new asset classifications, so `ASSETS.md` is unchanged. Route53 and
  EKS (also requested in the review) are separate follow-up PRs.
- Mocks regenerated; unit tests cover EC2 attributes and the ELB
  Type/State/IPAddresses/OwnerTag flow end-to-end.


internal/inventory/awsfetcher/fetcher_ec2_instance.go
```diff
@@ -97,14 +100,78 @@ func (e *ec2InstanceFetcher) Fetch(ctx context.Context, assetChannel chan<- inve
...
+		attrs["SubnetId"] = v
+	}
+	if i.State != nil {
+		if v := string(i.State.Name); v != "" {
+			attrs["State"] = v
+		}
+	}
+	if i.IamInstanceProfile != nil {
+		if v := pointers.Deref(i.IamInstanceProfile.Arn); v != "" {
+			attrs["RoleArn"] = v
```
romulets: Is `IamInstanceProfile.Arn` directly understandable from a ec2 instance `attributes.RoleArn`? Isn't it better perhaps to be more explicit on the naming?

Just asking, I'm ok with your call.

internal/inventory/asset.go
```diff
@@ -178,21 +178,25 @@ type URL struct {
...
+	Lifecycle  *Lifecycle     `json:"lifecycle,omitempty"`
 	AssetClassification
 
 	// non exported fields
 	relatedEntityId []string
 }
 
+// Lifecycle holds non-ECS lifecycle timestamps for an asset (non-ECS fields use UpperCamelCase).
+type Lifecycle struct {
+	CreatedAt *time.Time `json:"CreatedAt,omitempty"`
```
romulets: https://github.com/elastic/kibana/blob/main/x-pack/solutions/security/plugins/entity_store/common/domain/definitions/common_fields.ts#L140-L156

This field is not extracted. If we extract, I'm not sure we should have it capitalised, it doesn't follow the other fields

internal/inventory/azurefetcher/fetcher_resource_graph.go
```diff
@@ -138,3 +140,28 @@ func tryUnpackingVMProperties(m map[string]any) *vmProperties {
 	}
 	return o
 }
+
+// extractAzureCreatedAt tries common ARM property keys that carry a resource creation
+// timestamp. Values may arrive as time.Time (from typed SDK providers) or as RFC3339 strings
+// (from the Resource Graph JSON response). Returns nil when no creation time is found.
+func extractAzureCreatedAt(properties map[string]any) *time.Time {
+	for _, key := range []string{"createdTime", "timeCreated", "creationTime", "creationDate"} {
```
romulets, line 148: this is interesting lol. I won't challenge because I know how azure is, but not super ideal haha

internal/resources/providers/awslib/elb/load_balancer.go
```diff
@@ -30,6 +32,17 @@ type ElasticLoadBalancerInfo struct {
...
+
+// lookupOwnerTag returns the value of the "Owner" tag (case-insensitive), if present.
+func lookupOwnerTag(tags map[string]string) string {
+	for k, v := range tags {
+		if strings.EqualFold(k, "owner") {
+			return v
+		}
+	}
+	return ""
 }
```
romulets: not a biggie, but this is a repetitive code with internal/resources/providers/awslib/elb_v2/load_balancer_v2.go (and I have a feeling that I've seen more)

internal/resources/providers/awslib/rds/provider.go
romulets, line 97: memory thanks you

romulets (commented):
Biggest question is about the lifecycle event which we don't extract

internal/resources/providers/awslib/elb/load_balancer.go
kubasobon: Good point, I'll wrap it up into one helper.

internal/inventory/awsfetcher/fetcher_ec2_instance.go
kubasobon: `IamInstanceProfile.Arn` is a not strictly a Role ARN, according to AWS docs - it's a Identity Profile attached to the EC2 (which contains the role).

internal/inventory/asset.go
kubasobon: Yes, I'm aware. It's an untracked non-ECS, but I'm adding it to the integration on InfoSec's ask. It's not entity's creation date, but the resource's.

internal/inventory/azurefetcher/fetcher_resource_graph.go
kubasobon: Well, that's just Azure being Azure 😅

romulets (approved):

PR #7175, opened 2026-07-01
Fix: GH action test-runner.yml

groman92:
### Summary of changes
test-runner.yml was failing because the agent version pinned in bin/hermit.hcl was stale. CLOUDBEAT_VERSION was pinned to 9.4.0, so CI kept requesting the 9.4.0-SNAPSHOT artifact, which no longer exists, and the agent download returned a 404.
Root cause

bin/hermit.hcl pinned CLOUDBEAT_VERSION = 9.4.0, out of sync with the current version in version.go (9.5.0).
This caused the agent download step to request a snapshot that had already been cleaned up, failing the job ~30 minutes in with a 404, well after the run had already progressed.

### Changes

- Bumped CLOUDBEAT_VERSION in bin/hermit.hcl to 9.5.0 to match version.go.
- Fixed a cold-cache bug in the Hermit composite action: the "Log cached paths" step now no-ops (|| true) instead of failing when the cache directory doesn't exist yet.
- Bumped Chrome/ChromeDriver in the Kibana FTR action from 148 to 149 to fix a version mismatch that was breaking UI tests.

### Verification
Full flow (deploy → enroll → UI tests) passes end-to-end: [Working Run](https://github.com/elastic/cloudbeat/actions/runs/28497415353)

### Next steps / follow-ups

- Per discussion with @gurevichdmitry : automating the version bump isn't as simple as re-enabling the trigger — when the bump happens immediately, the corresponding agent artifacts aren't published yet, which can cause agent failures. That's why the bump is currently done manually a couple of days after a release. I'll investigate how to improve this process (e.g. delayed/scheduled bump, or gating the bump on artifact availability) as a follow-up.

- sync-internal-cloudbeat-version.yml is the workflow meant to keep hermit.hcl aligned with version.go, but its push trigger is commented out, so it only runs manually and never fired for this bump. We should re-enable it or add a schedule.

- Longer term, it'd be good to make get_build_info validate the version against artifacts-api.elastic.co so a bad pin fails fast with a clear error instead of dying ~30 minutes into a run.


gurevichdmitry (approved):
LGTM

gsarantid (approved):
LGTM

PR #7178, opened 2026-07-03
Enrich get build info lin

groman92:
### Summary of your changes

get_build_info() resolves the Elastic build ID for a given stack version from the artifact server. On failure it previously swallowed the error and returned an empty string, which produced a malformed agent download URL (missing the build ID). 

get_build_info() now raises a clear, actionable RuntimeError instead of returning "", and the message lists the versions currently published on the Elastic artifacts API so the fix is obvious:

```
RuntimeError: Could not resolve build info for '9.4.0-SNAPSHOT' (status 404 from
https://snapshots.elastic.co/latest/9.4.0-SNAPSHOT.json). The version is likely not
published on Elastic's artifact servers — check CLOUDBEAT_VERSION/ELK_VERSION in
bin/hermit.hcl. Available versions: [..., '9.5.0-SNAPSHOT']
```

- Added get_available_versions() helper that queries https://artifacts-api.elastic.co/v1/versions (fails soft — logs a warning and returns [] if the list can't be fetched, so it never masks the primary error).
- Added the ARTIFACTS_API_VERSIONS_URL constant.


tests/fleet_api/common_api.py
```diff
@@ -569,3 +574,13 @@ def get_telemetry(cfg: Munch) -> dict:
             f"API call failed, status code {api_ex.status_code}. Response: {api_ex.response_text}",
         )
         raise
+
+
+def get_available_versions() -> list:
+    """Return the versions currently published on the Elastic artifacts API."""
+    try:
+        response = perform_api_call(method="GET", url=ARTIFACTS_API_VERSIONS_URL)
+        return munchify(response).get("versions", [])
```
gurevichdmitry, line 583: nit: `munchify` isn’t needed — a single .get("versions", []) on the raw response is enough, same as get_stack_latest_version() / get_package_version() in this file. Fine to drop for consistency; no perf impact on the failure path.

gurevichdmitry (approved):
LGTM

PR #7179, opened 2026-07-03
Handle sync cloudbeat with snapshots not published

groman92:
### Summary of your changes
Makes the hermit.hcl version sync self-healing and safe to automate.

Two changes:

- scripts/sync_internal_cloudbeat_version.sh, before opening a sync PR, the script now checks that the target ${CLOUDBEAT_VERSION}-SNAPSHOT is actually published on the Elastic artifacts API (is_snapshot_published). If it isn't available yet, it reverts the local edit and exits cleanly, so we never pin a version whose agent artifacts don't exist. If the API can't be reached, it skips to stay safe.

- .github/workflows/sync-internal-cloudbeat-version.yml, added a daily schedule trigger (the previous push trigger was commented out, so the sync only ever ran manually and hermit.hcl drifted out of sync with version.go).

Together this closes the gap that caused [security-team#17923](https://github.com/elastic/security-team/issues/17923): hermit.hcl was stuck at 9.4.0 while version.go had moved on, so the test-runner requested a 9.4.0-SNAPSHOT agent that no longer exists. 

With this, the sync runs daily and opens a hermit.hcl bump PR automatically as soon as the new snapshot is published, respecting the "artifacts must exist first" constraint that previously required a manual run.


gurevichdmitry (commented):
Overall LGTM.
Suggestion: Once the snapshot is published the workflow opens a PR, but `main` still has the old `hermit.hcl` until that PR merges. The next scheduled run sees the same drift, passes the snapshot check again, and can open a second PR (`sync-cloudbeat-version-<timestamp>)`. Worth adding a quick check before gh pr create — e.g. skip if there's already an open PR matching the sync title/pattern (gh pr list --search "..." --state open).

scripts/sync_internal_cloudbeat_version.sh
```diff
@@ -18,12 +19,62 @@ set_hermit_cloudbeat_version() {
...
+        return
+    fi
+
+    # A sync PR for this version may already be open from a previous run
+    # (main stays on the old version until it merges) — don't open a duplicate.
+    if sync_pr_already_open; then
+        echo "An open sync PR for $CLOUDBEAT_VERSION already exists; skipping."
+        git checkout -- "$HERMIT_FILE"
+        return
+    fi
```
groman92, line 68: Added here a gatekeeper @gurevichdmitry

gurevichdmitry (approved):
LGTM

PR #7186, opened 2026-07-06
fix: bump main to next minor after minor branch-out

gsarantid:
## Summary

Prep for the Jul 07 minor FF ([context](https://elastic.slack.com/archives/C0JFN9HJL/p1780580577798249)): release-eng will trigger our pipeline with `WORKFLOW=minor BRANCH=9.5 NEW_VERSION=9.5.0` — `NEW_VERSION` is the version for the **new branch**, not for `main`, so each team owns bumping `main` to the next minor themselves.

`main` is already on `9.5.0`. Today's `post-minor-merge.sh` only branches out `9.5` and syncs `hermit.hcl` — nothing advances `main` afterward, so it would stay on `9.5.0` indefinitely. (Confirmed this isn't hypothetical: two ad-hoc "Bump to 9.6.0" PRs from the previous cycle, #5608/#5609, were closed unmerged.)

- Add `next_minor_version()` semver helper to `common.sh`
- Add `bump_main_to_next_minor` step to `post-minor-merge.sh`, run **after** branch-out (so `9.5` is cut while `main` is still `9.5.0`) and before the hermit sync — opens a `Bump cloudbeat version to 9.6.0` PR, same convention as prior manual bumps
- Reuses the existing `pr_exists` helper (matching `common.sh`'s documented `BUMP_BRANCH` contract) instead of duplicating the open-PR check
- Add matching PR body template

Replaces #7181, rebased on latest `main` (picks up the now-merged #7184 idempotency fix). Recreated on a fresh branch because pushing further commits to the old branch was rejected with `GH006: 4 of 4 required status checks are expected` — this has now happened on **two different branches** after their PRs accumulated completed CI checks (both plain pushes and force-pushes after rebase), while brand-new branches push fine every time. This looks like a real, systemic policy issue worth raising with release-eng/repo admins, since it'll recur on any PR that needs a follow-up push.

## Test plan

- [x] Dry-run verified end-to-end against a scratch clone with `WORKFLOW=minor BRANCH=9.5 NEW_VERSION=9.5.0 DRY_RUN=true` — correct sequencing, semver math (`9.5.0` → `9.6.0`), commit messages, and PR body
- [x] Verified idempotent skip when a bump PR is already open / main already at target version
- [x] Verified clean rebase onto main (post-#7184 merge) — no leftover conflicts, no duplicate helper functions, bump-minor.sh/bump-patch.sh untouched
- [x] Confirm `backport-v9.5.0` label is created in the repo before the FF (separate manual step, not covered here)
- [x] Real run against a fork with `gh` to validate actual PR creation

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.buildkite/scripts/release/common.sh
+++ b/.buildkite/scripts/release/common.sh
@@ -61,6 +61,16 @@ no_new_commits() {
     git diff --quiet "$1" HEAD
 }
 
+# next_minor_version <version>
+# Given "X.Y.Z", returns the next minor version "X.(Y+1).0".
+next_minor_version() {
+    local version="$1"
+    local major minor
+    major=$(echo "${version}" | cut -d. -f1)
+    minor=$(echo "${version}" | cut -d. -f2)
+    echo "${major}.$((minor + 1)).0"
+}
+
 # render_template <path>
 # Expands ${VAR} references in a template file using the caller's environment.
 render_template() {
--- a/.buildkite/scripts/release/post-minor-merge.sh
+++ b/.buildkite/scripts/release/post-minor-merge.sh
@@ -25,7 +25,7 @@ GH_REPO="elastic/${REPO}"
 DRY_RUN="${DRY_RUN:-false}"
 HERMIT_BRANCH="sync-cloudbeat-version-$(date +%s)"
 NEXT_MAIN_VERSION=$(next_minor_version "${NEW_VERSION}")
-BUMP_MAIN_BRANCH="bump-to-${NEXT_MAIN_VERSION}"
+BUMP_BRANCH="bump-to-${NEXT_MAIN_VERSION}"
 
 git fetch origin main
 git checkout main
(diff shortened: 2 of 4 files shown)


groman92:
There's a problem with syncing hermit.hcl here: if we update it before the -SNAPSHOT for the new version is published, the build breaks as the agent artifact won't exist yet to download.

We're fixing exactly this in [#7179](https://github.com/elastic/cloudbeat/pull/7179), where @gurevichdmitry  suggested, we add a gatekeeper that only bumps hermit.hcl once the target snapshot is actually published on the Elastic artifacts API.

So I'd suggest removing the hermit_pr step from this script and keeping the rest (next_minor_version / bump_main_to_next_minor). The gated sync in [#7179](https://github.com/elastic/cloudbeat/pull/7179) will then own hermit.hcl and only open the bump PR once the snapshot is available, which avoids the ungated bump here breaking the build.

gsarantid:
> There's a problem with syncing hermit.hcl here: if we update it before the -SNAPSHOT for the new version is published, the build breaks as the agent artifact won't exist yet to download.
> 
> We're fixing exactly this in [#7179](https://github.com/elastic/cloudbeat/pull/7179), where @gurevichdmitry suggested, we add a gatekeeper that only bumps hermit.hcl once the target snapshot is actually published on the Elastic artifacts API.
> 
> So I'd suggest removing the hermit_pr step from this script and keeping the rest (next_minor_version / bump_main_to_next_minor). The gated sync in [#7179](https://github.com/elastic/cloudbeat/pull/7179) will then own hermit.hcl and only open the bump PR once the snapshot is available, which avoids the ungated bump here breaking the build.

Good catch, thanks! Removed the `hermit_pr` step and the now-unused `HERMIT_BRANCH` var — this PR now only does branch-out + the `main` version bump. #7179 owns `hermit.hcl` going forward.

groman92 (approved):
thanks for your changes!

PR #7277, opened 2026-07-17
[Asset Inventory][AWS] Update EC2, ELB and RDS fetchers for InfoSec

kubasobon:
### Summary of your changes

> [!NOTE]
> **Stacked on #7473.** This PR is based on `asset-inventory/fix-entity-attributes`, which renames the `entity.attributes` (flattened) bag to `entity.Details` to fix the Entity Store generic-extraction bug. All new fields below are therefore emitted under **`entity.Details.*`** (not `entity.attributes.*`). Please review/merge #7473 first; this PR's diff shows only the InfoSec fetcher changes on top of it.

| Resource             | Field                             | Change                                                                                                                   |
| -------------------- | --------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
| **EC2**              | `entity.Details.Role`             | Added `LookupTag(tags, "role")` in `buildDetails`                                                                        |
| **RDS**              | `entity.Details.DBInstanceStatus` | Added `Status` field to wrapper struct; mapped from `DBInstanceStatus` in provider; emitted in fetcher                   |
| **ELB v2** (ALB/NLB) | `entity.Details.IPAddresses`      | `GetIPAddresses()` now also collects `PrivateIPv4Address` and `IPv6Address` per AZ address (previously only `IpAddress`) |
| **ELB v1** (Classic) | `entity.Details.State`            | `GetState()` returns `"active"` (hardcoded as classic API exposes no state field)                                       |
| **ELB v1** (Classic) | `entity.Details.IPAddresses`      | DNS-resolves the ELB `DNSName` at fetch time via injectable `hostResolver`; soft-fails to empty on error                 |

### Related Issues

Closes https://github.com/elastic/security-team/issues/18294

### Checklist


internal/resources/providers/awslib/elb/elb_mock.go
```diff
@@ -31,6 +31,101 @@ import (
 	mock "github.com/stretchr/testify/mock"
 )
 
+// newMockHostResolver creates a new instance of mockHostResolver. It also registers a testing interface on the mock and a cleanup function to assert the mocks expectations.
+// The first argument is typically a *testing.T value.
+func newMockHostResolver(t interface {
```
jeniawhite, line 36: I think that these are not being utilized. Is there a reason to implement them?
`newMockHostResolver, mockHostResolver, mockHostResolver_Expecter, mockHostResolver_LookupHost_Call`

internal/inventory/awsfetcher/fetcher_elb_test.go
```diff
@@ -84,6 +84,7 @@ func TestELBv1Fetcher_Fetch(t *testing.T) {
 				"PubliclyAccessible": false, // scheme is "internal"
 				"AccountID":          "123",
 				"LoadBalancerType":   "classic",
+				"State":              "active",
```
jeniawhite, line 91: Does it makes sense to add IPAddresses as well? Because we don't run the DNS resolver (like we do in the actual code to resolve the IPs).
Reference:
https://github.com/elastic/cloudbeat/pull/7277/changes#diff-45e6ce26ce27502d2d082db4502e0999a52d546b187833699c645fbfa297c146R88-R96

kubasobon:
@jeniawhite Thank you, both good catches. I've pushed an update.

PR #7423, opened 2026-07-22
Migrate environment workflow for agentless CSPM setup to /api/fleet/managed_integrations

seanrathier:
## Summary

- Replaces the deprecated two-step flow (`POST /api/fleet/agent_policies` with `supports_agentless: true` + `POST /api/fleet/package_policies`) with a single `POST /api/fleet/managed_integrations` call
- Adds `tests/fleet_api/managed_integration_api.py` with `create_managed_integration` and `delete_managed_integration` helpers
- Updates `purge_integrations.py` to call `DELETE /api/fleet/managed_integrations/{id}` (which handles full cleanup) when the policy was created via the new API

**Context:** `/api/fleet/agent_policies` and `/api/fleet/package_policies` are being deprecated for agentless integrations in 9.5 GA in favor of `/api/fleet/managed_integrations`. A feature flag to disallow the old path is being monitored and will be enabled in a future release. See [Slack thread](https://elastic.slack.com/archives/C02E83LRFMX/p1784631242374389) for context.

## Test plan

- [x] Run `install_agentless_integrations.py` against a serverless project and verify CSPM AWS/Azure/GCP integrations are created successfully via the new endpoint
- [x] Run `purge_integrations.py` and verify the managed integrations are deleted cleanly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/integrations_setup/purge_integrations.py
+++ b/tests/integrations_setup/purge_integrations.py
@@ -20,6 +20,7 @@ from fleet_api.agent_policy_api import (
     get_agents,
     unenroll_agents_from_policy,
 )
+from fleet_api.managed_integration_api import delete_managed_integration
 from fleet_api.package_policy_api import delete_package_policy
 from loguru import logger
 from state_file_manager import state_manager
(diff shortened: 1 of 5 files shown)


tests/integrations_setup/purge_integrations.py
```diff
@@ -29,12 +30,18 @@ def purge_integrations():
...
-
     agents = get_agents(cfg=cnfg.elk_config)
     # Delete policies based on the stored IDs
     for policy in state_manager.get_policies():
         logger.info("Deleting policy", policy.pkg_policy_id, policy.agnt_policy_id)
+
+        # When agnt_policy_id == pkg_policy_id the entry was created via the managed_integrations
+        # API (a single combined resource). Deleting the managed integration cleans up both the
+        # agent policy and package policy automatically.
+        if policy.agnt_policy_id == policy.pkg_policy_id:
```
jeniawhite: Comparing IDs is fragile, future policy type that stores the same ID in both fields would be routed to the wrong deletion path.
I would suggest adding an explicit `is_managed: bool = False` field to PolicyState?

jeniawhite (approved):

PR #8385, opened 2026-09-08
[CFT] Add Federated Identity template for Elastic Workload Identity (WII)

jeniawhite:
### Summary of your changes

Adds `deploy/cloudformation/federated-identity-wii-aws.yml`, the WII (Elastic Workload Identity Issuer) counterpart of `federated-identity-aws.yml`. Same role and grants, but the trust is an OIDC provider for the WII issuer of the customer's organization, region and CSP, and the role can only be assumed by one deployment (ECH) or project (Serverless):

```
<issuer-host>/orgs/<org-id>:aud = sts.amazonaws.com
<issuer-host>/orgs/<org-id>:sub = <deployment|project>:<id>
```

Inputs are the facts about the Elastic deployment or project, pre-filled by Kibana; the template composes the two JWT claims the trust is built from:

- `ElasticOrganizationId`, `ElasticCloudProvider` (`aws`/`azure`/`gcp`), `ElasticCloudRegion` and `ElasticCloudEnvironment` (`production`/`staging`/`qa`) give the issuer: `workload-identity-issuer.<region>.<provider>.<domain>/orgs/<org>`, with the domain per environment in a `Mappings` block. Only Elastic-owned issuers can result, and Kibana does not need to know the WII hostname scheme.
- `ElasticResourceType` (`deployment`/`project`) and `ElasticResourceId` (Kibana component ID or Serverless project ID) give the subject: `<type>:<id>`.

The custom resource returns the issuer it registered and the trust policy reads it back with `GetAtt`, so the issuer is composed in two places only (the function role's resource ARN and the provider URL). A failed `AssumeRoleWithWebIdentity` can be debugged by comparing the role's trust conditions against the two decoded JWT claims.

The template is also added to `scripts/publish_cft.sh`.

#### OIDC provider

IAM allows one OIDC provider per issuer URL per account, and `AWS::IAM::OIDCProvider` fails with `EntityAlreadyExists` if it exists. The provider is shared by every stack for the same org, region and CSP in an account: other ECH deployments and Serverless projects, repeated onboardings of the same one, and every template update (updates replace the stack). There is no shared state across those to record whether the provider already exists, so neither a parameter nor a separate provider template can be driven reliably.

The provider is therefore managed by a Lambda-backed custom resource (`WiiOidcProvider`) that:

- creates the provider only if it does not exist, and adds the `sts.amazonaws.com` audience if an existing provider lacks it;
- does nothing on delete, so replacing or deleting one stack never breaks the roles of the others (AWS: roles referencing a deleted provider stop working). The `OidcProviderArn` output tells the customer what to delete once the last stack is gone;
- retries transient failures with exponential backoff (2/4/8/16 s): `EntityAlreadyExists` (lost a race with another stack), `AccessDenied` (role policy still propagating), `OpenIdIdpCommunicationError` (IAM contacts the issuer during create), `ConcurrentModification`, `Throttling`, `ServiceFailure`. Every other error is reported back with the IAM message so the stack fails fast;
- has `ServiceTimeout: 600`, so a function that never answers fails the stack after 10 minutes instead of CloudFormation's default hour, while leaving room for Lambda's asynchronous retries.

The function role only has `iam:GetOpenIDConnectProvider`, `iam:CreateOpenIDConnectProvider` and `iam:AddClientIDToOpenIDConnectProvider` on the one provider ARN, plus the Lambda basic execution role so a failing create is visible in CloudWatch Logs: no delete, no network beyond the CloudFormation callback. It cannot register any other issuer even if invoked directly. The `sts.amazonaws.com` audience is fixed in the function rather than passed as a property. No `ThumbprintList`: IAM resolves it for publicly trusted CAs.

Prior art: AWS's own GitHub Actions sample uses a parameter for an existing provider ARN; eksctl and Datadog's EKS tooling create the provider from code before CloudFormation and never delete it; CDK's Lambda-backed provider does not handle an existing one and is now legacy. Nobody solves this inside CloudFormation without code.

#### Why not trust all WII endpoints

WII issues tokens with a per-region `iss`, so in production that is 63 issuers. IAM does not allow wildcards in `Federated` principals, so all 63 provider ARNs would have to be listed in the trust policy (~7.5k chars, default quota 2048, max 8192). Adding the `sub` condition needs one statement per issuer, ~29k chars, which cannot fit at all. A deployment or project lives in one region anyway, so one issuer per stack is both sufficient and the smallest trust.

#### Why no loop

`Fn::ForEach` can create N providers from a list parameter, but it cannot produce list items, so there is no way to feed a variable number of provider ARNs into the role's trust policy. It also needs the `AWS::LanguageExtensions` transform, and with one issuer per stack there is nothing to loop over.

#### Differences from the legacy template

Inline policy names drop the `-${AWS::StackName}` suffix: they only need to be unique within the role, and the suffix pushed long stack names past IAM's 128-character limit. The GuardDuty managed policy ARN uses `${AWS::Partition}`.

### Related Issues

- https://github.com/elastic/kibana-feature-flags/pull/207 — FF lock to not promote changes to customers and limit to internal users for all environments

### Checklist
- [x] `cfn-lint` passes; grants diffed against `federated-identity-aws.yml`
- [x] Custom resource handler exercised against every request type and IAM failure path (create/reuse/add audience/race/propagation/SCP deny/throttling/issuer unreachable/non-retryable/update with new issuer/failed update/delete/rollback delete)
- [x] Deployed against QA for the hosted and serverless flows (parameter-based version)
- [ ] Deploy the custom-resource version against QA: first stack, second stack for the same issuer, delete one, re-create

### DIFF
--- a/deploy/cloudformation/federated-identity-wii-aws.yml
+++ b/deploy/cloudformation/federated-identity-wii-aws.yml
@@ -1,0 +1,384 @@
+AWSTemplateFormatVersion: "2010-09-09"
+
+Description: 'Creates an IAM Role for Elastic Workload Identity Federation. The role can be assumed only by the given Elastic deployment or project and grants read-only permissions for all Elastic AWS integrations that support Identity Federation: the AWS package integrations (GuardDuty, Inspector, Config, CloudWatch, and metrics services)'
+
+Parameters:
+  ElasticOrganizationId:
+    Description: Elastic Cloud organization ID.
+    Type: String
+    AllowedPattern: '^[0-9]+$'
+    ConstraintDescription: must be numeric.
+
+  ElasticResourceType:
+    Description: deployment for Elastic Cloud Hosted, project for Elastic Cloud Serverless.
+    Type: String
+    AllowedValues:
+      - deployment
+      - project
+
+  ElasticResourceId:
+    Description: Kibana component ID of the deployment, or the Serverless project ID.
+    Type: String
+    AllowedPattern: '^[A-Za-z0-9]+$'
+    ConstraintDescription: must contain only letters and digits.
+
+  ElasticCloudProvider:
+    Description: Cloud provider the Elastic deployment or project runs on.
+    Type: String
+    AllowedValues:
+      - aws
+      - azure
+      - gcp
+
+  ElasticCloudRegion:
+    Description: Region the Elastic deployment or project runs in (for example eu-west-1, us-central1, eastus2).
+    Type: String
+    AllowedPattern: '^[a-z0-9-]+$'
+    ConstraintDescription: must be a lowercase region name.
+
+  ElasticCloudEnvironment:
+    Description: Elastic Cloud environment. Use production unless testing against an Elastic test environment.
+    Type: String
+    AllowedValues:
+      - production
+      - staging
+      - qa
+
+  CreateOidcProvider:
+    Description: Set to false if this account already has a stack for the same organization and region, so the existing OIDC provider is reused.
+    Type: String
+    Default: "true"
+    AllowedValues:
+      - "true"
+      - "false"
+
+Mappings:
+  EnvironmentDomain:
+    production:
+      Domain: svc.elastic.cloud
+    staging:
+      Domain: svc.staging.elastic.cloud
+    qa:
+      Domain: svc.qa.elastic.cloud
+
+Conditions:
+  ShouldCreateOidcProvider: !Equals [!Ref CreateOidcProvider, "true"]
+
+Resources:
+
+  # No ThumbprintList: IAM ignores it for publicly trusted CAs.
+  WiiOidcProvider:
+    Type: AWS::IAM::OIDCProvider
+    Condition: ShouldCreateOidcProvider
+    Properties:
+      Url: !Sub
+        - https://workload-identity-issuer.${ElasticCloudRegion}.${ElasticCloudProvider}.${Domain}/orgs/${ElasticOrganizationId}
+        - Domain: !FindInMap [EnvironmentDomain, !Ref ElasticCloudEnvironment, Domain]
+      ClientIdList:
+        - sts.amazonaws.com
+
+  ElasticWorkloadIdentityRole:
+    Type: AWS::IAM::Role
+    Properties:
+      RoleName: !Sub ElasticWorkloadIdentity-${AWS::StackName}
+      AssumeRolePolicyDocument: !Sub
+        - |
+          {
+            "Version": "2012-10-17",
+            "Statement": [
+              {
+                "Effect": "Allow",
+                "Principal": {
+                  "Federated": "${ProviderArn}"
+                },
+                "Action": "sts:AssumeRoleWithWebIdentity",
+                "Condition": {
+                  "StringEquals": {
+                    "workload-identity-issuer.${ElasticCloudRegion}.${ElasticCloudProvider}.${Domain}/orgs/${ElasticOrganizationId}:aud": "sts.amazonaws.com",
+                    "workload-identity-issuer.${ElasticCloudRegion}.${ElasticCloudProvider}.${Domain}/orgs/${ElasticOrganizationId}:sub": "${ElasticResourceType}:${ElasticResourceId}"
+                  }
+                }
+              }
+            ]
+          }
+        - Domain: !FindInMap [EnvironmentDomain, !Ref ElasticCloudEnvironment, Domain]
+          ProviderArn: !If
+            - ShouldCreateOidcProvider
+            - !GetAtt WiiOidcProvider.Arn
+            - !Sub
+              - arn:${AWS::Partition}:iam::${AWS::AccountId}:oidc-provider/workload-identity-issuer.${ElasticCloudRegion}.${ElasticCloudProvider}.${Domain}/orgs/${ElasticOrganizationId}
+              - Domain: !FindInMap [EnvironmentDomain, !Ref ElasticCloudEnvironment, Domain]
+      Path: /
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/AmazonGuardDutyReadOnlyAccess
+
+  ElasticAwsInspector:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: !Sub ElasticAwsInspector-${AWS::StackName}
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - inspector2:ListFindings
+            Resource: '*'
+      Roles:
+        - !Ref ElasticWorkloadIdentityRole
+
+  ElasticAwsCloudwatchLogs:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: !Sub ElasticAwsCloudwatchLogs-${AWS::StackName}
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - logs:DescribeLogGroups
+              - logs:DescribeLogStreams
+              - logs:FilterLogEvents
+              - logs:GetLogEvents
+              - logs:GetLogGroupFields
+            Resource: '*'
+      Roles:
+        - !Ref ElasticWorkloadIdentityRole
+
+  ElasticAwsMetrics:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: !Sub ElasticAwsMetrics-${AWS::StackName}
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - ec2:DescribeRegions
+              - cloudwatch:ListMetrics
+              - cloudwatch:GetMetricData
+              - tag:GetResources
+            Resource: '*'
+      Roles:
+        - !Ref ElasticWorkloadIdentityRole
+
+  ElasticAwsHealth:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: !Sub ElasticAwsHealth-${AWS::StackName}
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - health:DescribeEvents
+              - health:DescribeEventDetails
+              - health:DescribeAffectedEntities
+            Resource: '*'
+      Roles:
+        - !Ref ElasticWorkloadIdentityRole
+
+  ElasticAwsConfig:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: !Sub ElasticAwsConfig-${AWS::StackName}
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - config:DescribeConfigRules
+              - config:GetComplianceDetailsByConfigRule
+            Resource: '*'
+      Roles:
+        - !Ref ElasticWorkloadIdentityRole
+
+  ElasticAwsBilling:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: !Sub ElasticAwsBilling-${AWS::StackName}
+      PolicyDocument:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: Allow
+            Action:
+              - ce:GetCostAndUsage
+            Resource: '*'
+      Roles:
+        - !Ref ElasticWorkloadIdentityRole
+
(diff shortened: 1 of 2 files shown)


seanrathier:
> WII issues tokens with a per-region iss, so in production there are 63 issuers. IAM does not allow wildcards in Federated principals, so all 63 provider ARNs would have to be listed in the trust policy (~7.5k chars; default quota is 2048, max is 8192). Adding the sub condition requires one statement per issuer, pushing it to ~29k chars — which can't fit at all. Since a deployment or project lives in one region anyway, one issuer per stack is both sufficient and the smallest possible trust surface.

If I'm reading your ask correctly: is single-deployment trust good enough for the MVP, or do we need to solve the multi-deployment case first?

This means a RoleArn cannot be shared between, say, a production ECH deployment and a staging ECH deployment.

Sharing a RoleArn across deployments or projects wasn't something I'd accounted for in the upgrade CFT key flow, but it actually creates a real problem. Imagine: we verify the key and it doesn't match because the staging deployment has an extra integration, so we upgrade the CFT. Now the production deployment's trust is broken. We're in a catch-22 loop.

IMO, single-deployment trust is the right path for MVP, especially once you factor in the upgrade CFT flow. Is that aligned with what you're thinking?

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,384 @@
...
+    qa:
+      Domain: svc.qa.elastic.cloud
+
+Conditions:
+  ShouldCreateOidcProvider: !Equals [!Ref CreateOidcProvider, "true"]
+
+Resources:
+
+  # No ThumbprintList: IAM ignores it for publicly trusted CAs.
+  WiiOidcProvider:
```
bhapas, line 70: Should this have `DeletionPolicy: Retain`. 

IAM allows one provider per URL per account. The first stack owns that provider; later stacks set `CreateOidcProvider=false` and reference it by ARN. Deleting the first stack removes the provider and breaks every reused stack.

- Stack A creates the IAM OIDC provider (one per URL per account).
- Stack B sets `CreateOidcProvider=false` and points its role at the same provider ARN by convention.
- You delete stack A. AWS deletes resources stack A owns, including the OIDC provider (unless DeletionPolicy: Retain).
- Stack B still has CreateOidcProvider=false. It does not recreate the provider. Its role’s trust policy still names an ARN that no longer exists. `AssumeRoleWithWebIdentity` fails.

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,384 @@
...
+
+  ElasticResourceType:
+    Description: deployment for Elastic Cloud Hosted, project for Elastic Cloud Serverless.
+    Type: String
+    AllowedValues:
+      - deployment
+      - project
+
+  ElasticResourceId:
+    Description: Kibana component ID of the deployment, or the Serverless project ID.
```
bhapas, line 20:
```suggestion
    Description: Elastic deployment ID of the deployment, or the Serverless project ID.
```

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,384 @@
...
+        - https://workload-identity-issuer.${ElasticCloudRegion}.${ElasticCloudProvider}.${Domain}/orgs/${ElasticOrganizationId}
+        - Domain: !FindInMap [EnvironmentDomain, !Ref ElasticCloudEnvironment, Domain]
+      ClientIdList:
+        - sts.amazonaws.com
+    Condition: ShouldCreateOidcProvider
+
+  ElasticWorkloadIdentityRole:
+    Type: AWS::IAM::Role
+    Properties:
+      RoleName: !Sub ElasticWorkloadIdentity-${AWS::StackName}
```
bhapas, line 83: `ElasticWorkloadIdentity-` prefix is 26 characters; IAM role names max out at 64. 

Maybe Long stack names will fail create.

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,384 @@
+AWSTemplateFormatVersion: "2010-09-09"
+
+Description: 'Creates an IAM Role for Elastic Workload Identity Federation. The role can be assumed only by the given Elastic deployment or project and grants read-only permissions for all Elastic AWS integrations that support Identity Federation: the AWS package integrations (GuardDuty, Inspector, Config, CloudWatch, and metrics services)'
+
+Parameters:
+  ElasticOrganizationId:
+    Description: Elastic Cloud organization ID.
+    Type: String
+    AllowedPattern: ^[0-9]+$
```
bhapas, line 9: https://github.com/elastic/cloud/blob/master/scala-services/domain-core/src/main/scala/co/elastic/cloud/domain/ids/OrganizationId.scala#L34 states that it should be `[A-Za-z0-9]{1,32}`

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: Done — added `DeletionPolicy: Retain` and `UpdateReplacePolicy: Retain` (cfn-lint requires both to fully protect the resource).

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: Applied the suggestion.

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: Fixed — updated `AllowedPattern` to `^[A-Za-z0-9]{1,32}$` and the `ConstraintDescription` to match.

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: Fixed — switched from `${AWS::StackName}` to the last 12-hex-char segment of the stack ID (`!Select [4, !Split ["-", !Ref AWS::StackId]]`). Role name is now always `ElasticWorkloadIdentity-<12chars>` = 36 chars total, well under the 64-char IAM limit.

bhapas (approved):

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,462 @@
...
+              respond(event, "SUCCESS", arn, data={"Arn": arn})
+
+
+          def provider_arn(function_arn, url):
+              _, partition, _, _, account, _ = function_arn.split(":", 5)
+              return f"arn:{partition}:iam::{account}:oidc-provider/{url.removeprefix('https://')}"
+
+
+          def ensure_provider(arn, url, client_id):
+              for attempt in range(5):
```
bhapas, line 147: The retry loop gives up after 15 seconds total (1+2+4+8 s). That is tight for the two failure modes it is meant to absorb.
**Scenario:** fresh account, first stack. CloudFormation creates the Lambda role, then the function, then invokes it right away. IAM permission propagation regularly takes 10 to 20 seconds, so all 5 attempts hit AccessDenied, the handler reports FAILED, and the stack rolls back. The customer retries a minute later and it works, which looks like a flaky product. The function timeout is 120 s, so there is room.

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,462 @@
...
+          def handler(event, context):
+              if event["RequestType"] == "Delete":
+                  respond(event, "SUCCESS", event["PhysicalResourceId"])
+                  return
+              props = event["ResourceProperties"]
+              arn = provider_arn(context.invoked_function_arn, props["Url"])
+              try:
+                  ensure_provider(arn, props["Url"], props["ClientId"])
+              except Exception as e:
+                  # Report every failure; an unanswered request only fails after ServiceTimeout.
```
bhapas, line 133: The function role has no CloudWatch Logs permission, so an unhandled crash leaves no trace anywhere.
▎ The handler reports handled errors back to CloudFormation, which is good. But if anything fails before or inside `respond` (for example the callback URL is blocked by an egress rule, or a `KeyError` on an unexpected event shape), the stack hangs for the full 10-minute `ServiceTimeout` and then fails with a generic "did not stabilize" message. Nobody, including us on a support call, can see why. Attaching `AWSLambdaBasicExecutionRole` (or the three logs: actions scoped to this function's log group) costs nothing security-wise; the code does not log secrets.

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,462 @@
...
+              payload = json.dumps(body).encode()
+              request = urllib.request.Request(
+                  event["ResponseURL"],
+                  data=payload,
+                  method="PUT",
+                  headers={"Content-Type": "", "Content-Length": str(len(payload))},
+              )
+              urllib.request.urlopen(request, timeout=10)
+
+  WiiOidcProvider:
```
bhapas, line 192: Do we need to consider an optional `ExistingOidcProviderArn` parameter as an escape hatch for locked-down accounts.
**Scenario**: enterprise customer whose SCP blocks `lambda:CreateFunction`, or requires all Lambdas in a VPC with specific tags. The legacy template works for them; this one fails at the Lambda resource before touching IAM. If the parameter is set, a `Condition` skips the function, role and custom resource, and the trust policy uses the given ARN via `!If.` Kibana never needs to fill it; it is for manual deploys only.

deploy/cloudformation/federated-identity-wii-aws.yml
```diff
@@ -0,0 +1,462 @@
...
+
+
+          def handler(event, context):
+              if event["RequestType"] == "Delete":
+                  respond(event, "SUCCESS", event["PhysicalResourceId"])
+                  return
+              props = event["ResourceProperties"]
+              arn = provider_arn(context.invoked_function_arn, props["Url"])
+              try:
+                  ensure_provider(arn, props["Url"], props["ClientId"])
```
bhapas: `ClientId` is read from the resource properties rather than fixed in code.

Anyone in the account who can invoke the function directly can add an arbitrary audience to the shared provider. The role trust still requires `aud = sts.amazonaws.com`, so the risk is low, but hardcoding `sts.amazonaws.com` in the handler makes the "cannot do anything else even if invoked directly" claim fully true and removes a property. Similar to [line 150 ](https://github.com/elastic/cloudbeat/pull/8385/changes#diff-a9d70c7ceece0f811bf6d6a471307812832d5ed9e2b111f842e6c649e0a6e158R150)

bhapas (commented):
@jeniawhite Overall the lambda changes look good.. Just couple of questions/comments

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: The customer still has to create the provider manually either way, so this doesn't save them any steps compared to editing the YAML. It just adds an additional configuration.

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: Fixed — bumped to 2, 4, 8, 16 s (30 s total before the last attempt), which clears the typical 10–20 s IAM propagation window with room to spare.

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: Fixed — added AWSLambdaBasicExecutionRole to the function role so crashes show up in CloudWatch instead of leaving a silent timeout.

deploy/cloudformation/federated-identity-wii-aws.yml
jeniawhite: Hardcoded sts.amazonaws.com in the handler and dropped ClientId from the resource properties. Each template ships its own Lambda, so if we ever need a different audience it goes in that template's handler — no allowlist needed.

bhapas (approved):

gsarantid:
I see the PR intentionally retains the shared provider and exposes `OidcProviderArn` for manual cleanup. Do we need to define how customers identify and remove orphaned providers now, or is that intentionally out of scope for this phase?

bhapas (commented):
@jeniawhite Looks to be some services / inputs are missing here

I tried them in https://github.com/elastic/cloud-iac-provisioner/commit/c1b362a

jeniawhite:
@bhapas This template's permissions are based and identical to current CC static template which was tested and verified by @seanrathier. I will unblock this PR by merging (this template won't be used up until integration changes get merged). If we encounter any gaps, we will just update the template and the job will re-upload it to S3.

